Issue #1, opened 2015-09-15
Need code which will apply target selection cuts on grzW1W2 *fluxes*


moustakas:
The function cuts.select_target() takes as input grz magnitudes, but we need something which can be applied to the DECaLS catalogs, which report grz fluxes in nanomaggies (which can be negative).  

The flux version of the current set of cuts are here:
  https://desi.lbl.gov/trac/wiki/TargetSelection

djschlegel:
Yu Feng's implementation documented with examples here:
http://desihub.github.io/imaginglss/

moustakas:
That functionality used to be there (in analysis.cuts()) but doesn't seem to be there anymore.  I've posted an issue to imaginglss, but separate from this do we want a dependence in desitarget on imaginglss?  My inclination is that imaginglss could/should depend on desitarget, but not the other way around.

djschlegel:
John,

Re-factoring that way would be good, e.g. moving some of Yu Feng’s functions
into desitarget.

Note Yu Feng has been careful to exclude regions wherever the depth cut isn’t
met in any band.  In practice, I think we include those regions with shallower depth,
and then either mask or otherwise deal with it downstream.  After all, unlike SDSS,
there’s no avoiding patches on the sky by plugging elsewhere.

-David

rainwoodman:
We can probably just pass the depth downstream.

The implementations of the TargetSelection wiki page in imaginglss is there:

https://github.com/desihub/imaginglss/blob/master/imaginglss/analysis/targetselection.py

The main script is 

https://github.com/desihub/imaginglss/blob/master/scripts/select_objs.py

I believe you can ask for any sigma level confidence as a parameter. (-sigma-z, ...)

My version doesn't use the depth information in the catalogue since I expect to read the depth-images for 'randoms' as well. Since you don't deal with randoms, you don't need this complication.

rainwoodman:
But are we surely will define targets in insufficiently deep areas?

rainwoodman:
I looked at this and seems like the easiest solution is to convert the input to a dictionary. 

If not opposed I will file a PR?

rainwoodman:
I attempted but find this won't magically switch to nanomaggies, thus give up.

What if I just port the code in imaginglss here?

sbailey:
Pull request #2 addresses this with example cuts on fluxes.  Closing this ticket.  Please open new tickets for other specific issues.

PR #2, opened 2015-09-18
Port ImagingLSS target selection cuts to desi-target.

rainwoodman:
The testing code in example.py works on python2, but I am not sure how we plan to consolidates the infrastructure.

We shall move example.py outside of the package tree -- is there a desihub guideline for locations of these scripts?


rainwoodman:
This should have implemented the features in Issue #1 , also added the cuts by shape and brick-primary-ness

moustakas:
Thanks for working on this, Yu!  Would it be possible to develop a lower-level script which is agnostic as to where the fluxes come from?  For example, I would like to select targets from among a set of galaxy templates, and Jamie is working on selecting targets from various mocks.  Is this easy enough to implement, while still retaining the flexibility that you need for imaginglss?  (Also note that these other selections don't care about depth.)

rainwoodman:
- targetselection.py uses no depth

My current thoughts is to implement TS code of each different input source as a .py file.
The rationale behind this is that 
- some fields are just not applicable to the mocks. e.g. mocks likely have no extinction, or brick primary.
- small level of duplication of the nano-mags cuts are probably fine. It is easy to maintain this by the person responsible for that mock.
- way may be able to refactor the duplicated code away later anyways; but it is hard to do this with pure imagination before the code is there.

If you send me some example data of any of the examples down there, I can spawn a file similar to decals.py for it.

moustakas:
A very simple example of how I *used* to call your target-selection code in imaginglss is here:
  https://github.com/desihub/desisim/blob/desisim-templates/py/desisim/templates.py#L193
It would be great to be able to apply the nominal cuts this simply.

Another issue to think about -- but which we don't have to solve right now -- is how to deal with modifications to the target-selection criteria.  For example, if someone wants to test different criteria (whether for one or more types of objects), how should we best incorporate those changes without affecting the nominal cuts?

rainwoodman:
While that example was simple, it could deal with BrickPrimary etc. What about this:

```
python desi-ts.py --profile=decals.py -t LRG -t ELG -t BGS cataloguefilename outputdir
python desi-ts.py --profile=mock.py -t LRG -t ELG -t BGS mockfilename outputdir
python desi-ts.py --profile=lightconemock.py -t LRG -t ELG -t BGS lightconefilename outputdir
```
So no matter what comes in, the same thing comes out.

To test a new TS criteria, we just need to 
```
cp mock.py mock-dev.py
vim mock-dev.py
python desi-ts.py --profile=mock.py -t LRG -t ELG -t BGS mockfilename outputdir-dev
```

And once satisfied,
```
vimdiff mock-dev.py mock.py
```

rainwoodman:
My concern over `BrickPrimary` or `shapedev`, `mw_transmission` was that it feels awkward to strip it out from the main DECALS TS query, just because the mocks don't have them.

rainwoodman:
- added mock.py. It is decals.py stripped out DECALS specifics. 

This shall give you familiar outputs:

```
import mock

selection = where(mock.LRG(dict(r=rmag, g=rmag, z=zmag, W1=w1mag, W2=w2mag)))
```

- We just need a driver (some refactor of example.py and cuts.py) to make use of these. It would be of great help if you can send me a few input files so I can run a few test cases...

Then the main structure is ready to go and easy to for production.

moustakas:
Can you explain a little what npyquery is doing, and why it's necessary?  Is it a speed issue?  Importing mocky.py is throwing a 
```
    268 
    269     def __getitem__(self, index):
--> 270         return self.operands[i]
    271 
    272     def flatten(self, operands):

NameError: global name 'i' is not defined
```
error in npyquery for me.  

Also, the code is still taking as input the grz and W1W2 magnitudes, when we want it to take as input the corresponding fluxes (or perhaps have it optionally parse input magnitudes, but not have that be the default).

rainwoodman:
Thanks for catching the bugs!

There was another typo in mock.py which I just fixed. I didn't have data to run a test I won't be surprised to observe more problems.

1. npyquery is a small tool I played around in imaginglss for querying large arrays. It does scale better at querying large arrays as it is more efficient in memory usage than plain numpy boolean operations. More importantly,it allows us to split the policy (TS selection conditions) from the implementation (code). The concept is somewhat similar to several other popular python tools (eg bcolz or TinyDB) but this guy is just one file. We may get rid of it for now, but eventually may need to pull something very similar back in. 

2. To change to flux just edit the top lines in mock.py. I will mock up an example with fake data later of the day.

moustakas:
To test the code you could just create a scatterplot of grzW1W2 fluxes with the following (magnitude) ranges:
```
r = [16,24]
g-r = [-0.5,2]
r-W1 = [-2,6]
r-z = [-0.5,2.5]
```
and then check to be sure the correct points in color-color space are chosen.

It's your call as to whether to keep or utilize npyquery.  I think as long as we keep 'mock.py' (or decals.py and their equivalents) reasonably simple, then others can feel free to develop alternative target-selection criteria. 

A couple additional comments:

1. With desitarget in my PYTHONPATH the import statements needed to be
```
from desitarget.npyquery import Column as C
from desitarget.npyquery import Max, Min
```

rainwoodman:
-  The current code worked without 'desitarget' in my path.

```
from npyquery import Column as C
from npyquery import Max, Min
```

We need to take a look how stephen deal with in-package dependency from other packages.

- I added an example. The leaner approach would be to say the code will take nothing but FLUX. We can use a function, like TOFLUX to convert mag to flux at the `caller` side.

rainwoodman:
HI @moustakas and @sbailey 
I committed another possible solution, as mock2.py

The idea is to make two ends meet.
- Simplifying decals TS cuts by making assumptions on the specific columns in decals;
- Convert the FLUX variables into a 'minimal' schema (just fluxes) with a simple helper function.

It seems to be a balance between 'readness' and 'usability', depending if you like the line
MyQSO = make_most_assumptions(MyQSO). 

I'd like to get over this and start looking into the output schema...

sbailey:
Hi Yu.  I must admit that I don't really understand this.  I was expecting to see code something like:

```
objects = read_data(filename)  #- or read_mocks(filename)
islrg = select_lrg(objects)    #- returns boolean array
```
(or variants of that such as `targetmask = apply_target_selection(objects)` returning bitmask array...)

Reading in different sources would be replacing `read_data()` with `read_mocks()` or `read_db()`; each of those would standardize the data into a common set of columns and fill in any necessary blanks.

To try different TS cuts, one would replace `select_lrg(objects)`.

I'm having a hard time mapping those two ideas onto this code.  Is `makedecals` + `make_most_assumptions` the pieces that standardize inputs to a common set of columns?  And decals.py contains the canonical cuts, while MyQSO is an example of how one would apply custom cuts?

sbailey:
Is this PR ready to go again after the "massive clean up", or are there any more little pending changes before the final review and merge?

rainwoodman:
As far as the major code structure are concerned, this PR is ready for a final review.

Issue #3, opened 2015-10-13
targetselection.py doesn't work with astropy 1.0.5


sbailey:
With astropy 1.0.5, targetselection.py dies with:

```
$ python $DESITARGET/bin/targetselection.py /data/legacysurvey/dr1/tractor/000/tractor-0009p187.fits blat.fits
Traceback (most recent call last):
  File "/Users/sbailey/desi/git/desitarget/bin/targetselection.py", line 44, in <module>
    main()
  File "/Users/sbailey/desi/git/desitarget/bin/targetselection.py", line 35, in main
    mask = cut.apply(candidates)
  File "/Users/sbailey/desi/git/desitarget/py/desitarget/internal/npyquery.py", line 145, in apply
    result[s] = v.visit(self)
  File "/Users/sbailey/desi/git/desitarget/py/desitarget/internal/npyquery.py", line 369, in visit
    return self.m[t](node)
  File "/Users/sbailey/desi/git/desitarget/py/desitarget/internal/npyquery.py", line 415, in visit_expr
    ops = [self.visit(a) for a in node.operands]
  File "/Users/sbailey/desi/git/desitarget/py/desitarget/internal/npyquery.py", line 369, in visit
    return self.m[t](node)
  File "/Users/sbailey/desi/git/desitarget/py/desitarget/internal/npyquery.py", line 415, in visit_expr
    ops = [self.visit(a) for a in node.operands]
  File "/Users/sbailey/desi/git/desitarget/py/desitarget/internal/npyquery.py", line 369, in visit
    return self.m[t](node)
  File "/Users/sbailey/desi/git/desitarget/py/desitarget/internal/npyquery.py", line 409, in visit_column
    return node.apply(self.array, self.s)
  File "/Users/sbailey/desi/git/desitarget/py/desitarget/internal/npyquery.py", line 243, in apply
    return array[self.name][s]
  File "/Users/sbailey/anaconda/envs/numpy1.9/lib/python2.7/site-packages/astropy/io/fits/fitsrec.py", line 480, in __getitem__
    return self.field(key)
  File "/Users/sbailey/anaconda/envs/numpy1.9/lib/python2.7/site-packages/astropy/io/fits/fitsrec.py", line 620, in field
    field = _get_recarray_field(base, name)
  File "/Users/sbailey/anaconda/envs/numpy1.9/lib/python2.7/site-packages/astropy/io/fits/fitsrec.py", line 1138, in _get_recarray_field
[...]

PR #4, opened 2015-10-14
don't try to force UPPERCASE column names; FITS_rec will do that for you

sbailey:
Updates desitarget.read_tractor() to just return the astropy.io.fits.FITS_rec object as-is rather than trying to force the column names to be capitalized.  FITS_rec objects will already support upper- or lower-case column names without having to do anything special.  Trying to force it breaks with more recent versions of astropy (perhaps starting with 1.0.2?).  Fixes issue #3 .

Review question for @rainwoodman: did you actually have a case where you had to force the column names to be uppercase, or were you just expecting to need to do this?  If you added this code because of some case that was broken without it, please recheck because I may have broken that.

### DIFF
--- a/py/desitarget/io.py
+++ b/py/desitarget/io.py
@@ -46,15 +46,8 @@ def read_tractor(filename):
 
         Returns:
             ndarray with the tractor schema, uppercase field names.
-        
     """
-    data = fits.open(filename)[1].data.copy()
-
-    # convert all fields to upper case. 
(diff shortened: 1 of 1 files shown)


rainwoodman:
Let's merge this to make sure we can run on astropy 1.0.5. 

I will look into getting normalized cases into in astropy.io.fits. (cfitsio has this feature, btw).

PR #7, opened 2015-10-21
Scan through a directory.

rainwoodman:
```
targetselection.py src/ dest/
```

Add tsbits to all tractor files in src/ and save the result to dest/

### DIFF
--- a/bin/targetselection.py
+++ b/bin/targetselection.py
@@ -1,15 +1,16 @@
 from __future__ import print_function
 import numpy
-from desitarget.io import read_tractor, write_targets
+from desitarget.io import read_tractor, iter_tractor, write_targets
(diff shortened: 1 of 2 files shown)


sbailey:
It looks like the iteration over tractor files will raise a ValueError if it encounters any non-tractor files in the directory tree.  It seems like it should just harmlessly step over them instead.

rainwoodman:
Ha! Let me explain:

- parse_filename would raise a ValueError, but the iteration would catch it and pass though. 

- the function parse_filename is there to highlight that in other DRs the format of the filename may change.

sbailey:
Gotcha.  I added comments to make that more obvious; removed a leftover (?) print statement, and make the regex more specific.  Will merge now.

rainwoodman:
Thanks for the clean up!

Issue #8, opened 2015-10-26
npyquery crashes and performance issues


sbailey:
Branch sjb_multifile_ts refactors targetselection.py to write one target catalog output for for many tractor file inputs.  This uncovered two problems with the npyquery cuts:

**1. It crashes after ~350 input bricks** with "Fatal Python error: deallocating None".  This is not a catchable exception, it is a hard hard fatal crash.  It's likely that the actual bug is in one of the underlying compiled libraries, but the complexity of the npyquery tree code makes it difficult to isolate exactly where and why.

To see this, use branch npyquery_crash, which is a branch off of sjb_multifile_ts that uses the npyquery cuts instead of the workaround vanilla numpy cuts in sjb_multifile_ts:

```
[edison sjbailey] targetselection.py /project/projectdirs/cosmo/data/legacysurvey/dr1/tractor blat.fits --verbose
50 bricks; 22.0 bricks/sec
100 bricks; 22.7 bricks/sec
150 bricks; 21.1 bricks/sec
200 bricks; 24.4 bricks/sec
250 bricks; 23.0 bricks/sec
300 bricks; 24.6 bricks/sec
350 bricks; 27.0 bricks/sec
Fatal Python error: deallocating None
Aborted
```

**2. It is ~2.5x - 9x slower than just naively making cuts on the input numpy structured arrays**.

```
from desitarget.cuts import select_targets as select_oldstyle
from desitarget.cuts_npyquery import select_targets as select_npyquery
import fitsio
from astropy.io import fits

d1 = fitsio.read('tractor-0001m002.fits', 1, upper=True)
d2 = fits.getdata('tractor-0001m002.fits')

#- Using numpy arrays from fitsio, npyquery is 2.5x slower
%timeit x = select_oldstyle(d1)   #- 2.77 ms
%timeit x = select_npyquery(d1)   #- 7.07 ms

#- And using FITS_rec objects from astropy.io.fits, npyquery is 9x slower
#- (and both are much slower than the real numpy arrays from fitsio... grr astropy...)
%timeit x = select_oldstyle(d2)  #- 13.4 ms
%timeit x = select_npyquery(d2)  #- 126 ms
```

[...]

PR #9, opened 2015-10-26
single target selection output from multiple tractor inputs

sbailey:
This PR updates targetselection.py to take multiple inputs and write a single target selection file as output (instead of one output file per input).  It does this in an optionally parallel manner such that on edison DECaLS DR1 can be processed in under 4 minutes (!).  Several other important changes:

1. Switches to using fitsio, since astropy.io.fits is slow and buggy in ways that prevented it from being effective here.  This does require fitsio 0.9.8rc1 or later since fitsio also had a bug parsing boolean columns (like BRICK_PRIMARY...)

2. The npyquery cuts are moved to a separate file and replaced with simple numpy based cuts which are faster and don't crash (see issue #8).  We may bring back npyquery in the future (thus not deleting it from repo) but for now it isn't used.

4. Various changes to work around DR1 data model inconsistencies, handling version numbers, updating the output data model, iterating over nested directories of tractor files, etc.

Still to do and not included in this pull request:

1. Add FSTD (F-star standard stars) WDSTD (white dwarf standard stars) targets.
2. Define MWS (Milky Way Survey) target bit(s), even though that target selection will happen through an independent GAIA-based selection.


rainwoodman:
I will take a look and merge this -- then take a look at issue #8 .

rainwoodman:
I am tempting to do some refactoring of this PR:

- fix_dr1_dtype is attracted to read_tractor(); I'd rather move it out of the main executable.

- calc_nobs can treat multi-obs objects as different object types, and then converted boolean masks to values in objs in target.finalize; this way we have all final transformation of the data in one location, and the bridge is always booleans.

- now that the crash part of #8 is solved, the 2x performance with fitsio issue critical? If it is critical then we shall just drop npyquery, as soon as possible. Otherwise I would like to massage this back to npyquery before you start other object types, and start looking and constructing SQL queries.

- astropy.fits do have a uncertain future as they are planning for a major change in the memory storage to astropy/Tables (which may affect other desihub projects as well). To isolate us from this change, we should be able to convert astropy.fits output to an ndarray, though this is another `won't fix` on their side.

PR #10, opened 2015-10-27
Simple main

rainwoodman:
This PR simplifies the main function to avoid duplicated loop.

This is possible through a more flexible multiprocessing library, with a well defined reduction operation that is guaranteed to occur on the master process as jobs are finished.

Notice how the processing functions can now be defined nested in main, thus passing in the context for t0 via scoping, and how nbrick is naively modified in 'collect' to track the progress.

### DIFF
--- a/bin/targetselection.py
+++ b/bin/targetselection.py
@@ -55,7 +55,6 @@ def main():
         return fix_tractor_dr1_dtype(targets)
 
     nbrick = np.zeros((), dtype='i8')
-    targets = [] 
 
     def collect_results(result):
         ''' wrapper function for the critical reduction operation,
@@ -64,10 +63,11 @@ def main():
             rate = nbrick / (time() - t0)
             print('{} bricks; {:.1f} bricks/sec'.format(nbrick, rate))
         nbrick[...] += 1
-        targets.append(result)
+        return result
 
-    map_tractor(_select_targets_brickfile, ns.src, \
-        bricklist=bricklist, numproc=ns.numproc, reduce=collect_results)
+    bnames, bfiles, targets = \
+        map_tractor(_select_targets_brickfile, ns.src, \
+            bricklist=bricklist, numproc=ns.numproc, reduce=collect_results)
 
     #- convert list of per-brick items to single arrays across all bricks
     t1 = time()
--- a/bin/targetselection.py
+++ b/bin/targetselection.py
@@ -54,6 +54,8 @@ def main():
 
         return fix_tractor_dr1_dtype(targets)
 
+    # a numpy scalar allows updating nbrick in python 2
+    # c.f https://www.python.org/dev/peps/pep-3104/
     nbrick = np.zeros((), dtype='i8')
 
     def collect_results(result):
@@ -62,7 +64,10 @@ def main():
         if ns.verbose and nbrick%50 == 0:
             rate = nbrick / (time() - t0)
             print('{} bricks; {:.1f} bricks/sec'.format(nbrick, rate))
+
+        # this is an in-place modification
         nbrick[...] += 1
+
         return result
 
     bnames, bfiles, targets = \
--- a/bin/targetselection.py
+++ b/bin/targetselection.py
@@ -25,22 +25,6 @@ ap.add_argument('-v', "--verbose", action='store_true')
 ap.add_argument('-b', "--bricklist", help='filename with list of bricknames to include')
 ap.add_argument("--numproc", type=int, help='number of concurrent processes to use', default=1)
 
-def _select_targets_brickfile(filename):
-    '''Wrapper function for performing target selection on a single brick file
-    
-    Used by _map_tractor() for parallel processing'''
-    objects = read_tractor(filename)
-    targetflag = select_targets(objects)
-    keep = (targetflag != 0)
-
-    objects = objects[keep]
-    targetflag = targetflag[keep]
-    numobs = calc_numobs(objects, targetflag)
-
-    targets = desitarget.targets.finalize(objects, targetflag, numobs)
-
-    return fix_tractor_dr1_dtype(targets)
-
 def main():
     ns = ap.parse_args()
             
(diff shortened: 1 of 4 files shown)


moustakas:
Just a quick comment: The DECaLS team has committed to delivering 360 'sweep' files (one for each 1-degree slice of RA) with each data release.  These will contain many more objects than the individual tractor catalogs, but will contain a smaller set of columns (but specifically the minimum set of columns required for target selection), so this may alleviate some of these performance issues.

Along these lines, if the code that is being developed here can help *make* the sweep files (since it requires the time-consuming operation of looping over every brick), that would be fantastic and very helpful.

sbailey:
Good eyes, John.  We are indeed developing some tools to simplify iterating over tractor files and doing parallel processing on them.  We even considered developing this within legacysurvey code, but decided to start out with self contained desitarget while we sort out what is most useful.  Pieces could then be moved to legacysurvey if that seems like the better long term home.

iter_tractor() and map_tractor() may already be what you need; let's chat when you are ready to make the sweeps.  We should also extend this to iterate over sweep files instead, and then this target selection code will be way fast (doesn't work for DR1 because sweeps are missing MW extinction).

More comments on this PR itself later.

sbailey:
I can see the utility of the extra features that sharedmem provides, but I want to do some due diligence before bringing this in:

Overall this does simplify the main loop, but at the cost of replacing 59 lines with 844 lines of expert code that will require someone at Yu Feng's level to maintain if anything goes wrong.  All of the simplifications except the ability to print progress could have been handled with far fewer lines of code (for this particular case).  But I will admit that sharedmem  does look pretty cool and could be more useful than vanilla multiprocessing.Pool.  We might consider moving it to desiutil for usage elsewhere too if it is rock solid robust.

This use of "reduce" is somewhat non-standard.  Typically "reduce" is used to combine results from the map steps, and outputs of one reduce can be used as the input to other reduces, and they aren't necessarily run in order and they could even be run in parallel.  This implementation seems to treat reduce just as something that runs after each map function call.  I think the phrase "reduction step that is guaranteed to run on the master process's scope" means that the reduce calls are guaranteed to happen one at a time and thus avoid race conditions (e.g. for appending items to the targets list, in this case).  Is that correct?  I think this is just a matter of nomenclature, but I want to make sure that the reduce steps cannot happen simultaneously (by construction, not by luck).

Are the results of sharedmem.MapReduce().map() guaranteed to be in the same order as the input sequence?

Did you bring this in to desitarget just by copying the code from http://github.com/rainwoodman/sharedmem, or did you use something fancier like git submodule or git subtree?  Both are technically "better" than just copying code, but a real usability disaster for non-experts and we should be wary about relying upon people using them correctly.

bin/targetselection.py
```diff
@@ -45,54 +36,51 @@ def main():
...
+
+        objects = objects[keep]
+        targetflag = targetflag[keep]
+        numobs = calc_numobs(objects, targetflag)
+
+        targets = desitarget.targets.finalize(objects, targetflag, numobs)
+
+        return fix_tractor_dr1_dtype(targets)
+
+    nbrick = np.zeros((), dtype='i8')
```
sbailey: Can you add a comment about why the nbricks counter is a 0-dim numpy array instead of just a python int variable?  The use of `nbrick[...] += 1` below was a bit surprising.

bin/targetselection.py
```diff
@@ -45,54 +36,51 @@ def main():
...
+    targets = [] 
+
+    def collect_results(result):
+        ''' wrapper function for the critical reduction operation,
+            that occurs on the main parallel process '''
+        if ns.verbose and nbrick%50 == 0:
+            rate = nbrick / (time() - t0)
+            print('{} bricks; {:.1f} bricks/sec'.format(nbrick, rate))
+        nbrick[...] += 1
+        targets.append(result)
```
sbailey: `map_tractor` returns `(bricknames, brickfiles, results)`, where `results = pool.map(function, brickfiles, reduce=reduce)`.  i.e. I think that `results` is the same as the `targets` array that is constructed inside the `reduce=collect_results` function.  Is that correct?  Is there any reason why the reduce function needs to construct that vs. just using the output from `map_tractor` directly?  It appears that `collect_results` may only be needed for coordinating printing progress from the parallel map workers (which is cool...)

Side note: the parallel code is now so fast that printing every 50 bricks is a bit overkill.  :)

rainwoodman:
Hi Stephen,

I initially wrote sharedmem in 2012, and has been used for 3 years (and three versions) -- most bugs have been explored if not all. I do recall two small issues than I haven't yet track down(since they do not affect productivity). The test suite on travis covers both python 2 and 3 for most use cases. 

Indeed, the 'reduce' operation implemented here differs from the canonical case you mentioned in the sense that it never runs in parallel; but it still serves as a 'reduction' of the results. In this case, the reduction is simply appending to a list. Note that 
'reduce' shall not block 'map'. (by definition, but I do not have a test case to show this.)

In fact, reduce is simply called to transform the result of 'map' before it is finally put into a heap, sorted, and returned. (https://github.com/rainwoodman/sharedmem/blob/master/sharedmem/sharedmem.py#L677). So yes the result of map is in-order, and yes, reduce is non-racing, out-of-order; all by construct.

Allowing any functions in map, (rather than module level functions) allows one to write less verbose code. Using reduce as a progress counter is just a simple case. I tried to work out a progress counter with mp.Pool yesterday, but failed. The smell was that I may need to override a connection manager to provide a global context -- hence I decided to drop in sharedmem and see.

This example is very naive. The true power of sharedmem is that it provides *critical section* and *order* constructs, as well as ndarray based shared memory blocks that the slave processes can *directly* write to. For example, we can avoid pickling by preallocating buffer on master with sharedmem.empty, fill them 'in-order', till near full in 'map' payload, and directly save them in a critical section. This can be how we build the sweep files if pickling becomes a bottleneck.

I copied the file here. I squashed all files into one 877 line file a few month ago, so that it can be dropped into any projects. The thought being a single file provides the same amount of tracking of a submodule yet needs no usability disaster.

I will address comments in the code separately.

bin/targetselection.py
rainwoodman: I added this commit as a second thought, mostly to demonstrate another way of building 'targets' list.(which is out-of-order)
The commit before this was still using the result of map_tractor (which is in-order).

I can revert this commit.

bin/targetselection.py
rainwoodman: nbrick[...] += 1 explicitly says the value stored at nbrick is updated where the object reference nbrick is intact. 

It is clearer than nbricks += 1, which depends on the type of nbricks behave differently, and breaks in python 2 (lack of nonlocal keyword). So [...] syntax will raise syntax error if nbrick is by mistake declared as a python int.

This use in code is derived from the construct in https://www.python.org/dev/peps/pep-3104/ (example 1) I will add a comment and reference in the code.

PR #14, opened 2015-10-28
WIP: build sweep files

rainwoodman:
This is a prototype. map_tractor API do not really allow safe direct memory
access (need critical sections)

Some functions are in wrong locaktions.

Lack of filtering of columns (need a definition of the sweep schema)

Currently grouping by bricks/ what about grouping by number of objects?

shall targetselection.py read sweep files?

This is a WIP and comments are welcomed. Please comment or add commits to the 'sweep' branch!

### DIFF
--- a/bin/buildsweep.py
+++ b/bin/buildsweep.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python
+
+from __future__ import print_function, division
+
+import numpy as np
+
+import desitarget
+import desitarget.targets
+from desitarget.io import read_tractor, write_targets
+from desitarget.io import iter_tractor, map_tractor
+from desitarget.io import fix_tractor_dr1_dtype
+
+from argparse import ArgumentParser
+import os, sys
+from time import time
+
+ap = ArgumentParser()
+### ap.add_argument("--type", choices=["tractor"], default="tractor", help="Assume a type for src files")
+ap.add_argument("src", help="Tractor file or root directory with tractor files")
+ap.add_argument("dest", help="Output sweep file")
+ap.add_argument('-v', "--verbose", action='store_true')
+ap.add_argument('-b', "--bricklist", help='filename with list of bricknames to include')
+ap.add_argument("--numproc", type=int, help='number of concurrent processes to use', default=1)
+
+def write_sweep(filename, data):
+    import fitsio
+    hdr = {}
+    fitsio.write(filename, data, extname='SWEEP', header=hdr, clobber=True)
+
+def main():
+    ns = ap.parse_args()
+            
+    #- Load list of bricknames to use
+    if ns.bricklist is not None:
+        bricklist = np.loadtxt(ns.bricklist, dtype='S8')
+    else:
+        bricklist = None
+        
+    #- Loop over bricks collecting target selection flags
+    #- and targets that passed the cuts
+    t0 = time()
+
+    def filter(filename):
+        '''Wrapper function for performing target selection on a single brick file
+        
+        Used by _map_tractor() for parallel processing'''
+        objects = read_tractor(filename)
+
+        return fix_tractor_dr1_dtype(objects)
+
+    # a numpy scalar allows updating nbrick in python 2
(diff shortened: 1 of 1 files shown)


moustakas:
This is fantastic!  Some questions and answers:

1. The development can continue to happen here, but eventually this code should be moved to legacypipe, since each imaging survey will deliver sweep files.  Target selection can then operate on those sweeps rather than on the individual Tractor/brick catalogs.  (I don't know that this has actually been decided, but it makes sense to me.)

2. As far as I know the SWEEP schema isn't documented anywhere (but will be for DR2), but here is the list of required columns I know about.  (@dstndstn should verify that these columns are still part of the Tractor data model.)
```
BRICKID
BRICKNAME
OBJID
BRICK_PRIMARY
TYPE
RA
RA_IVAR
DEC
DEC_IVAR
DECAM_FLUX
DECAM_FLUX_IVAR
DECAM_MW_TRANSMISSION
DECAM_NOBS
DECAM_RCHI2
DECAM_FRACFLUX
DECAM_FRACMASKED
DECAM_FRACIN
OUT_OF_BOUNDS
DECAM_ANYMASK
DECAM_ALLMASK
WISE_FLUX
WISE_FLUX_IVAR
WISE_MW_TRANSMISSION
WISE_NOBS
WISE_FRACFLUX
WISE_RCHI2
DCHISQ
FRACDEV
EBV
```

3. Grouping by bricks is presumably easiest, since grouping by objects requires inspecting every Tractor catalog.  Plus the "generate the sweeps" step will presumably only need to be run once per imaging data release.

4. In addition to building the DECaLS sweeps (which I was tasked with, so thank you for this code!) we also need to cross-match against the SDSS/DR12 specz catalog, as well as the corresponding visually inspected quasar catalog, DR12Q.fits.  Here is the ticket for this issue, for reference:
  https://github.com/legacysurvey/legacypipe/issues/69.  The code I wrote (in IDL) does this matching on the Tractor-catalog level at the same time as building the sweeps so that the catalogs don't need to be read in again.  With the parallelization that has occurred, however, the matching can probably most easily be done as a separate step (but will require another piece of code, also probably belonging to legacypipe).  For reference here is my IDL script:
  https://github.com/moustakas/moustakas-projects/blob/master/decals/build_dr1_specz.pro

rainwoodman:
Thanks. 

- I will plugin a schema with just these keywords, and leave out the headers for now.

- Do you want me to re-implement the IDL code as a python script? I think I did something similar a few month ago and can try to dig it out or just write another one and dump it in.

- We will likely need to refactor tractor related functions in desitarget.io if it is to be moved to legacypipe. E.g. a desitarget.io.tractor that can be easily moved to legacypipe. This can happen later in a different PR.

moustakas:
If you could (re)implement the "build-the-sweep" files code in Python (in legacypipe) that would be fantastic.  It strikes me that buildsweep.py is >90% of the way there.  Thanks!

sbailey:
I agree that this looks like great functionality to have, but that it should move over into legacypipe instead of merging it into master here.  If you need to have dependencies, I think it would be ok for desitarget to depend upon legacypipe but not vice-a-versa.

rainwoodman:
OK. I will move them there. 

I still wonder if the format of sweep files is finalized -- I think sweep files in a pure column format is easier to use than what we currently have, in probably any reasonable scenarios. 

I have examples of those 'column' files for DR1 at /project/projectdirs/m779/imaginglss/dr1/cache/DR1/catalogue/

One can basically mmap and directly access any chunks from any column from these files. There is no FITS sitting in the middle to slow things down.

moustakas:
We'll still want the FITS files as part of the official DECaLS data releases and for non-Python experts to develop/play around with target-selection.  But for "production" DESI target selection we should definitely keep this type of solution as a possibility.

rainwoodman:
This PR is moved to legacypipe. I am closing it.

rainwoodman:
The link to the new thread: https://github.com/legacysurvey/legacypipe/pull/78

Issue #26, opened 2016-01-07
Refactor desitarget.cuts.apply_cuts to deal with individual spectral classes


moustakas:
Refactoring desitarget.cuts.apply_cuts so that the nominal target-selection cuts can be applied to individual spectral classes would be very helpful for developing / testing target selection and for the templates work (in desisim).

sbailey:
+1 on this.  I suggest that `apply_cuts` still take the tables from tractor/sweep files as it currently does, calculates the extinction corrected fluxes as needed, and then call functions that explicitly take the variables that they need to make the cuts for each target class, e.g.`isELG(gflux, rflux, zflux)` rather than `isELG(objects)`.  These would return a boolean array that is the same length as the inputs, and `apply_cuts` would be the thing that assembles those back together into the target bitmasks.

moustakas:
@rainwoodman Would it be possible for you to tackle this refactoring? If not, no problem -- thanks!

rainwoodman:
No problem. I will file a PR on this along the lines Stephen suggested.

Yu

sbailey:
Thanks @rainwoodman for implementing this in PR #27 which is now merged.  The remaining piece is standard stars, which would have a messy function something like
```
def isStdStar(gflux, rflux, zflux, givar, rivar, zivar, gfracflux, rfracflux, zfracflux, psftype):
    ...
```
Is that what we want?  Other ideas?

moustakas:
We can reopen this if additional refactoring is needed in the future.

PR #27, opened 2016-01-31
Attempt to do #26

rainwoodman:
This is an attempt at #26 

The approach is to make each function of the target type easier to use. However the limitation is that FSTD will be very hard to implement along these lines -- Look how many arguments would FSTD take!

### DIFF
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -19,7 +19,7 @@ A collection of helpful (static) methods to check whether an object's
 flux passes a given selection criterion (e.g. LRG, ELG or QSO).
 """
 
-def LRG(rflux, zflux, w1flux, primary=None):
+def isLRG(rflux, zflux, w1flux, primary=None):
     """Target Definition of LRG. Returning a boolean array.
 
     Args:
@@ -49,7 +49,7 @@ def LRG(rflux, zflux, w1flux, primary=None):
 
     return lrg
 
-def ELG(gflux, rflux, zflux, primary=None):
+def isELG(gflux, rflux, zflux, primary=None):
     """Target Definition of ELG. Returning a boolean array.
 
     Args:
@@ -75,7 +75,7 @@ def ELG(gflux, rflux, zflux, primary=None):
 
     return elg
 
-def PSFLIKE(psftype):
+def psflike(psftype):
     """ If the object is PSF """
     #- 'PSF' for astropy.io.fits; 'PSF ' for fitsio (sigh)
     #- this could be fixed in the IO routine too.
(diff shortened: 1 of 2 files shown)


sbailey:
Conceptually this looks good, though it is failing tests, specifically that target selection works on a single row of a table in addition to on a full table.  @rainwoodman if you'll be at LBL this week, let's chat in person — it isn't obvious to me how to best fix this.

When apply_cuts(objects) gets a single row of a table, then type=objects['TYPE'] is a string like 'PSF' instead of an array.  PSFLIKE('PSF') is True, but ~PSFLIKE('PSF') evaluates to -2 instead of False, and this line in BGS() breaks:
```
        bgs &= ~PSFLIKE(type)
```

However, there is another problem (which wasn't caused by these changes, but was highlighted by debugging them): the bgs array itself is crazy wrong.  When apply_cuts(objects) gets a single row of a table, len(objects) is something like the number of columns rather than 1.  So code like this in unextinct_fluxes() return the wrong number of fluxes.
```
    dtype = [('GFLUX', 'f4'), ('RFLUX', 'f4'), ('ZFLUX', 'f4'),
             ('W1FLUX','f4'), ('W2FLUX','f4'), ('WFLUX', 'f4')]
    result = np.zeros(len(objects), dtype=dtype)
```

rainwoodman:
One way around the 0-length issue is probably to have two layers of cuts. The first layer will take 'object' arrays and used internally. The second layer is for human to use and takes kwargs. The second layer will take care of constructing objects array (and scalar stuff). It will be a lot of code, but seems to be the cleanest approach. 
(clean, consistent internal API vs messy, commandline friendly public API).

The second issue I am not sure what's going on. I will be at the lab soon. Let's take a look. numpy is at times not as solid as we thought it is...

sbailey:
Chatted with @rainwoodman, and fixed up support for both tables and rows again.

For the record, @rainwoodman recommended requiring only pure numpy arrays as input to apply_cuts() and providing a wrapper to convert astropy.io.fits.fitsrec.FITS_rec objects into real numpy arrays instead of trying to support both.  For end-user convenience I'm continuing to support the abomination that is returned by astropy.io.fits.getdata() without requiring the user to convert it.  The test coverage of this is pretty good, but if this gets out of hand we can fall back to @rainwoodman's suggestion.

The current factoring with functions like `ELG(gflux, rflux, zflux, primary=None)` seems fine to me; I don't see a compelling need to also provide an `ELG(table)` function.

Cuts for standard stars not yet factored out, but getting this comment in before proceeding.

rainwoodman:
Thanks for the clean up and merging!

PR #29, opened 2016-02-14
import numpy bug fix; added placeholder isFSTD function

moustakas:
This PR fixes a small bug and also adds a placeholder isFSTD() function which just applies color cuts for selecting F-type standards (as needed by desisim.templates).  This change basically completes #26 .


sbailey:
I renamed isFSTD to isFSTD_colors to (possibly) emphasize that this is only the color cuts, not the complete set of cuts make on FSTD targets. I also added tests that would have caught the numpy typo.

@moustakas if you agree with this renaming, go ahead and merge.

Issue #24, opened 2016-01-05
desitarget.cuts.apply_cuts borks if an input catalog has just one object / row


moustakas:
```
from astropy.io import fits
from desitarget.cuts import apply_cuts
cat = fits.getdata('tractor-3409p277.fits',1)
desi_target, bgs_target, mws_target = apply_cuts(cat[0])
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-5-3a3aa394b215> in <module>()
----> 1 desi_target, bgs_target, mws_target = apply_cuts(cat[0])

/Users/ioannis/repos/git/desihub/desitarget/py/desitarget/cuts.py in apply_cuts(objects)
     70 
     71     #- undo Milky Way extinction
---> 72     flux = unextinct_fluxes(objects)
     73     gflux = flux['GFLUX']
     74     rflux = flux['RFLUX']

/Users/ioannis/repos/git/desihub/desitarget/py/desitarget/cuts.py in unextinct_fluxes(objects)
     38 
     39     dered_decam_flux = objects['DECAM_FLUX'] / objects['DECAM_MW_TRANSMISSION']
---> 40     result['GFLUX'] = dered_decam_flux[:, 1]
     41     result['RFLUX'] = dered_decam_flux[:, 2]
     42     result['ZFLUX'] = dered_decam_flux[:, 4]

IndexError: too many indices for array
```

sbailey:
The error message is hardly illuminating, but the problem is that you are trying to apply_cuts on a single row instead of a table.  Note that in python a table with 1 row is different than the row itself.  Compare:
```
desi_target, bgs_target, mws_target = apply_cuts(cat)       # works: entire table
desi_target, bgs_target, mws_target = apply_cuts(cat[0:1])  # works: a table with 1 row
desi_target, bgs_target, mws_target = apply_cuts(cat[0])    # fails: 1 row of a table
```
So I think what you want is one of the first two, not the last one.  If you agree, please close this issue, otherwise please clarify the goal of apply_cuts(cat[0]).

Note that apply_cuts works with numpy.ndarray (e.g. from fitsio), astropy.io.fits.fitsrec.FITS_rec (e.g. when read using astropy.io.fits), and astropy.table.table.Table (e.g. when read with astropy.table.Table.read() .  One could do some gymnastics to catch the case of being passed a single row from each of these types, but I'd rather not unless there is a compelling reason to support single rows in addition to tables.

[...]

PR #32, opened 2016-02-24
merged target list

sbailey:
This PR adds desitarget.mtl module for managing the "merged target list" for input to fiber assignment.  It merges the target list from select_targets with any updates from the observations + spectro pipeline, and assigns priorities and number of requested observations based upon the parameters in targetmask.yaml .  The basic usage is:
```
mtl = desitarget.mtl.make_mtl(targets, zcatalog)   #- zcatalog is optional
```

this PR also updates that desitarget.targets.calc_priorities() and .calc_numobs() functions to be used by make_mtl().

This PR is a companion to [fiberassign PR #24](https://github.com/desihub/fiberassign/pull/24).

### DIFF
--- a/py/desitarget/__init__.py
+++ b/py/desitarget/__init__.py
@@ -38,6 +38,6 @@ def gitversion():
         return 'unknown'
 
 # desitarget.targetmask makes more sense?
-from .targetmask import obsconditions, obsstate
+from .targetmask import obsconditions, obsmask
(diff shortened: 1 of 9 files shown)


forero:
Tested it as shown [in this python script] (https://github.com/forero/SandboxDESI/blob/master/PR_test/20150226_test.py). It works as I expected.

PR #33, opened 2016-02-25
remove deprecated npyquery code

sbailey:
This PR removes the npyquery code that we aren't using any more, for the reasons listed in issue #8.  The crash turned out to be a numpy bug rather than an npyquery bug, but the cleverness and sophistication of npyquery makes it a maintenance risk and there isn't a compelling need for it.

If the postgres version of target selection becomes necessary then we could revive npyquery for its potential ability to generate both SQL and python cuts from the same code instead of having to maintain cuts in two different languages.  But for now looping over the sweep files is both faster and simpler, so npyquery is deprecated.


moustakas:
Sounds good to me.

sbailey:
Our npyquery dependency has been removed, but for the record the latest working version with a bug fix is sitting at git hash ab73e034c04e23baf2651dd1f1b0ceb60f901a7b on the npyquery_crash_fix branch that we are about to delete.  The hash and history will still be there, but the stale branch won't be.

PR #34, opened 2016-02-27
optionally allow the user to specify the tractor catalog columns to read

moustakas:
It would be helpful for a variety of target selection tests (and science!) to be able to specify the desired tractor catalog columns to read across many bricks.  This PR allows a user to optionally (explicitly) choose these columns.

### DIFF
--- a/py/desitarget/io.py
+++ b/py/desitarget/io.py
@@ -113,42 +113,46 @@ def read_mock_durham(core_filename, photo_filename):
     
     return data
 
-def read_tractor(filename, header=False, mycolumns=None):
+tscolumns = [
+    'BRICKID', 'BRICKNAME', 'OBJID', 'TYPE',
+    'RA', 'RA_IVAR', 'DEC', 'DEC_IVAR',
+    'DECAM_FLUX', 'DECAM_MW_TRANSMISSION',
+    'DECAM_FRACFLUX', 'DECAM_FLUX_IVAR',
+    'WISE_FLUX', 'WISE_MW_TRANSMISSION',
+    'WISE_FLUX_IVAR',
+    'SHAPEDEV_R', 'SHAPEEXP_R',
+    ]
+
+def read_tractor(filename, header=False, columns=None):
     """ 
-        Read a tractor catalogue. Always the latest DR. 
+        Read a tractor catalogue file.
         
         Args:
             filename: a file name of one tractor file
             
         Optional:
             header: if true, return (data, header) instead of just data
-            mycolumns: optionally specify the desired Tractor catalog columns to read
+            columns: optionally specify the desired Tractor catalog columns
+                to read; defaults to desitarget.io.tscolumns
 
         Returns:
             ndarray with the tractor schema, uppercase field names.
     """
     check_fitsio_version()
 
-    if mycolumns:
-        columns = mycolumns
+    if columns is None:
+        readcolumns = list(tscolumns)
     else:
-        #- Minimum columns needed for target selection and/or passing forward
-        columns = [
-            'BRICKID', 'BRICKNAME', 'OBJID', 'TYPE',
-            'RA', 'RA_IVAR', 'DEC', 'DEC_IVAR',
-            'DECAM_FLUX', 'DECAM_MW_TRANSMISSION',
-            'DECAM_FRACFLUX', 'DECAM_FLUX_IVAR',
-            'WISE_FLUX', 'WISE_MW_TRANSMISSION',
-            'WISE_FLUX_IVAR',
-            'SHAPEDEV_R', 'SHAPEEXP_R',
-            ]
+        readcolumns = list(columns)
         
     fx = fitsio.FITS(filename, upper=True)
     #- tractor files have BRICK_PRIMARY; sweep files don't
-    if 'BRICK_PRIMARY' in fx[1].get_colnames():
-        columns.append('BRICK_PRIMARY')
+    fxcolnames = fx[1].get_colnames()
+    if (columns is None) and \
+       (('BRICK_PRIMARY' in fxcolnames) or ('brick_primary' in fxcolnames)):
+        readcolumns.append('BRICK_PRIMARY')
     
-    data = fx[1].read(columns=columns)
+    data = fx[1].read(columns=readcolumns)
     if header:
         hdr = fx[1].read_header()
         fx.close()
--- a/py/desitarget/io.py
+++ b/py/desitarget/io.py
@@ -113,7 +113,7 @@ def read_mock_durham(core_filename, photo_filename):
     
     return data
 
-def read_tractor(filename, header=False):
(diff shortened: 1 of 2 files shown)


sbailey:
As is, this feature seems redundant with just directly reading the tractor file with fitiso:
```
data = desitarget.io.read_tractor(filename, mycolumns=columns)
# vs
data = fitsio.read(filename, columns=columns)
```
i.e. the primary functionality provided by desitarget.io.read_tractor is having it automatically pick the columns needed for target selection, and this PR bypasses that.  We might still encourage use of the desitarget.io.read_tractor, but that seems the only reason to include this mycolumns option as-is.

However — should we extend this functionality into desitarget.cuts.select_targets, so that you can run target selection and have it include non-default columns in the output?  That would be genuinely new functionality, by providing the parallelism, looping over files, etc. while also getting non-default columns.  In that case, it might be best to have an "extracolumns" keyword (or similar) so that it would pick the minimum needed for target selection, plus the extras requested by the user.

Thoughts?

moustakas:
Sorry for not giving more context. The idea was to leverage the tremendous parallelism of desitarget.cuts.select_cuts in order to build arbitrary (large) subsamples of the data.  (Up until late last night I was using the DR2 db for this -- with great help from @kaylanb and @nugent68 -- which worked quite well, but I gave up for two reasons: 1) the db changes the data model of the output catalogs which makes all my other existing code useless and 2) it was too frustrating trying to convert the output ASCII files to FITS files.  Plus the Python interface is infinitely more convenient and code is self-documenting.)

So what I did was to hack desitarget/bin/select_targets (which should be hyphenated, right?), write my own personalized desitarget.select_targets.apply_cuts function and then specify the Tractor catalog columns I wanted (the sweeps don't work because they only contain a fraction of the measurements I need).  The hacked, undocumented final script -- generously stolen from existing code -- is reproduced below.  I don't know if we (DESI-Data) wants to provide this generalized level of functionality but I at least -- wearing my Target Selection and DECaLS hats -- will be using this code many many times.

```
#!/usr/bin/env python

from __future__ import print_function, division

import os, sys
from time import time
import numpy as np

from desitarget.internal import sharedmem
from desitarget import io

mycolumns = [
    'BRICKID', 'BRICKNAME', 'OBJID', 'TYPE',
    'RA', 'RA_IVAR', 'DEC', 'DEC_IVAR',
    'DECAM_FLUX', 'DECAM_MW_TRANSMISSION',
    'DECAM_FRACFLUX', 'DECAM_FLUX_IVAR', 'DECAM_NOBS',
    'DECAM_ANYMASK', 'DECAM_DEPTH', 'DECAM_GALDEPTH',
    'WISE_FLUX', 'WISE_MW_TRANSMISSION',
    'WISE_FLUX_IVAR',
    'SHAPEDEV_R', 'SHAPEEXP_R',
    ]

def apply_cuts(objects):
    """Apply the cuts we want."""
    zfaint = 22.0

    #import pdb ; pdb.set_trace()
    keep = np.where((np.sum((objects['DECAM_NOBS'][:,[1,2,4]]>=3)*1,axis=1)==3)*
                    (np.sum((objects['DECAM_ANYMASK'][:,[1,2,4]]>0)*1,axis=1)==0)*
                    (objects['DECAM_FLUX'][:,4]<(10**(-0.4*(zfaint-22.5)))))[0]
    return keep

def select_targets(infiles, numproc=4, verbose=False):
    """
    Process input files in parallel to select targets
    
    Args:
        infiles: list of input filenames (tractor or sweep files),
            OR a single filename
        
    Optional:
        numproc: number of parallel processes to use
        verbose: if True, print progress messages
        
    Returns:
        targets numpy structured array: the subset of input targets which
            pass the cuts, including extra columns for DESI_TARGET,
            BGS_TARGET, and MWS_TARGET target selection bitmasks. 
            
    Notes:
        if numproc==1, use serial code instead of parallel
    """
    #- Convert single file to list of files
    if isinstance(infiles, (str, unicode)):
        infiles = [infiles,]

    #- Sanity check that files exist before going further
    for filename in infiles:
        if not os.path.exists(filename):
            raise ValueError("{} doesn't exist".format(filename))
    
    #- function to run on every brick/sweep file
    def _select_targets_file(filename):
        '''Returns targets in filename that pass the cuts'''
        from desitarget import io
        objects = io.read_tractor(filename, mycolumns)
        keep = apply_cuts(objects)
        
        return io.fix_tractor_dr1_dtype(objects[keep])

    # Counter for number of bricks processed;
    # a numpy scalar allows updating nbrick in python 2
    # c.f https://www.python.org/dev/peps/pep-3104/
    nbrick = np.zeros((), dtype='i8')

    t0 = time()
    def _update_status(result):
        ''' wrapper function for the critical reduction operation,
            that occurs on the main parallel process '''
        if verbose and nbrick%50 == 0 and nbrick>0:
            rate = nbrick / (time() - t0)
            print('{} files; {:.1f} files/sec'.format(nbrick, rate))

        nbrick[...] += 1    # this is an in-place modification
        return result

    #- Parallel process input files
    if numproc > 1:
        pool = sharedmem.MapReduce(np=numproc)
        with pool:
            targets = pool.map(_select_targets_file, infiles, reduce=_update_status)
    else:
        targets = list()
        for x in infiles:
            targets.append(_update_status(_select_targets_file(x)))
        
    targets = np.concatenate(targets)
    return targets

def main():

    # Build a catalog of all the grz, 3-pass sources.
    outfile = 'decals-dr2-fulldepth.fits'
    
    ramin = 240
    ramax = 241 # 250
    rarange = np.arange(ramin, ramax+1, 1).astype(str)

    tracdir = '/project/projectdirs/cosmo/data/legacysurvey/dr2/tractor'
    infiles = [io.list_tractorfiles(os.path.join(tracdir, this)) for this in rarange]
    infiles = np.concatenate(infiles).flatten()
    print('Found {} tractor catalogs.'.format(len(infiles)))

    cat = select_targets(infiles, verbose=True)
    print('Selected {} objects.'.format(len(cat)))
    io.write_targets(outfile, cat)
    
if __name__ == "__main__":
    main()
```

sbailey:
My original plan was to provide functionality like that to easily spin over tractor files in parallel to apply an arbitrary function (cuts or otherwise).  But then I thought I might just be getting too fancy and backed out to the simpler version.  But given that users want a more general tool for spinning over tractor files in parallel, we can do that.  You shouldn't have to hack select_targets at that level next time.

This PR is a small step in that direction, but I won't hold it up just because it isn't the full featured refactor.  I did make some small updates:
  * rename mycolumns -> columns to match fitsio syntax for specifying columns
  * desitargets.io.tscolumns is a list of the columns that will be read if columns=None (previously that we buried inside the function body and inaccessible without reading code).
  * added tests (of course)
  * fixed a bug with BRICK_PRIMARY when the input file has that in lower case (discovered by the tests... :)

PR #36, opened 2016-03-07
desitarget fix to MWS_STAR colors

crockosi:
Adds isMWSSTAR_colors to get a more reasonable g-r distribution for an old stellar population.

### DIFF
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -105,6 +105,38 @@ def isFSTD_colors(gflux, rflux, zflux, primary=None):
 
     return fstd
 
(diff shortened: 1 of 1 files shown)


sbailey:
Should we also update cuts.apply_cuts to call isMWSSTAR_colors and set a bit in mws_target for PSF-like objects in some magnitude range?  Even if the eventual MWS selection comes from GAIA, we could fill in the bits as a placeholder for now.

sbailey:
From a separate email, it sounds like trying to make MWS-like cuts on DECaLS data is premature.  I'll merge this now to get the color cuts in place for desisim to use, but we won't try to set MWS_TARGET bits during target selection yet.

PR #38, opened 2016-03-17
fix make_mtl calculation

sbailey:
This PR fixes a critical bug in make_mtl.  The previous code incorrectly assumed that `x = astropy.table.join(a, b)` would create `x` that was row-matched to `a` if `a` was a superset of `b`.  This is not the case -- `x` can end up in a random order relative to `a`, which was causing NUMOBS_MORE to be assigned to the wrong targets.

It also fixes a bug with NUMOBS_MORE for observed low-z QSOs.  Previously their priority was set to 0 (i.e. we don't need to continue observing them) but they still had NUMOBS_MORE > 0.  Now PRIORITY=0 implies NUMOBS_MORE=0.

### DIFF
--- a/py/desitarget/mtl.py
+++ b/py/desitarget/mtl.py
@@ -45,6 +45,10 @@ def make_mtl(targets, zcat=None, trim=True):
     ### mtl['NUMOBS_MORE'] = ztargets['NUMOBS_MORE']
     mtl['PRIORITY'] = calc_priority(ztargets)
 
+    #- If priority went to 0, then NUMOBS_MORE should also be 0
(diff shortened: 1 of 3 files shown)


forero:
Tests on cori/edison using ```/project/projectdirs/desi/mocks/preliminary/mtl/lowfat/``` data and code on ```https://github.com/forero/SandboxDESI/blob/59be52923caa4fb70ebf13b6979f2adab8195ebb/end2end/lowfat/simple_loop.py``` give reasonable results for the mtl files provided that the ```trim``` option is set to ```True```.

PR #39, opened 2016-03-18
work around astropy Table bugs

sbailey:
This PR works around astropy.table.Table bugs [#4707](https://github.com/astropy/astropy/issues/4707) and [#4708](https://github.com/astropy/astropy/issues/4708).  The combination of those bugs was causing Merged Target List tables with NUMOBS_MORE=1 to be masked when read back in, but only when those Tables had been generated including feedback from previous spectroscopic observations.

All tests pass, including a new test that verifies that this workaround works (i.e. if they "fix" it in a way that breaks our workaround, we'll find out).

### DIFF
--- a/py/desitarget/mtl.py
+++ b/py/desitarget/mtl.py
@@ -64,4 +64,9 @@ def make_mtl(targets, zcat=None, trim=True):
         notdone = mtl['NUMOBS_MORE'] > 0
         mtl = mtl[notdone]
 
+    #- filtering can reset the fill_value, which is just wrong wrong wrong
+    #- See https://github.com/astropy/astropy/issues/4707
+    #- and https://github.com/astropy/astropy/issues/4708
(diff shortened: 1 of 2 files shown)


forero:
Tests on cori/edison using ```/project/projectdirs/desi/mocks/preliminary/mtl/lowfat/``` data and [python code here to produce a simple survey](https://github.com/forero/SandboxDESI/blob/91cd38f93c299c310cb809f8d18943e8ac0b1680/end2end/lowfat/simple_loop.py) and [Ipython notebooks to process it](https://github.com/forero/SandboxDESI/blob/91cd38f93c299c310cb809f8d18943e8ac0b1680/end2end/lowfat/explore_results.ipynb) and [visualize it](https://github.com/forero/SandboxDESI/blob/91cd38f93c299c310cb809f8d18943e8ac0b1680/end2end/lowfat/leaky_pipeline.ipynb) give reasonable results.

Issue #21, opened 2015-12-11
make target selection QA plots


sbailey:
Make QA plots for target selection as a standard part of the output.

Version 0.2.0 produced target densities considerably different than expected; the first sanity check for bugs is to make the basic color color plots and see what the distributions look like pre- and post- cuts.  The underlying distributions may be much different than expected, or there may be a simple sign error that would be obvious from a plot but easy to miss when reading code.

geordie666:
One necessary piece of QA: Make sure no brick contains more targets than we will encode in `TARGETID`

moustakas:
Initial QA code incorporated by @geordie666 in #210.

PR #40, opened 2016-03-28
update color cuts to match the FDR

moustakas:
This PR attempts (hopefully successfully) to update the color cuts to match the latest cuts documented in Chapter 3 of the [Final Design Report](http://desi.lbl.gov/tdr) and on the [Target Selection wiki](https://desi.lbl.gov/trac/wiki/TargetSelectionWG/TargetSelection).

I need to make some QA plots to make sure there aren't any sign errors (closing #21 will be very useful in this regard) and if there are other suggested tests to increase coverage it would be great to incorporate them before merging.

@nathalieGit could you please take a look at the quasar cuts when you get a chance? In particular, note that the WFLUX and GRZ composite fluxes are now defined in desitarget.cuts.isQSO() rather than desitarget.cuts.unextinct_fluxes(), among other changes.

### DIFF
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -149,13 +149,13 @@ def psflike(psftype):
     psflike = ((psftype == 'PSF') | (psftype == 'PSF '))
     return psflike
 
-def isBGS(rflux, type=None, primary=None):
+def isBGS(rflux, objtype=None, primary=None):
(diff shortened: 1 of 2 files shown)


sbailey:
Looks good to me.  OK to merge after Natalie reviews the QSO cuts.

PR #42, opened 2016-04-12
typo (inequality sign) in qso TS

nathalieGit:
Correction of two inequality signs that were wrong in the definition of the main sequence for the QSO targets.

### DIFF
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -215,8 +215,8 @@ def isQSO(gflux, rflux, zflux, w1flux, w2flux, objtype=None,
     # Clip to avoid warnings from negative numbers raised to fractional powers.
     rflux = rflux.clip(0)
     zflux = zflux.clip(0)
-    mainseq &= rflux**(1+1.5) < gflux * zflux**1.5 * 10**((-0.075+0.175)/2.5)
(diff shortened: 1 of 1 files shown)


moustakas:
This looks fine to me and it's a small enough change that @nathalieGit should just merge.  Or should we specify an overseer for reviewing PRs to desitarget?  I was going to suggest @geordie666 but it turns out that he's not yet a part of desihub!

sbailey:
@geordie666 has been invited to desihub.

Changes to the target selection cuts should definitely keep @nathalieGit, @moustakas, and @geordie666 in the loop, but I don't think you need to specify a single overseer for approval of changes that are fixing typos.

For the record, issue #21 about making target selection QA plots probably would have helped with this: basic color-color plots highlighting what passed vs. didn't probably would have made it really obvious that we had a sign error in the cuts.

@moustakas does this imply that any QSO templates that I've recently generated have the wrong colors?

nathalieGit:
@sbailey, it  shouldn't make a big difference: these code lines are only used to define a region where main sequence stars lie in order to place stricter cuts on the QSO targets within this region (to reduce stellar contamination). With the wrong (ie previous) signs, the region contained 0 objects, the stricter cuts were thus not applied at all and the qso targets were a bit more contaminated than with the right (ie current) signs.

moustakas:
This bug did not affect the template colors (which still lack WISE photometry in any case).

PR #49, opened 2016-06-03
Mocks

forero:
This PR includes functions and scripts to create mock target files for the dark time survey.
[Here is a script](https://github.com/forero/SandboxDESI/blob/master/PR_test/20160603_make_mock_quick_survey.sh) that tests in the first two lines the new scripts in this PR and in the following two lines checks compliance with the quicksurvey scripts in desisim.


sbailey:
Thanks!  A few questions / comments / suggestions:

Let's switch naming from qsoI, qsoII to lya, qso or qsolya, qsotracer or something that is more informative about what is meant, especially since we aren't referring to what is typically meant by a type 1/2 QSO.

Please add a placeholder for fake ELGs (right now it appears that we have fake QSOs and fake LRGs, but not fake ELGs).

Naming: this is very dark time specific, and very specific to how things have been currently done in the preliminary mocks.  Consider whether this naming/structure will survive into future types of mocks (e.g. from Peder or Risa) and whether we should add another layer of structure, e.g.

```
desitarget/mock/darktime.py
desitarget/mock/brighttime.py
desitarget/mock/io.py
```

Style comments:

`build_mock_target(**kwargs)` etc : I prefer using `**kwargs` only when they could be genuinely anything, e.g. for passing through one function into another.  In this case, there is a defined set of what those kwargs should be, and thus it would be better to explicitly list them so that pydoc, ipython, etc. function introspection can see exactly which keywords are required and not rely upon the docstring alone.  e.g. `build_mock_target(qsolya_dens, qsotracer_dens, qsofake_dens, ...)`  It makes for a long argument list, but if they are all required, let's be explicit about that.

`select_population` returns a really long tuple of results; should that returns a dictionary instead so that we aren't dependent upon the exact order and have the flexibility to add other parameters later?

forero:
Thanks for the very thorough comments. I will implement them.

forero:
@sbailey this is now ready for you to review again.

moustakas:
This looks good.  My one comment is that the ```reduce``` function randomly selects a subset of objects. Do we want to initialize the random number generator (object) with an optional seed, so the results are reproducible?

forero:
@moustakas Good comment! I've just included the option to initialize the random seed.

PR #53, opened 2016-07-11
DR2 Tractor Catalogues psql tables and load scripts

kaylanb:
-- Version of code and tables used to create current dr2 schema in the psql db 
-- Updated files in db/ and bin/ so that db/*.dr1 schemas are now db/*.dr2 schemas and bin/*.py scripts for loading to psql db use the dr2 tractor catalogue naming scheme instead of dr1.


sbailey:
Thanks for working on this.  Questions / comments:

* Does the unit test have zero side effects?  It appears to load to an actual production database at NERSC.  Does it leave anything behind, or is the commented out "#con.commit()" mean that nothing is actually committed after the connection is closed?  At minimum a comment about that would be helpful so that someone doesn't naively uncomment it.
* Having a unittest is good, but it shouldn't be included in bin/ where it would get installed for all users to see.  Normally the unittests would be part of a the py/desitarget/test/ area, but those are run whenever someone does a "python setup.py test" or when code is pushed to github (where it won't be able to see the NERSC server anyway).  Consider putting this in the db/ directory instead.
* *.dr1 –> *.dr2 schema files.  It would be better for these not to have the version number in the file itself, but rather use git tags to reflect which version was used for loading the drN database.  If you wanted to keep multiple schemas around simultaneously, they could be grouped by db/drN/ subdirectories.  But since this is just replacing *.dr1 with *.dr2, I'd rather just let git versioning handle the versioning and drop the versions from the name.
* dr3 is imminent.  Will this also work for dr3, or is that another change?

I haven't reviewed the content at all; this is just commentary on the structure.

kaylanb:
Hi Stephen, 

I answered you questions below. Heres an overview:

This dr2 branch is what I did to get dr2 working. The dr3 version is incredibly better and  I will pull request it once this one is merged into master. For dr3 all tables are automatically created (in contrast everything from dr1/dr2 was hardcoded) and the load can be done in parallel (either python mpi4py or multiprocessing modules)

Does the unit test have zero side effects?  ...
* yes, it loads to the db but does not commit the load. I added a comment saying to leave commented out
Having a unittest is good, but it shouldn't be included in bin/ where it would get installed for all users to see.  ... Consider putting this in the db/ directory instead.
* done
dr1 –> dr2 schema files.  It would be better for these not to have the version number in the file itself, but rather use git tags to reflect which version was used for loading the drN database. 
* Okay, can you help me with git tags?
dr3 is imminent.  Will this also work for dr3, or is that another change?
* after this is merged I do a pr on the dr3 version. 
I haven't reviewed the content at all; this is just commentary on the structure.
* the dr3 pr will have lots of new code so the time reviewing structure should be spend on teh next pr. This dr2 pr simply updates existing files so very little/no time should be needed for reviewing strtucture...

sbailey:
OK, it sounds like the dr3 change is the one we really want, in particular the auto-generation of the tables instead of hardcoding everything.

git tagging: the basic idea is that you get everything into the state you want, and then you make a tag to say "this is a reference copy that I don't want to forget about".  ideally we would:
* merge this pull request
* make the tag
* install that tag
* use that installed copy of the code to load the database; this guarantees that the tag actually represents the version used.

For the actual mechanics of the tagging, please see https://desi.lbl.gov/trac/wiki/Computing/Software/Guidelines#CodeTags .  If something isn't clear about that, let's fix it in the documentation so that it will be clear for the next person.  That example uses versions like x.y.z .  We should make a tag like that for this code, but I wouldn't be opposed to also having a tag named "dr2" to be clear even though that is technically a non-standard version name.

Let's go ahead and leave the .dr2 on the schema files if this is going to be the last version of those, with the dr3+ equivalents auto-generated.  If I misunderstood, we can clean up this point in a later pass.  I'll merge now and you can proceed with the tagging off of master.

PR #54, opened 2016-07-11
automated loading of DR3 tractor catalogues to the DB

kaylanb:
Done
* tractor_load.py completely redone, parallel, to be compatible with any version of tractor catalugue should only need 1 or 2 new lines
* sbatch_tractor_load.py submit job to queue

Needed before merged
* drop .dr3 from table names
* more comments


kaylanb:
Additional before merge
* unit test that grabs a row from db and finds identical numbers to the corresponding tractor catalogue

kaylanb:
Ready for merge!

Finished unit test code for `tractor_load.py`. After tactor catalogues are loaded to the DB the unit test script `tractor_load_unittest.py` randomly chooses a subset of those catalogues and random rows for each one, then compares what it finds on the DB to what is in the catalogue (for all columns and any type: float, text, etc)

rainwoodman:
off topic, but is the DR3 coadd ready yet?

PR #56, opened 2016-07-28
color-cuts (targeting) functions now take all fluxes as input

moustakas:
Simple change to the low-level target-selection functions in ```desitarget.cuts``` (e.g., ```isLRG```, ```isFSTD```, etc.) so that they now take all the grzW1W2 fluxes as input, even though not all the fluxes are used.  This is needed for a major refactoring happening in https://github.com/desihub/desisim/pull/132.

The only update needed that I could find was to ```desisim.test.test_cuts```.

### DIFF
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -18,7 +18,7 @@ A collection of helpful (static) methods to check whether an object's
 flux passes a given selection criterion (e.g. LRG, ELG or QSO).
 """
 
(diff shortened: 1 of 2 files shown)


sbailey:
I'll merge now, but on your next update please also clarify that these are extinction corrected fluxes.

Thinking toward the future, we'll need to add PHOTSYS (north/south), and possibly other refactors for object specific terms like star/galaxy or S/N metrics.  But this at least gets us there for colors.

PR #59, opened 2016-08-18
fix LRG selection (z < 20.46)

mehdirezaie:


sbailey:
Good catch.  This is another case where #21 (make target selection QA plots) would have made it obvious that we were making the wrong cut in the code.

moustakas:
Nice catch, @mehdirezaie!  That was my typo.

mehdirezaie:
Thanks! :)

PR #62, opened 2016-08-23
ran 2to3 changes on everything except mocks and updated code. Haven't…

geordie666:
… tested functionality yet (against unit tests).

I have checked everything still runs against reference for desitarget in Python 2.7.11 and will now move on to fixing unit tests.

The only thing that broke in the 2->3 transition was:

import Queue as queue

in py/desitarget/internal/sharedmem.py

which Python 2.7 wants to be "import Queue" and Python 3.5 wants to be "import queue". Could maintain backwards compatibility with:

try:
    import Queue as queue
except ImportError:
    import queue

but we aren't trying to be explicitly backwards-compatible, I think.

It would be good if Kaylan could check that I didn't break tractor_load as I'm not up-to-speed on the db.

### DIFF
--- a/bin/tractor_load.py
+++ b/bin/tractor_load.py
@@ -14,7 +14,7 @@ import traceback
 from subprocess import check_output
 
 def print_flush(words):
-    print words
+    print(words)
     sys.stdout.flush()
 
 def read_lines(fn):
@@ -38,21 +38,21 @@ def write_index_cluster_files(schema,table=None):
     outname= '%s_index' % table
     rem_if_exists(outname)
     fin=open(outname,'w')
-    for key in indexes.keys(): 
+    for key in list(indexes.keys()): 
         for coln in indexes[key]: 
             if coln == 'radec': fin.write('CREATE INDEX q3c_%s_idx ON %s (q3c_ang2ipix(ra, dec));\n' % (key,key))
             else: fin.write('CREATE INDEX %s_%s_idx ON %s (%s);\n' % (key,coln,key,coln)) 
     fin.close()   
-    print "wrote index file: %s " % outname 
+    print(("wrote index file: %s " % outname))
     # Write cluster
     outname= '%s_cluster' % table
     rem_if_exists(outname)
     fin=open(outname,'w')
-    for key in indexes.keys():
+    for key in list(indexes.keys()):
         if 'radec' in indexes[key]:
             fin.write('CLUSTER q3c_%s_idx ON %s;\n' % (key,key))
             fin.write('ANALYZE %s;\n' % key)
-    print "wrote cluster file: %s " % outname 
+    print(("wrote cluster file: %s " % outname))
 
 
 def write_schema(schema,table_name,keys,sql_dtype,addrows=[]):
@@ -64,7 +64,7 @@ def write_schema(schema,table_name,keys,sql_dtype,addrows=[]):
     #add indexing names
     for row in addrows: fin.write('\n'+'\t'+row+',')
     #add catalogue's names
-    for key in keys:
+    for key in list(keys):
         stri= '\n'+'\t'+key.lower()+' '+sql_dtype[key]
         if key != keys[-1]: stri+= ','
         fin.write(stri)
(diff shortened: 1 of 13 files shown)


sbailey:
You removed the lines like `from __future__ import absolute_import, division, print_function`.  Although `2to3` does recommend that, I think that is only because they are unnecessary in python 3, but I think they are still valuable in python 2 during the transition.  I'd suggest that we liberally use that until we have done the final switch.  `unicode_literals` is the only one that I wouldn't import.  You can prevent `2to3` from making this recommendation by using the `-x future` option.

The following minor comments, but for consideration:

`2to3` seems to suggest more parentheses than necessary for print statements using formatted strings, e.g.

```
-    print "wrote cluster file: %s " % outname 
+    print(("wrote cluster file: %s " % outname))
```

should be fine just as

    print("wrote cluster file: %s " % outname)

I haven't figured out why it recommends those extra parentheses, e.g. a bare python file with just that line doesn't recommend that.  It smells like a 2to3 parsing bug, but that also seems surprising.

Stylistically, 2to3 often recommends changing:

    for key in data.keys():

to

    for key in list(data.keys()):

but I'm pretty sure that this is unnecessary since `data.keys()` returns an iterator in python 3 and we are just iterating over it in the for loop anyway.  In `desisim` and `desispec`, I converted this to

    for key in data:

which works for python 2 and 3 and keeps `2to3` happy too, but I'll admit that it is less clear for non-experts since it isn't obvious that data is a dictionary or that iterating over a dictionary iterates over its keys and not its values or its (key,value) pairs.  I could be convinced to go back to `for key in data.keys():` to keep it clear and just ignore `2to3` on this, or go with `for key in data:`, but `for key in list(data.keys()):` seems unnecessarily cluttered.

geordie666:
OK, will switch these things back today. I'll close the pull request for now.

The double parentheses aren't harmful, they're just clunky. But it looks like they are a genuine parsing bug in 2to3, rather than being due to any expectation of the future direction of the Python language...e.g., see the bottom of:

https://docs.python.org/3.0/library/2to3.html

So I'll switch those back too.

moustakas:
@geordie666 FYI -- You can keep the pull request open and continue to commit to this branch until it's ready to be re-reviewed and/or merged.

geordie666:
Thanks for the tip...I'll do that.

Ad.

kaylanb:
Adam, Thanks for pushing on this! For `tractor_load.py`, I'm loading dr3 using the 0.6.0 tagged version so no worries there. I just see modified print statements in the new version so if something breaks in `tractor_load.py` or related it'll be an easy fix.

weaverba137:
Try `2to3 -p`.  That treats print as a function, so I think that it won't insert the parentheses if they're already there.  Your mileage may vary.

geordie666:
I backtracked on SB's 2->3 requests:

(1) added back in from __future__ import absolute_import, division, print_function
(2) removed extra sets of parentheses on print statements
(3) changed things like 

for key in data.keys(): 

to:

for key in data:

In addition, had to add a couple of byte-string compliant issues for select_targets to run in both Py2 and Py3:

In psflike in cuts.py:

#psflike = ((psftype == 'PSF') | (psftype == 'PSF '))                                                                                              
psflike = ((psftype == 'PSF') | (psftype == 'PSF ') |(psftype == b'PSF') | (psftype == b'PSF '))

and in write_targets in io.py:
#hdr.add_record(dict(name='DEPVER01', value=gitversion(), comment='git revision'))
hdr.add_record(dict(name='DEPVER01', value=gitversion().decode('utf-8'), comment='git revision'))

These can probably be made more eloquent with a rewrite of io.py and/or a new version of fitsio.

Should be ready for merging before I move on to checking unit tests in Py3.

bin/tractor_load.py
```diff
@@ -38,21 +38,21 @@ def write_index_cluster_files(schema,table=None):
...
     rem_if_exists(outname)
     fin=open(outname,'w')
-    for key in list(indexes.keys()): 
+    for key in indexes: 
         for coln in indexes[key]: 
             if coln == 'radec': fin.write('CREATE INDEX q3c_%s_idx ON %s (q3c_ang2ipix(ra, dec));\n' % (key,key))
             else: fin.write('CREATE INDEX %s_%s_idx ON %s (%s);\n' % (key,coln,key,coln)) 
     fin.close()   
-    print(("wrote index file: %s " % outname))
+    print("wrote index file: %s " % outname)
```
weaverba137: I know this is much more work, and probably a separate PR, but you should really be using `format()`.  In this case `"wrote index file: {0}".format(outname)`

sbailey:
Forgot about this one.  Merging now, and then future PRs can deal with any updates necessary for py3 itself, updating to `.format()` etc.

PR #63, opened 2016-09-08
add morphology cuts

Cyeche:
Updated the QSO cut selection, in particular add a tighter cut on morphology

### DIFF
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -176,7 +176,7 @@ def isBGS(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None, objtype=
     return bgs
 
 def isQSO(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None, objtype=None,
-          wise_snr=None, primary=None):
+          wise_snr=None, deltaChi2=None, primary=None):
     """Target Definition of QSO. Returning a boolean array.
 
     Args:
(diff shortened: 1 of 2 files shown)


Cyeche:
A link to the Nathalie's presentation explaining the new color cut selection
https://desi.lbl.gov/trac/attachment/wiki/TargetSelectionWG/TSTeleconMinutes/DESITS160825/16QSOSel_Aug.pdf

py/desitarget/cuts.py
```diff
@@ -228,6 +228,7 @@ def isQSO(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None, objtype=
 
     if objtype is not None:
         qso &= psflike(objtype)
+        qso &= deltaChi2>40.
```
moustakas: The delta-chi2 cut should be set off in its own ```if``` statement:
```
if objtype is not None:
    qso &= psflike(objtype)
if deltaChi2 is not None:
    qso &= deltaChi2>40.
````

moustakas:
@Cyeche Thank you for the changes to the code and the pull request, but in general before merging you should give others an opportunity to review and comment on your changes.  Other than the one comment I made above, the code changes look fine to me.  But it would be good to also have @geordie666 or @nathalieGit take a look and sign off.  

The only thing missing in terms of code is a unit test in ```desitarget.test.test_cuts``` which tests the delta-chi2 input variable (for example, passing an array of delta-chi2 spanning values and ensuring the right objects are returned).  Adding a test to ```desitarget.test.test_io``` would also be valuable.

Finally you also need to update the wiki page which documents our nominal target selection cuts:
  https://desi.lbl.gov/trac/wiki/TargetSelectionWG/TargetSelection

Cyeche:
Sorry for that… I was too fast in the merging. I'm a svn guy ;-). 

Christophe.

PR #64, opened 2016-09-09
add tests

Cyeche:
Add tests proposed by John and cut on deltachi2 can be optional


sbailey:
This looks fine to me.  @moustakas et al?

I'll raise the nit-picking comment that `deltachi2` would match our style closer than the C++-ish `deltaChi2`, but I'm ok with it as-is and will defer to @moustakas / @geordie666 for how consistently they want to enforce the interface.

PR #66, opened 2016-09-15
Add bdt4 qso

Cyeche:
Context:
The BDT selection has been implemented in the cuts.py
select_targets  module was run over stripe 82  with a preliminary version of dr3 sweep. The results (see attached plots) look as they were expected: an improvement with BDT for faint and high-z QSOs.
It was run over whole dr2 sweep without problem. The target density for EDR region for BDT is ~260 deg^-2.

Technical information:
- The new version of cuts.py required the scikit-learn package which is by default in anaconda but not in the "mini" anaconda used by DESI
- An argument is added in select_targets module allowing to switch from color cut selection to BDT selection. The default is BDT selection now.

Christophe. 

completness_desi_target_dr3_vs_redshift
completness_desi_target_dr3_vs_rmag

### DIFF
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -3,6 +3,8 @@ from time import time
 import os.path
 import numpy as np
 from astropy.table import Table, Row
+from pkg_resources import resource_filename
+from sklearn.externals import joblib
 
 from desitarget.internal import sharedmem
 import desitarget.targets
@@ -236,6 +238,89 @@ def isQSO(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None, objtype=
         
     return qso
 
+
+def isQSO_BDT(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None, objtype=None,
+         deltaChi2=None, primary=None):
+    """Target Definition of QSO using a BDT. Returning a boolean array.
+
+    Args:
+        gflux, rflux, zflux, w1flux, w2flux: array_like
+            The flux in nano-maggies of g, r, z, W1, and W2 bands.
+        objtype: array_like or None
+            If given, the TYPE column of the Tractor catalogue.
+        deltaChi2: array_like or None
+             If given, difference of chi2 bteween PSF and SIMP morphology
+        primary: array_like or None
+            If given, the BRICK_PRIMARY column of the catalogue.
+
+    Returns:
+        mask : array_like. True if and only the object is a QSO
+            target.
+
+    """
+    #----- Quasars
+    if primary is None:
+        primary = np.ones_like(gflux, dtype='?')
+
+    # build variables for BDT    
+    nfeatures=11 # number of variables in BDT
+    nbEntries=len(rflux)
+    colors, r = getColors(nbEntries,nfeatures,gflux,rflux,zflux,w1flux,w2flux)
+  
+    #Load BDT
+    pathToBDT = resource_filename('desitarget', "data/bdt_model_dr3/bdt.pkl") 
+    # Compute BDT probability
+    bdt = joblib.load(pathToBDT) 
+    objects_bdt = bdt.predict_proba(colors)
+    prob = objects_bdt[:,1]
+
+    #define pcut, relaxed cut for faint objects
+    pcut = np.where(r>20.0,0.95 - (r-20.0)*0.08,0.95)
+
+    qso = primary.copy()
+    qso &= r<22.7    # r<22.7
+  
+    if objtype is not None:
+        qso &= psflike(objtype)
+
+    if deltaChi2 is not None:
+        qso &= deltaChi2>40.
+
+    qso &= prob>pcut
+        
+    return qso
+
+
+def getColors(nbEntries,nfeatures,gflux,rflux,zflux,w1flux,w2flux):
+ 
+    limitInf=1.e-04
+    gflux = gflux.clip(limitInf)
+    rflux = rflux.clip(limitInf)
+    zflux = zflux.clip(limitInf)
+    w1flux = w1flux.clip(limitInf)
+    w2flux = w2flux.clip(limitInf)
+
+    g=np.where( gflux>limitInf,22.5-2.5*np.log10(gflux), 0.)
+    r=np.where( rflux>limitInf,22.5-2.5*np.log10(rflux), 0.)
+    z=np.where( zflux>limitInf,22.5-2.5*np.log10(zflux), 0.)
+    W1=np.where( w1flux>limitInf, 22.5-2.5*np.log10(w1flux), 0.)
+    W2=np.where( w2flux>limitInf, 22.5-2.5*np.log10(w2flux), 0.)
+
+    colors  = np.zeros((nbEntries,nfeatures))
+    colors[:,0]=g-r
+    colors[:,1]=r-z
+    colors[:,2]=g-z
+    colors[:,3]=g-W1
+    colors[:,4]=r-W1
+    colors[:,5]=z-W1
+    colors[:,6]=g-W2
+    colors[:,7]=r-W2
+    colors[:,8]=z-W2
+    colors[:,9]=W1-W2
+    colors[:,10]=r
+
+    return colors, r
+
 def _is_row(table):
     '''Return True/False if this is a row of a table instead of a full table
     
@@ -342,8 +427,13 @@ def apply_cuts(objects):
     elg = isELG(primary=primary, zflux=zflux, rflux=rflux, gflux=gflux)
 
     bgs = isBGS(primary=primary, rflux=rflux, objtype=objtype)
-
-    qso = isQSO(primary=primary, zflux=zflux, rflux=rflux, gflux=gflux,
+    
+    useBDT = True
+    if useBDT :
+         qso = isQSO_BDT(primary=primary, zflux=zflux, rflux=rflux, gflux=gflux,
+                w1flux=w1flux, w2flux=w2flux, deltaChi2=deltaChi2, objtype=objtype)
+    else :
+        qso = isQSO(primary=primary, zflux=zflux, rflux=rflux, gflux=gflux,
                 w1flux=w1flux, w2flux=w2flux, deltaChi2=deltaChi2, objtype=objtype,
                 wise_snr=wise_snr)
 
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -4,7 +4,6 @@ import os.path
 import numpy as np
 from astropy.table import Table, Row
 from pkg_resources import resource_filename
-from sklearn.externals import joblib
 
 from desitarget.internal import sharedmem
 import desitarget.targets
@@ -278,6 +277,7 @@ def isQSO_BDT(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None, objt
     #Load BDT
     pathToBDT = resource_filename('desitarget', "data/bdt_model_dr3/bdt.pkl") 
     # Compute BDT probability
+    from sklearn.externals import joblib
     bdt = joblib.load(pathToBDT) 
 #    objects_bdt = bdt.predict_proba(colors)
     if (len(colorsReduced)>0) :
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -388,7 +388,7 @@ def unextinct_fluxes(objects):
     else:
         return result
 
-def apply_cuts(objects):
+def apply_cuts(objects,QSOSelection):
     """Perform target selection on objects, returning target mask arrays
 
     Args:
@@ -448,14 +448,13 @@ def apply_cuts(objects):
 
     bgs = isBGS(primary=primary, rflux=rflux, objtype=objtype)
     
-    useBDT = True
-    if useBDT :
-         qso = isQSO_BDT(primary=primary, zflux=zflux, rflux=rflux, gflux=gflux,
-                w1flux=w1flux, w2flux=w2flux, deltaChi2=deltaChi2, objtype=objtype)
-    else :
+    if QSOSelection==0 :
         qso = isQSO(primary=primary, zflux=zflux, rflux=rflux, gflux=gflux,
                 w1flux=w1flux, w2flux=w2flux, deltaChi2=deltaChi2, objtype=objtype,
                 wise_snr=wise_snr)
+    else :    
+        qso = isQSO_BDT(primary=primary, zflux=zflux, rflux=rflux, gflux=gflux,
+                w1flux=w1flux, w2flux=w2flux, deltaChi2=deltaChi2, objtype=objtype)
 
     #----- Standard stars
     fstd = isFSTD_colors(primary=primary, zflux=zflux, rflux=rflux, gflux=gflux)
@@ -503,7 +502,7 @@ def apply_cuts(objects):
 
     return desi_target, bgs_target, mws_target
 
-def select_targets(infiles, numproc=4, verbose=False):
+def select_targets(infiles, numproc=4, verbose=False, QSOSelection=1):
     """
     Process input files in parallel to select targets
     
@@ -537,7 +536,7 @@ def select_targets(infiles, numproc=4, verbose=False):
         '''Returns targets in filename that pass the cuts'''
         from desitarget import io
         objects = io.read_tractor(filename)
-        desi_target, bgs_target, mws_target = apply_cuts(objects)
+        desi_target, bgs_target, mws_target = apply_cuts(objects,QSOSelection)
         
         #- desi_target includes BGS_ANY and MWS_ANY, so we can filter just
         #- on desi_target != 0
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -234,7 +234,7 @@ def isQSO(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None, objtype=
 
     if deltaChi2 is not None:
         qso &= deltaChi2>40.
-        
+
     return qso
 
 
@@ -263,7 +263,9 @@ def isQSO_BDT(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None, objt
 
     # build variables for BDT    
     nfeatures=11 # number of variables in BDT
-    nbEntries=len(rflux)
+#    nbEntries=len(rflux)
+    nbEntries=rflux.size
+#    print 'length =',len(rflux),' v2 =',rflux.size
     colors, r, DECaLSOK = getColors(nbEntries,nfeatures,gflux,rflux,zflux,w1flux,w2flux)
 
     #Preselection to speed up the process, store the indexes
@@ -271,25 +273,24 @@ def isQSO_BDT(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None, objt
     preSelection = (r<rMax) &  psflike(objtype) & DECaLSOK 
     colorsCopy = colors.copy()
     colorsReduced = colorsCopy[preSelection]
-    colorsIndex =  np.arange(0,nbEntries,dtype=np.int)
+    colorsIndex =  np.arange(0,nbEntries,dtype=np.int64)
     colorsReducedIndex =  colorsIndex[preSelection]
-  
+
     #Load BDT
     pathToBDT = resource_filename('desitarget', "data/bdt_model_dr3/bdt.pkl") 
     # Compute BDT probability
     from sklearn.externals import joblib
     bdt = joblib.load(pathToBDT) 
-#    objects_bdt = bdt.predict_proba(colors)
-    if (len(colorsReduced)>0) :
+    prob = np.zeros(nbEntries)     
+
+    if (colorsReducedIndex.any()) :
         objects_bdt = bdt.predict_proba(colorsReduced)
-    
-    # add BDT probability
-    prob = np.zeros(nbEntries)
-    j=0
-    for i in colorsReducedIndex :
-         prob[i]=objects_bdt[j,1]
-         j += 1
-#    prob = objects_bdt[:,1]
+        # add BDT probability to preselected objects
+        j=0
+        for i in colorsReducedIndex :
+            prob[i]=objects_bdt[j,1]
+            j += 1
+
 
     #define pcut, relaxed cut for faint objects
     pcut = np.where(r>20.0,0.95 - (r-20.0)*0.08,0.95)
@@ -304,7 +305,10 @@ def isQSO_BDT(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None, objt
     if deltaChi2 is not None:
         qso &= deltaChi2>30.
 
-    qso &= prob>pcut
+    if nbEntries==1 : #to pass one test...
+        qso &= prob[0]>pcut
+    else :
+        qso &= prob>pcut        
         
     return qso
 
@@ -388,7 +392,7 @@ def unextinct_fluxes(objects):
     else:
         return result
 
-def apply_cuts(objects,QSOSelection):
+def apply_cuts(objects,qso_selection='bdt'):
     """Perform target selection on objects, returning target mask arrays
 
     Args:
@@ -448,7 +452,7 @@ def apply_cuts(objects,QSOSelection):
 
     bgs = isBGS(primary=primary, rflux=rflux, objtype=objtype)
     
-    if QSOSelection==0 :
+    if qso_selection=='colorcuts' :
         qso = isQSO(primary=primary, zflux=zflux, rflux=rflux, gflux=gflux,
                 w1flux=w1flux, w2flux=w2flux, deltaChi2=deltaChi2, objtype=objtype,
                 wise_snr=wise_snr)
@@ -502,10 +506,11 @@ def apply_cuts(objects,QSOSelection):
 
     return desi_target, bgs_target, mws_target
 
-def select_targets(infiles, numproc=4, verbose=False, QSOSelection=1):
+def select_targets(infiles, numproc=4, verbose=False, qso_selection='bdt'):
     """
     Process input files in parallel to select targets
     
+
     Args:
         infiles: list of input filenames (tractor or sweep files),
             OR a single filename
@@ -536,7 +541,7 @@ def select_targets(infiles, numproc=4, verbose=False, QSOSelection=1):
         '''Returns targets in filename that pass the cuts'''
         from desitarget import io
         objects = io.read_tractor(filename)
-        desi_target, bgs_target, mws_target = apply_cuts(objects,QSOSelection)
+        desi_target, bgs_target, mws_target = apply_cuts(objects,qso_selection)
         
         #- desi_target includes BGS_ANY and MWS_ANY, so we can filter just
         #- on desi_target != 0
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -1,6 +1,7 @@
 import warnings
 from time import time
 import os.path
+import sys
 import numpy as np
 from astropy.table import Table, Row
 from pkg_resources import resource_filename
@@ -238,9 +239,9 @@ def isQSO(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None, objtype=
     return qso
 
 
-def isQSO_BDT(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None, objtype=None,
+def isQSO_MVA(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None, objtype=None,
          deltaChi2=None, primary=None):
-    """Target Definition of QSO using a BDT. Returning a boolean array.
+    """Target Definition of QSO using a MultiVariate Analysis (MVA). Returning a boolean array.
 
     Args:
         gflux, rflux, zflux, w1flux, w2flux: array_like
@@ -261,11 +262,9 @@ def isQSO_BDT(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None, objt
     if primary is None:
         primary = np.ones_like(gflux, dtype='?')
 
-    # build variables for BDT    
-    nfeatures=11 # number of variables in BDT
-#    nbEntries=len(rflux)
+    # build variables for MVA    
+    nfeatures=11 # number of variables in MVA
     nbEntries=rflux.size
-#    print 'length =',len(rflux),' v2 =',rflux.size
     colors, r, DECaLSOK = getColors(nbEntries,nfeatures,gflux,rflux,zflux,w1flux,w2flux)
 
     #Preselection to speed up the process, store the indexes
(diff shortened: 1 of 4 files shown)


sbailey:
This brings in a new dependency on scikit-learn, which is breaking the travis tests.  We'll want to:
* install scikit-learn for the travis tests
* isolate the import statements so that you can use the rest of desitarget without requiring scikit-learn unless you actually need the BDT QSO selection
* make sure that we have unit tests that actually cover the new code

dkirkby:
Does the new BDT selection actually require all ~800 of the new binary .npy files?

Cyeche:
Yes...

Cyeche:
I will check…

sbailey:
I'm concerned about the structure of having a command line option change the meaning of a targeting bit.  The definition of the bits will evolve as we adjust the cuts/methods from one version of the code to another but at least then the target bit + code version uniquely defines the meaning of the bit, and presumably at some point pre-survey we freeze the meaning of the bit and don't allow any more code changes (or we cut a new targeting if we do need another cut).  But having a command line option change the meaning of what is written smells dangerous for traceability.

The npy files are just keeping one numpy array each?  Despite their similar names, it seems that some are tiny arrays, others are larger arrays, and others are table-like objects.  Randomly grabbing one and gzipping it had a factor of 6.6x compression which is worth considering.  Could these be stored in a single FITS or hdf5 file instead for easier data management?  Do they have any meaning outside of some internal scikit-learn data format?  i.e. could one imagine wanting to use them for anything outside of python?

moustakas:
I won't be able to fully review this for a while, but regarding the ~800 files and @sbailey's comments, is there any way to adopt the relevant piece of I/O code in the scikit-learn BDT package to read and correctly unpack the necessary data?  For example, in ```desisim.templates``` @dkirkby helped adopt the Gaussian mixture model routine from scikit-learn which prevented us from having to take on that (big) dependency.  Just a thought.

Cyeche:
Hi,
> I'm concerned about the structure of having a command line option change the meaning of a targeting bit. The definition of the bits will evolve as we adjust the cuts/methods from one version of the code to another but at least then the target bit + code version uniquely defines the meaning of the bit, and presumably at some point pre-survey we freeze the meaning of the bit and don't allow any more code changes (or we cut a new targeting if we do need another cut). But having a command line option change the meaning of what is written smells dangerous for traceability.
> 
> 

I agree but the issue raised, is more general than just BDT or Color cut selections.

1) As a user of target selection  code (i.e not a developer), I need a mechanism which allows me to use the algorithm that I want in order to be able to compare the different algorithms.
2) As a provider of the target list to the mountain, I need to keep track of the algorithm used to define the final list.

For 1) I pass the request of the selection through one argument of the  module but any other method is fine with me. 

For 2) we can write it in the header of the fits file containing the target list. 


> The npy files are just keeping one numpy array each? Despite their similar names, it seems that some are tiny arrays, others are larger arrays, and others are table-like objects. Randomly grabbing one and gzipping it had a factor of 6.6x compression which is worth considering. Could these be stored in a single FITS or hdf5 file instead for easier data management? Do they have any meaning outside of some internal scikit-learn data format? i.e. could one imagine wanting to use them for anything outside of python?
> 
> 
> 
I'm not convinced that it is a real issue but I will try to find a better solution than what it is currently used. 

Christophe.

geordie666:
On Thu, Sep 15, 2016 at 8:01 PM, Cyeche <notifications@github.com> wrote:

> Hi,
> > I'm concerned about the structure of having a command line option change
> the meaning of a targeting bit. The definition of the bits will evolve as
> we adjust the cuts/methods from one version of the code to another but at
> least then the target bit + code version uniquely defines the meaning of
> the bit, and presumably at some point pre-survey we freeze the meaning of
> the bit and don't allow any more code changes (or we cut a new targeting if
> we do need another cut). But having a command line option change the
> meaning of what is written smells dangerous for traceability.
> >
> >
>
> I agree but the issue raised, is more general than just BDT or Color cut
> selections.
>
> 1) As a user of target selection code (i.e not a developer), I need a
> mechanism which allows me to use the algorithm that I want in order to be
> able to compare the different algorithms.


This is simple, I think...we would just need a new set of bits that can be
messed around with as much as you like but have no actual meaning for
official survey runs. Something like "DESI_TARGET_TEST" or
"DESI_TARGET_SANDBOX". At that point, cutting a new set of targets is no
different than just running a series of test selections. Those bits would
just have no meaning for traceability, any more than messing around with
color cuts off-record on your private machine is traceable.

Probably the safest way to do that is to have an additional arm of the code
that is all just for testing purposes and writes "TEST" bits. This isn't
the most compact way of writing the code, but it prevents somebody
forgetting to pass a "test" flag or option.

In general I agree with Stephen about traceability...at least in that the
"official" part of the desitarget code should be used for official target
production runs and testing selections should occur somewhere else, or in
some other way.

Ad.

sbailey:
FYI I moved the sklearn import into the `isQSO_BDT()` function so that you only need to have sklearn installed if you are calling `isQSO_BDT()`.

py/desitarget/cuts.py
```diff
@@ -283,7 +388,7 @@ def unextinct_fluxes(objects):
     else:
         return result
 
-def apply_cuts(objects):
+def apply_cuts(objects,QSOSelection):
```
sbailey: Adding `QSOSelection` without a default has broken the unit tests.  Whenever contributing new code, please always run "python setup.py test" from the top-level directory to make sure that tests pass before pushing them to github.

Please add a description of the valid `QSOSelection` values to the docstring (the triple-quoted string just after the function definition).  From a quick look, I think you implemented this as an integer.  Let's use a more user-friendly string, e.g. QSOSelection='colorcuts' or QSOSelection='bdt' .

Let's also add a default for `QSOSelection` to `apply_cuts` so that `apply_cuts(objects)` will still work, using whatever the current default choice should be.  e.g.
```python
def apply_cuts(objects, QSOSelection='colorcuts')
```
That would fix the failing unit tests, though another unit test should also be added to explicitly test the BDT method too.

I would prefer "qso_selection" or something in lower case to match our python style.  Generally starting with a Capital letter is reserved for classes, not variables.

Cyeche:
BTW, Stephen have already changed it or do you like that I change it in the current version of the branch that I 'm using?

Christophe.

sbailey:
I only changed the sklearn import statement so that it would be imported only if needed.  I did *not* change:
* Making QSOSelection optional, documenting it, and making it a user-friendly string instead of an integer
* Adding unit tests of the BDT code
* Adding other targeting bits (test or otherwise).

Please proceed with those ("git pull origin addBDT4QSO" first on your local copy to get my minor change)

I like the idea of test bits to be used more informally than the official targeting bits, but I still think that the distinction between BDT vs. color cuts is big and important enough that we might want to have official bits and cuts for both in the end.

Cyeche:
Hi,

I have implemented all the Stephen's suggestions (except the target bits about versioning which is a higher level development).

BTW I have protected the division in unextinct_fluxes by zero. Not the sure that the default (zero) when the flux is divided by zero, is what everybody wants….

Christophe.

sbailey:
Thanks for the updates; test now pass too.

I'd still like to get separate bits for BDT vs. colorcuts.  We might end up targeting the superset of them, while allowing analyses to choose if they want to use the more optimal BDT selection or the more conservative but easier to understand systematics of the color cuts.  At minimum the output file needs to include the provenance for what selection was used to set the bit (the code version used to be sufficient for that, but now it requires code version + knowledge of command line options used).

A few items to note:
* `np.divide(flux, mw_transmission, where=(mw_transmission!=0))`
  * the `where` option to `np.divide` is a handy trick to know, but appears to be undocumented.  How did you learn of it?  Are we sure it is fully supported for the future?  I might want to use that a lot more in other places where we have messy code to avoid dividing by 0.
  * do we actually have cases of Milky Way transmission set to 0 within our imaging footprint?  That sounds like an error in the inputs and we may want to actually raise and error instead of silently work around it as if targets have 0 flux at those locations.
* I'm not thrilled about the 800 tiny numpy-specific format files, but that seems to be a "feature" of scikit-learn.  110 MB is 10% of the loosely enforced github repo size recommendations, which doesn't give us a lot of breathing room for updating versions.  Keeping these as a compressed tarball would save a factor of 5 in size and only takes a few seconds to pack/unpack, at the cost of that extra step to remember to do (best would be to automate that at install time).  We could also consider other forms of I/O wrappers as suggested by @moustakas.  Adding these files is a kind of change that is hard to back out of in git, so let's get this right from the start.
* this introduces a dependency upon scikit-learn.  If BDT cuts are included by default, this also becomes a required dependency and not just an optional dependency.  I think that is ok but we should be explicit about that.

@moustakas?  @geordie666?  Others?

Cyeche:
> Thanks for the updates; test now pass too.
> 
> I'd still like to get separate bits for BDT vs. colorcuts. We might end up targeting the superset of them, while allowing analyses to choose if they want to use the more optimal BDT selection or the more conservative but easier to understand systematics of the color cuts. At minimum the output file needs to include the provenance for what selection was used to set the bit (the code version used to be sufficient for that, but now it requires code version + knowledge of command line options used).
> 
> 

I don't think that we want a separate bit in desi_target, one for colorcuts and one for bdt option. This is the responsibility of the TS WG to decide the selection that we use at the end.  
However, I agree we need to keep the information at the level of the produced fits file (in the header for instance). I think that it is a higher level code and this is not just for QSO selection. We need to discuss this point and we have to define what we need to keep as information (the algorithms used, the photometry used (DECaLS, BASS…), the photometry release dr3,…drx, ..)  


> A few items to note:
> 
> np.divide(flux, mw_transmission, where=(mw_transmission!=0))
> the where option to np.divide is a handy trick to know, but appears to be undocumented. How did you learn of it? Are we sure it is fully supported for the future? I might want to use that a lot more in other places where we have messy code to avoid dividing by 0.
It is a fast solution, to allow me to run dr3. I agree  that it hides a potential problem, that extinction corrections are missing in the file. We can change the logic, here if we divide by zero we return zero. Instead we can return the uncorrected flux. Anyway the issue is still the same, we hide a potential problem.



> do we actually have cases of Milky Way transmission set to 0 within our imaging footprint? That sounds like an error in the inputs and we may want to actually raise and error instead of silently work around it as if targets have 0 flux at those locations.

> 
> I'm not thrilled about the 800 tiny numpy-specific format files, but that seems to be a "feature" of scikit-learn. 110 MB is 10% of the loosely enforced github repo size recommendations, which doesn't give us a lot of breathing room for updating versions. Keeping these as a compressed tarball would save a factor of 5 in size and only takes a few seconds to pack/unpack, at the cost of that extra step to remember to do (best would be to automate that at install time). We could also consider other forms of I/O wrappers as suggested by @moustakas. Adding these files is a kind of change that is hard to back out of in git, so let's get this right from the start.
The easiest thing will be that git does it for us ;-)… I don't know if it is possible.  Another option would be to store a compressed file in git and to gunzip it when we type in  desi_target  "python setup.py develop" 



> this introduces a dependency upon scikit-learn. If BDT cuts are included by default, this also becomes a required dependency and not just an optional dependency. I think that is ok but we should be explicit about that.

Christophe.

weaverba137:
Let me voice my concern about these files.  Both the number and the type of files deviates strongly from all other forms of data that we have in our repositories.  I myself would not approve this PR until these files are wrapped up into a FITS or HDF5 file.  There is *absolutely* no reason why this not possible.  It is just data after all.

geordie666:
I'm going to:

(1) Agree with @Cyeche  on the bits. The bit structure should not be experimental, it should be final. We don't want to track every possible trial color-cut that someone is trying out. If needed, for experimentation, there should be a completely separate set of trial bits. 

Admittedly, under a framework where we're definitely having 2 bits in the data model for a given selection ("loose cut" and "tight cut") and will always intend to have 2 bits, then I'm happy to be less strict. But I don't want to slip into a regime where each individual target selection has N multiple bits because N different people want to use N different selections.

@sbailey  and I can hurl abuse at each other over this next week. 

(2) Agree with @weaverba137 on the file format. If there's a way around numpy-specific format files we should pursue it.

sbailey:
bits: I agree that we shouldn't generate a new bit for every experimental change in target selection.  I'm advocating that the *final* DESI target selection should include a conservative color cuts bit and a more aggressive BDT bit (or other type of machine learning type bit) somewhat analogous to BOSS core vs. bonus.  There could plausibly also be a bit for variability selected QSOs and loose vs. tight versions of these bits included in the *final* set of bits.

One bit or many, I will absolutely insist that the target catalog file contain sufficient provenance to know what cuts/methods were applied.

BDT training files: this will take more work but the majority view tilts towards repackaging these, though for different reasons.  The primary reason not to repackage is that using the formats as is "just works" without any extra futzing.  Do we have a volunteer from the "repackage" camp to help with this?

@Cyeche where is the code and training files that generated these in the first place?

Cyeche:
Another simple option is that we store in git a tar gziped file and that for the call of desi_target, the file is untared and that's all.

Nothing to develop, and it takes time just for the first call, after we have just to check that the directory is created….

weaverba137:
A tar file solution does not solve the problem of a platform-independent data format.  I hereby volunteer to do the repackaging.  It will probably take me a couple of hours.

weaverba137:
@sbailey are you saying that the code that generated the files is not part of this PR?

Cyeche:
Ok thanks.

weaverba137:
@Cyeche, since you're happy to have me repackage, can you point me to *exactly* where the original files were generated?

sbailey:
The contents of these files are so specific to the details of the scikit-learn implementation of boosted decision trees that I'm not sure there is much value to making them platform/language independent, since it is hard to imagine using them except via python/scikit-learn.  If repackaging them is easy for both using them and for updating them the next time the BDT is retrained, then ok.  But if this turns out to take more time than expected, please loop back for discussion.  I'm more concerned about their number and size and playing nicely with git, thus I'd be fine with the auto-unpacking tarball option if needed.

Caveat: if numpy array dump files have endianness or related issues that would make them unusable under numpy 2.2 in 2020 on some crazy new NERSC machine, then wrappering in a more archival format is more important.

@Cyeche can confirm, but I think the code for generating those files is elsewhere.

Cyeche:
To store the BDT

from sklearn import cross_validation,datasets,svm
from sklearn.externals import joblib
from sklearn.ensemble import RandomForestClassifier

modelDir='….'
 bdt = RandomForestClassifier(200)
 joblib.dump(bdt, modelDir+'bdt.pkl') 


To load the BDT
 bdt = joblib.load(modelDir+'bdt.pkl') 

All the files are so far in
desitarget/py/desitarget/data/bdt_model_dr3

The loading of the BDT is done in:
desitarget/py/desitarget/cuts.py

Attached the script to generate the BDT.

Christophe.

Cyeche:
BTW, this takes less than 1 second on edison… And we can do just once….


import os
os.chdir('./')
os.system("tar -xzf bdt_model_dr3.tar.gz -C test")




> The contents of these files are so specific to the details of the scikit-learn implementation of boosted decision trees that I'm not sure there is much value to making them platform/language independent, since it is hard to imagine using them except via python/scikit-learn. If repackaging them is easy for both using them and for updating them the next time the BDT is retrained, then ok. But if this turns out to take more time than expected, please loop back for discussion. I'm more concerned about their number and size and playing nicely with git, thus I'd be fine with the auto-unpacking tarball option if needed.
> 
>

weaverba137:
NO!  That is not how data files are stored in Python.  I'm sorry, but that is not a viable solution.  Please leave this to me.  I *will* come up with a solution.

Cyeche:
And I have forgotten 

du -m bdt_model_dr3.tar.gz gives 10 M

It is not really a large file, less than any ppt presentation now….



> The contents of these files are so specific to the details of the scikit-learn implementation of boosted decision trees that I'm not sure there is much value to making them platform/language independent, since it is hard to imagine using them except via python/scikit-learn. If repackaging them is easy for both using them and for updating them the next time the BDT is retrained, then ok. But if this turns out to take more time than expected, please loop back for discussion. I'm more concerned about their number and size and playing nicely with git, thus I'd be fine with the auto-unpacking tarball option if needed.
> 
>

sbailey:
Another housecleaning issue: if you are using a random forest classifier, we shouldn't call this option BDT (Boosted Decision Tree).  They are related but different algorithms.

weaverba137:
```python
from sklearn import cross_validation,datasets,svm
from sklearn.externals import joblib
from sklearn.ensemble import RandomForestClassifier

modelDir='….'
bdt = RandomForestClassifier(200)
joblib.dump(bdt, modelDir+'bdt.pkl')
```
I'm quite confused here.  This only references a pickle file.  Also, is `'....'` a valid directory path?

Cyeche:
Yes it is a path for instance in cuts.py there is:

    pathToBDT = resource_filename('desitarget', "data/bdt_model_dr3/bdt.pkl") 
    from sklearn.externals import joblib
    bdt = joblib.load(pathToBDT)

weaverba137:
I'm just going to have to do this myself then.

weaverba137:
The documentation for `sklearn.externals.joblib.dump()` says "Persist an arbitrary Python object into one file."  Note the "one file".  Not 800 files.  Where do the other 800 files come from?

Cyeche:
I have already checked that all the files are required when the bdt is loaded… I haven't dug more ….

weaverba137:
According to http://scikit-learn.org/0.17/modules/model_persistence.html#security-maintainability-limitations

> Models saved in one version of scikit-learn might not load in another version.

Since the scikit developers themselves are saying this, I think this argues very persuasively for a different serialization method.

sbailey:
Thanks for finding that "security and maintainability" section of the documentation.  scikit-learn/0.17.1 is almost certainly not going to be anywhere near current when we do final target selection (in 2019?).  Writing our own I/O wrappers won't magically give us better maintainability across versions (e.g. if the internal way of storing the tree changes so that it doesn't map onto our I/O formats anymore), but at least it does give us more control if we need to update a reader to work with an older writer.  This is a pretty fundamental issue for target selection traceability, since re-running the BDT (or RF) on mock datasets will be very important for understanding completeness, well after we "finished" with running target selection.  I'll also quote the rest of that section that Ben found:

> In order to rebuild a similar model with future versions of scikit-learn, additional metadata should be saved along the pickled model:
  * The training data, e.g. a reference to a immutable snapshot
  * The python source code used to generate the model
  * The versions of scikit-learn and its dependencies
  * The cross validation score obtained on the training data
>
> This should make it possible to check that the cross-validation score is in the same range as before.
If you want to know more about these issues and explore other possible serialization methods, please refer to this talk by Alex Gaynor.

weaverba137:
The title of the Alex Gaynor talk is "Pickles are for Delis, not Software."  I'm inclined to agree.

sbailey:
Documenting an in-person conversation with Christophe:

Before merging this PR:
  * [ ] give @weaverba137 a few days to explore/implement/recommend alternate I/O models; otherwise we'll proceed with the compressed tarball version of the existing files and save the refactoring of the I/O for when scikit-learn actually makes a change that breaks it or when we actually need to read these files from outside of python
  * [ ] rename "bdt" to "randomforest" or similar, since this isn't a BDT.  Maybe there will be additional QSOSelection options in the future including "adaboost" or "bdt" or "neuralnetwork" or ...
  * [ ] Add what QSOSelection option was used to the output file header.  I haven't given up on having a small number of QSO targeting bits for color cuts vs. random forests, but in the meantime at minimum the file must contain the provenance of what method was used.

Short term, but not directly blocking for this PR:
  * [ ] organize the training and validation files and put those on NERSC; ask @weaverba137 for where they best go and how to back them up.
  * [ ] provide code in `desitarget/py/desitarget/training/randomforest.py` (or similar) that converts the training data into the files used in `desitarget/py/desitarget/data/bdt_model_dr3` (which will also be renamed to not "bdt").
  * [ ] provide a technote or other documentation about how to use that training code
  * [ ] figure out how to set random number seeds for scikit-learn so that the training is reproducible

weaverba137:
The existing scikit-learn pickled data consists of an actual pickle file, which contains "pointers" to the NumPy-serialized data files.  I can't open the pickle file in Python 3, so we are *already* in a state where we have broken maintainability.  This could be due to str/unicode incompatibility or different pickle protocols in 2 versus 3.

```python
>>> with open('bdt.pkl') as p:
...     pickletools.dis(p)
... 
  File "<stdin>", line 2, in <module>
  File "/Users/weaver/Documents/local/products/anaconda/envs/py3/lib/python3.5/pickletools.py", line 2398, in dis
    for opcode, arg, pos in genops(pickle):
  File "/Users/weaver/Documents/local/products/anaconda/envs/py3/lib/python3.5/pickletools.py", line 2234, in _genops
    code = data.read(1)
  File "/Users/weaver/Documents/local/products/anaconda/envs/py3/lib/python3.5/codecs.py", line 321, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte
```

Cyeche:
I have used scikit-learn version corresponding to python 2 which certainly explains the issue that you observed.

weaverba137:
Documenting some further digging on the scikit-learn pickle-based data persistence:

1. The serialization is done by `sklearn.externals.joblib`.  There's lots of documentation in the `__init__.py` file: https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/externals/joblib/__init__.py
2. joblib itself is not technically part of scikit-learn.  It is an external package that scikit-learn incorporated into their own. Here's the link to the "real" joblib package: http://pythonhosted.org/joblib/
3. So not only does scikit-learn not recommend this serialization method, they don't even maintain the package which does the serialization.

In other news, I'm learning about the structure of the *.npy files.  There are 801 of them; the Random Forest contains 200 separate Decision Tree objects.  Each Decision Tree needs 4 data files to reconstruct.  Then there appears to be one small file left over, which might have a couple of parameters needed for the whole forest (talk about not being able to see the forest for the trees...).

weaverba137:
The documentation for `joblib.dump()`, http://pythonhosted.org/joblib/generated/joblib.dump.html#joblib.dump, seems to imply that if you select a *compressed* pickle file, it will save everything to one file.

weaverba137:
OK, so just by turning on compression, we can turn 802 files into one file:

```
>>> from sklearn.externals import joblib
>>> foo = joblib.load('bdt_model_dr3/bdt.pkl')
>>> foo
RandomForestClassifier(bootstrap=True, class_weight=None, criterion='gini',
            max_depth=None, max_features='auto', max_leaf_nodes=None,
            min_samples_leaf=1, min_samples_split=2,
            min_weight_fraction_leaf=0.0, n_estimators=200, n_jobs=1,
            oob_score=False, random_state=None, verbose=0,
            warm_start=False)
>>> bar = joblib.dump(foo, 'rf_model_dr3.pkl.gz', compress=9)
>>> bar
['rf_model_dr3.pkl.gz']
>>> foo2 = joblib.load('rf_model_dr3.pkl.gz')
>>> foo2
RandomForestClassifier(bootstrap=True, class_weight=None, criterion='gini',
            max_depth=None, max_features='auto', max_leaf_nodes=None,
            min_samples_leaf=1, min_samples_split=2,
            min_weight_fraction_leaf=0.0, n_estimators=200, n_jobs=1,
            oob_score=False, random_state=None, verbose=0,
            warm_start=False)
>>> foo[0]
DecisionTreeClassifier(class_weight=None, criterion='gini', max_depth=None,
            max_features='auto', max_leaf_nodes=None, min_samples_leaf=1,
            min_samples_split=2, min_weight_fraction_leaf=0.0,
            presort=False, random_state=64661933, splitter='best')
>>> foo2[0]
DecisionTreeClassifier(class_weight=None, criterion='gini', max_depth=None,
            max_features='auto', max_leaf_nodes=None, min_samples_leaf=1,
            min_samples_split=2, min_weight_fraction_leaf=0.0,
            presort=False, random_state=64661933, splitter='best')
```
Also the compressed file is 10 MB, versus 56 MB for the bdt_model_dr3 directory.  I didn't do an exact timing, but saving the compressed file took on order 1 minute, however, *reloading* the compressed file was quite fast, of order 1 second.

weaverba137:
I agree.  Let's move forward with the single, compressed file for now, also ensuring that it is created with Python 3 in mind.

For reference, here's the snippet that should be used:

```python
#
# The file should go in desitarget's data directory, use this
# or the equivalent to ensure it gets there.
#
filename = resource_filename('destarget', 'data/rf_model_dr3.pkl.gz')
dumped_files = joblib.dump(rfdata, filename, compress=9)
```

Cyeche:
Many thanks for your help.

> I agree. Let's move forward with the single, compressed file for now, also ensuring that it is created with Python 3 in mind.
> 
> For reference, here's the snippet that should be used:
> 
>

dkirkby:
Should we consider rewriting the git history to remove those 802 separate files and trim all repos by 56Mb?

weaverba137:
@dkirkby, I would support that, though I don't think I have the time to come up with the commands myself.  We should document the commands that are run.

Cyeche:
Yes I have planned to remove the 801 files and use the compressed version, in the current version of the branch that I'm using. But I'm not sure to understand what you mean by "rewriting the git history"...

dkirkby:
Everyone who checks out this package in future will always be downloading those ~800 files (even if you `git rm` them and they are never merged to the master branch) since git never forgets :-/

Cyeche:
Vive svn!

> Everyone who checks out this package in future will always be downloading those ~800 files (even if you git rm them and they are never merged to the master branch) since git never forgets :-/
> 
> —
>

sbailey:
Yes, let's get these files out of the git history.

@dkirkby can you do the git-fu to remove those from the git history?

@Cyeche after David is done, could you regenerate the single training file using python 3.5 and the I/O commands that @weaverba137 suggested, and then commit that?  You probably still have a copy of those files outside of git, but be warned that after @dkirkby is done and you do a "git pull", those files will completely disappear from your git checkout, never to be found again (with git).

If I understand correctly, this file will work only under python 3 and we will not be able to use desitarget QSO selections under python 2.7 after this merge (at least until we write our own I/O wrapper alternatives, which still sounds like a good idea given the fragility of this format).  Bold new world!

Cyeche:
> If I understand correctly, this file will work only under python 3 and we will not be able to use desitarget QSO selections under python 2.7 after this merge (at least until we write our own I/O wrapper alternatives, which still sounds like a good idea given the fragility of this format). Bold new world!

I missed this point! When do you plan to migrate to Python3?  We need this migration before implementing it.

sbailey:
The master branch is already python 3 compatible and all our other packages are too.  We are in the final steps of testing the full simulation + spectro pipeline chain on python 3.  This would be the first change that would break compatibility with python 2.7.

dkirkby:
I won't have time to research the necessary git-fu anytime soon (certainly not today) but, when I have done this before, I used [BFG](https://rtyley.github.io/bfg-repo-cleaner/), in case anyone else wants to take a shot at this.  Everything should be tested first on a fork, of course.

Cyeche:
I propose that I implement all your suggestions (except the compressed file) and then once the python 3 migration is finalized, I will add this new feature with the compressed file. 

In practice, this means that this branch (with a few improvements following your suggestions) will be merged and then I create a new branch with the compressed file.

Cyeche:
> If I understand correctly, this file will work only under python 3 and we will not be able to use desitarget QSO selections under python 2.7 after this merge (at least until we write our own I/O wrapper alternatives, which still sounds like a good idea given the fragility of this format). Bold new world!

This solution works both with python 2.7 and 3.5. The issue is that you cannot load with python3 what is compressed with python2. I would suggest an heresy. I'm pretty sure to be burnt… 

Why we don't have two compressed files? 'rf_model_dr3_python3.pkl.gz' and 'rf_model_dr3_python2.pkl.gz'  and we test with the python version?

import sys
sys.version_info[0]

weaverba137:
> This solution works both with python 2.7 and 3.5. 

I *never* said that!  Who gave you that idea?

Cyeche:
Anyway, the question is:
Are we allowed to test over the python version?

sbailey:
> This solution works both with python 2.7 and 3.5.

In the sense that the same method can be used for both versions, yes.  But a file written by one can't be used by the other.

> Why we don't have two compressed files? 'rf_model_dr3_python3.pkl.gz' and 'rf_model_dr3_python2.pkl.gz'  and we test with the python version?

That could work as a temporary measure to support both, *if* you can get the random number seed stuff sorted out such that they actually do the exact same thing and aren't use approximately equivalent but different trainings.

At some point we will officially deprecate 2.7 and proceed with code that only works under 3.5, but it would be preferable to be stabling running under 3.5 for awhile before breaking 2.7.

> Are we allowed to test over the python version?

Yes.  The master branch already does this for the Travis tests, but that occurred after you created your branch.  On my laptop with anaconda I have separate python 2.7 and 3.5 environments to be able to run separate tests.  I will merge master into this branch to get those changes so that we can test this branch on both as well.  @geordie666 will watch over my shoulder to help spread the blame if something goes wrong.

Cyeche:
> Documenting an in-person conversation with Christophe:
> 
> Before merging this PR:
> 
>  give @weaverba137 a few days to explore/implement/recommend alternate I/O models; otherwise we'll proceed with the compressed tarball version of the existing files and save the refactoring of the I/O for when scikit-learn actually makes a change that breaks it or when we actually need to read these files from outside of python

Two compressed files (10M) are implemented one for python2 and one for python3. The two random forests are exactly identical! The code tests against the python version to load the correct file.

>  rename "bdt" to "randomforest" or similar, since this isn't a BDT. Maybe there will be additional QSOSelection options in the future including "adaboost" or "bdt" or "neuralnetwork" or …
I use a generic name mva= MultiVariate Analysis. The name of the file stored contains an indication of the type of algorithm used for instance: rf_model_dr3_py2.pkl.gz  = random forest for python2 trained over dr3
If in the future we have several algorithms, I will replace "mva" by several names such as "rndf" "adaboost", "nn".... My impression is that we will finish with only one mva but I don't know which one yet… This is the reason why I chose the generic name. 

>  Add what QSOSelection option was used to the output file header. I haven't given up on having a small number of QSO targeting bits for color cuts vs. random forests, but in the meantime at minimum the file must contain the provenance of what method was used.

There is an additional word in the header of the fits file "qso-selection" which can be either "mva" or "colorcuts"

All is ready for the merging ;-)….

I have forgotten! Someone has to delete from the history, the directory data/bdt_model_dr3. I'd prefer that someone much more expert than me with git, does it.

sbailey:
Adam and I are working on this now, which will involve some git surgery.  Please no one else commit anything to this branch until you hear more from us.

moustakas:
I've been following this thread closely -- thank you to everybody for the great work and sleuthing. Simultaneously wearing my Target Selection and Data Systems hats, let me just echo @sbailey's request that we fully document the creation and analysis of the training sample both in code and text (cf the "Short term, but no directly blocking for this PR" list of tasks about 50 messages above this one.

sbailey:
Adam and I have cleaned up the 801 small files and merged this branch + master into a new branch, qso_py3, which includes all of the changes here plus the py3 updats.  I'm going to close out this PR and remove this branch; PR #80 now supersedes this one.

PR #72, opened 2016-09-21
py3 updates and MTL refactor

### LINKED ISSUE #60: Python 2->3 Upgrade
Need to rewrite for python 3.5 upgrade. ADM will do this except for the mocks.

sbailey:
This fixes #60 to make desitarget compatible with both py2.7 and py3.5.  For better or worse, it also brings in a bunch of changes from the apc_mocks branch.  We may want to remerge this branch and the apc_mocks branch before the final merge into master, but I'm opening up the PR to get some eyes on it now that tests are passing again (on my laptop at least; we'll see about Travis...)


sbailey:
For the record: "python setup.py test --coverage" on travis includes checking syntax even for lines that are never run.  The same command on my laptop runs the tests but can miss syntax errors on unexecuted lines of code.

weaverba137:
Ah, right, I think that's because coverage has to import an entire module file to determine how much is *not* covered, not just the functions that are actually used by other parts of the code.

sbailey:
Grumble.  Yesterday tests were passing using qt=4.8.7 and sip=4.18.  Today, without making any changes to our .travis.yml file, travis+conda is giving us qt=4.8.7 and sip=4.16.9 (downgraded!) instead, which is actually incompatible with qt/4.8.7 causing the sphinx build to fail:
```
RuntimeError: the sip module implements API v11.0 to v11.2 but the PyQt4.QtCore module requires API v11.3
```
Going off to our friends google and stackoverflow...

weaverba137:
Also check astropy and conda issues on GitHub.

sbailey:
It has been reported to anaconda, but no workaround identified yet:

https://github.com/ContinuumIO/anaconda-issues/issues/1090

sbailey:
Conda fixed it; restarted tests worked again without requiring changes to our .travis.yml .  In a separate email @geordie666 said "I made one minor clean-up in the py3 branch and now it's OK to merge".  In the end this PR brings in a lot of changes related to mocks and fiberassign preparation and not just py3 changes.  merging now.

PR #73, opened 2016-09-22
bug fix for DR3 db load

kaylanb:
Added a script, list_remaining_cats.py, to determine which Tractor Catalogues successfully loaded into the db and which didn't (need to be reloaded) or have not been loaded yet. Note, the poorly named "v0.6.0" is the name I chose when I checked out the commit hash for tag 0.6.0.

Only I worked on this branch, and the files it changed, so merging should be trivial.


kaylanb:
@sbailey 

From the travis failed details, I couldn't determine what was wrong. This PR adds [list_remaining_cat.py](https://github.com/desihub/desitarget/blob/v0.6.0/bin/list_remaining_cats.py) for finding the remaining tractor catalogues to load into the db, given the completed job ids, and [submit_loaddb.sh](https://github.com/desihub/desitarget/blob/v0.6.0/bin/submit_loaddb.sh) which is the bash script for submitting the db load jobs.

I'll be shocked is this isn't an easy fix for Travis. Thanks!

sbailey:
A couple issues here:
* `bin/` should be reserved for command line scripts, not admin-level SQL scripts like the new `decam_*` ones here.  Please put those in `db/` instead.
* `bin/list_remaining_cats.py` is pretty special-purpose and overly generically named when viewed in the context of DESI software as a whole (target catalogs?  imaging catalogs?  redshift catalogs?  LSS catalogs?  warm fuzzy creatures that purr?)  Let's put this in db too.
* `list_remaining_cats.py` isn't python 3 compatible (print statements, but possibly other things too)
* `submit_loaddb.sh` has your email address hard coded into it and is also very special purpose but generically named.

We're refining the guidelines as we go about where scripts like this should live, but in general:
* Let's put admin-level special purpose scripts for loading and verifying the database into `db/`; i.e. they would not be in a typical user's path when they use desitarget.
* User-facing scripts that access the db can go in bin/ (I'm not sure we have any of these yet).
* When naming scripts in bin/, remember that desitarget has to be loaded for using desisim and desispec too, so pick names that won't be confusing in other contexts (e.g. "load_zcatalog_db" vs. "load_tractor_db" but not just "loaddb")

weaverba137:
Also, command-line scripts should not end in ".py".

kaylanb:
I fixed all of the issues above in commit d324b84. Lets merge when checks pass

PR #76, opened 2016-09-23
apc_mocks into master

forero:
This PR adds files to create mock Target/Truth for the BGS/MWS program. It runs with Python3 and includes sample configuration files to run on NERSC.


py/desitarget/mock/io.py
```diff
@@ -476,7 +470,7 @@ def read_mock_bgs_mxxl_brighttime(root_mock_dir='',mock_prefix='',brickname_list
     full_data   = dict()
     n_per_file  = list()
     if len(target_list) > 0:
-        for k in target_list[0].keys():
+        for k in list(target_list[0]):
```
sbailey: minor: if `target_list[0]` is a dictionary, then `for k in target_list[0]:` iterates over the keys without requiring `list(...)` or `.keys()` and works in both python 2.7 and 3.5 without 2to3 complaining about it.

sbailey:
I made one minor stylistic comment and one minor fix to calc_numobs (code would fail with NameError if input targets didn't include DESI_TARGET, even though it included logic about `if 'DESI_TARGET' in targets...`).

OK to merge as soon as travis tests re-pass.

PR #77, opened 2016-09-27
Fixes bug in whitespace_fits_read

apcooper:
When len(data) == 2 and no header was requested, the second element of data was treated as the header; now check directly for 'header' in kwargs.

Although perhaps there was a reason for not checking explictly for the header keyword?


geordie666:
No particular reason, just sloppy coding. Your approach is more elegant.

PR #78, opened 2016-09-27
Replace astropy with fitsio

apcooper:
These changes from astropy.io.fits work locally for me, although making these fixes showed up that a full run on the actual mocks gets stuck on a problem with certain bitmask values/datatypes [here](https://github.com/desihub/desiutil/blob/4fb1b442cb4deed1f2f289fd050ee46aef34d1f7/py/desiutil/bitmask.py#L241) (something like [this issue](https://github.com/numpy/numpy/issues/2955)). That's not related to fitsio (hacked round it to confirm that the changes in this PR work) so I'll open a separate issue.

### DIFF
--- a/bin/mocks_to_fa_input
+++ b/bin/mocks_to_fa_input
@@ -17,11 +17,12 @@ import time
 import yaml
 
 import astropy
-import astropy.io.fits as fits
+import fitsio
 from   astropy.table import Table, Column
 from   desitarget import mtl
 import desitarget.mock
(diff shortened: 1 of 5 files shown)


apcooper:
That separate issue with bitmasks is #79.

rainwoodman:
For speeding up, keep an eye on dask.array. the promise is somewhat similar
to npyquery but they actually implemented multi thread + cluster scheduling
and io optimizations. It is funded.

PR #80, opened 2016-09-30
Add random forest QSO selection

sbailey:
This PR supersedes PR #66 .  It includes the random forest QSO selection from @Cyeche but also merges in the python 3 changes that were on master such that we can test on both py2.7 and py3.5 prior to merging.

In addition to merging those two branches, we also changed `qso_selection` option name from 'mva' (multi variate analysis) to 'randomforest' since this really is a random forest.  If we want a different MVA option, we should add a new option and a new function with a new descriptive name.  This avoids having a target selection file saying that the qso selection was "mva" but we couldn't remember which MVA it was...

HOWEVER: this works under scikit-learn 0.17.1, but breaks under scikit-learn 0.18 because the pickle files generated from one version don't work with the next.  The theoretical concern has been realized.  Do not merge this branch until that issue is resolved.  We will need to make a more robust file format and our own I/O wrappers (and perhaps slurp in the random forest evaluator code...)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ env:
         # These packages will always be installed.
         - CONDA_DEPENDENCIES='pyyaml qt=4'
         # These packages will only be installed if we really need them.
-        - CONDA_ALL_DEPENDENCIES='pyyaml h5py coverage==3.7.1 requests scikit-learn=0.17.1'
+        - CONDA_ALL_DEPENDENCIES='pyyaml h5py coverage==3.7.1 requests scikit-learn'
         # These packages will always be installed.
         - PIP_DEPENDENCIES=''
         # These packages will only be installed if we really need them.
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ env:
         # These packages will always be installed.
         - CONDA_DEPENDENCIES='pyyaml qt=4'
         # These packages will only be installed if we really need them.
-        - CONDA_ALL_DEPENDENCIES='pyyaml h5py coverage==3.7.1 requests'
+        - CONDA_ALL_DEPENDENCIES='pyyaml h5py coverage==3.7.1 requests scikit-learn'
         # These packages will always be installed.
         - PIP_DEPENDENCIES=''
         # These packages will only be installed if we really need them.
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ env:
         # These packages will always be installed.
         - CONDA_DEPENDENCIES='pyyaml qt=4'
         # These packages will only be installed if we really need them.
-        - CONDA_ALL_DEPENDENCIES='pyyaml h5py coverage==3.7.1 requests scikit-learn'
+        - CONDA_ALL_DEPENDENCIES='pyyaml h5py coverage==3.7.1 requests scikit-learn=0.17.1'
         # These packages will always be installed.
         - PIP_DEPENDENCIES=''
         # These packages will only be installed if we really need them.
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ env:
         # - NUMPY_VERSION=1.10
         # - SCIPY_VERSION=0.17
         - ASTROPY_VERSION=1.1.1
-        - DESIUTIL_VERSION=1.4.0
+        - DESIUTIL_VERSION=1.8.0
         - MAIN_CMD='python setup.py'
         # These packages will always be installed.
         - CONDA_DEPENDENCIES='pyyaml qt=4'
--- a/bin/qso_training
+++ b/bin/qso_training
@@ -1,0 +1,19 @@
+#!/usr/bin/env python
+
+
+import os, sys
+import numpy as np
+
+from desitarget.train.train_mva_decals import train_mva_decals
+
+from argparse import ArgumentParser
+ap = ArgumentParser()
+ap.add_argument('--step',choices={'test','train'},help="Training or test of qso TS algorithm")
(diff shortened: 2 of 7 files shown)


sbailey:
Having demonstrated that the code works with scikit-learn 0.17.1 in the initial pull request, I removed that requirement in .travis.yml such that it now installs the latest version 0.18 and runs the tests with that.  This generates a warning message
```
test_astropy_fits (desitarget.test.test_cuts.TestCuts) ... /home/travis/miniconda/envs/test/lib/python2.7/site-packages/sklearn/base.py:315: UserWarning: Trying to unpickle estimator DecisionTreeClassifier from version pre-0.18 when using version 0.18. This might lead to breaking code or invalid results. Use at your own risk.
```
and then proceeds to crash with
```
ERROR: test_astropy_fits (desitarget.test.test_cuts.TestCuts)
  File "/home/travis/build/desihub/desitarget/py/desitarget/test/test_cuts.py", line 112, in test_astropy_fits
    self._test_table_row(targets)
  File "/home/travis/build/desihub/desitarget/py/desitarget/test/test_cuts.py", line 100, in _test_table_row
    desi, bgs, mws = cuts.apply_cuts(targets)
  File "/home/travis/build/desihub/desitarget/py/desitarget/cuts.py", line 474, in apply_cuts
    w1flux=w1flux, w2flux=w2flux, deltaChi2=deltaChi2, objtype=objtype)
  File "/home/travis/build/desihub/desitarget/py/desitarget/cuts.py", line 289, in isQSO_randomforest
    rf = joblib.load(pathToRF)
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/sklearn/externals/joblib/numpy_pickle.py", line 573, in load
    return load_compatibility(fobj)
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/sklearn/externals/joblib/numpy_pickle_compat.py", line 226, in load_compatibility
    obj = unpickler.load()
  File "/home/travis/miniconda/envs/test/lib/python2.7/pickle.py", line 864, in load
    dispatch[key](self)
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/sklearn/externals/joblib/numpy_pickle_compat.py", line 177, in load_build
    Unpickler.load_build(self)
  File "/home/travis/miniconda/envs/test/lib/python2.7/pickle.py", line 1223, in load_build
    setstate(state)
  File "sklearn/tree/_tree.pyx", line 633, in sklearn.tree._tree.Tree.__setstate__ (sklearn/tree/_tree.c:7860)
KeyError: 'max_depth'
```
i.e. they changed the `Tree` object underlying structure from 0.17.1 to 0.18.0 such that Trees pickled under one version can't be used in the next.   This is the core issue that we need to fix, and the fix can't be "just use scikit-learn 0.18.0 forever".

Cyeche:
> .e. they changed the Tree object underlying structure from 0.17.1 to 0.18.0 such that Trees pickled under one version can't be used in the next. This is the core issue that we need to fix, and the fix can't be "just use scikit-learn 0.18.0 forever".

Why not? Atlas and CMS decided that they used "root v5-34-16" till Higgs discovery ;-) (not sure about -16…) and the impact of such a decision was much important than  scikit-learn which is just used by TS so far….

tskisner:
Maintaining ancient versions of one or more packages for many years 
requires far more effort than just writing code that survives updating 
of packages.  In the NERSC environment the OS is upgraded several times 
a year, and even though we are taking snapshots of anaconda and not 
touching them, those might also have to be updated periodically due to 
low-level incompatibilities with an updated OS.  Some people will say 
to use shifter or some other virtualization technology, but those are 
mostly just a big hack of a solution at the moment on the Cray systems.

Either way, we can spend people resources writing more robust software 
or we can spend it writing infrastructure for using the same software 
for 10 years and then forcing everyone we work with to use that same 
infrastructure and ancient software...

Sorry I am late to this discussion :-)  Is there any way that we can 
write the data to a simple hdf5 dataset (with h5py)?  Or have the 
fundamental types of data (not just their serialized form) changed 
between versions?

rainwoodman:
I'd just retrain if scikit learn updates. In this sense we may want to
include the training scripts to reproduce the rf files in desitarget too.

Yu

Cyeche:
In the branch qso_py3 of desitarget, I pushed a script to train Random Forest and adaboost.
The training samples are available on nersc at:
/global/project/projectdirs/desi/target/qso_training/

Christophe. 
 

> I'd just retrain if scikit learn updates. In this sense we may want to
> include the training scripts to reproduce the rf files in desitarget too.
> 
> Yu

geordie666:
I'm nervous about including a dependency that potentially changes which targets are selected if the code is updated. It potentially means that target selection is not reproducible. 

As @tskisner points out, it's not feasible to freeze the version of scikit at 0.18.0. 

But, I don't like @rainwoodman's suggestion either, because DESI targeting will run for ~5 years, and if we rerun TS in Year 5 and find that we have different targets to those that were observed in Year 1 then TS is not reproducible.

I think that the only way forward is to hack out the part of scikit that trains the forest, and the part that applies the forest to test data and just maintain those as our own lines of Python code. Sorry, @Cyeche, but I'm going to be against merging anything with a direct scikit dependency.

Cyeche:
Ok, I have understood.  We will concentrate on the reverse-engineering part, this means computing the probability independently of scikit, starting with the 4 arrays per decision tree.  

I'm not surprised by this concern about stability when we use a programming language in which 1/2 is equal to 0 or 0.5 depending on the version ;-).

sbailey:
superseded by #92; closing without merging.

PR #85, opened 2016-10-17
Keep original targetid

rncahn:
This removes the line that changed TARGETID to the list (0, len(mtl)).  This was a mistake.  TARGETID should never be changed.  Also, this removes the line that checked that TARGETID was less than len(mtl), which it shouldn't be, now that we'ver kept the proper TARGETID.


sbailey:
I think these changes are OK but I'd like @forero to make sure they fit with the big picture.

Also note that this PR brings in changes for setting the Z and ZWARN values to -1 for unobserved targets, though they are also masked.  I'm not sure if this is necessary or a leftover debugging attempt.

rncahn:
Setting Z and ZWARN to -1 seemed consistent with what was done in the
absence of zcat.

forero:
This looks good to me. For mock files ```TARGETID``` is now created in ```desitarget```.

Issue #82, opened 2016-10-10
place to create unique TARGETID for mocks


forero:
Where in desitarget should we do the unique TARGETID computation? What logic should we follow to do that if we have hundreds of input mock files?

apcooper:
On the first point, I like @forero's solution in the email discussion, to do the assignment of TARGETIDS in the script that creates the merged Table of targets that gets passed to `make_mtl` (i.e. `mocks_to_fa_input`).

geordie666:
Ultimately, we're going to want to set the unique TARGETID once for all targets, so I think the correct place to do it is before the mtl code. Probably, though, using similar code to encode_mtl_targetid in:

https://github.com/desihub/desitarget/blob/master/py/desitarget/targets.py

But with a name like encode_targetid instead (as we want TARGETID to be set before the mtl stage)!

This, of course, does not preclude the code in targets.py being called from the code for the mocks.

Note that @sbailey and I discussed the form of TARGETID late in the summer, and came up with the following schema: 

https://desi.lbl.gov/svn/docs/technotes/targeting/targetID/trunk/

Apologies that this is currently only in SVN, I'll upload it to DocDB later today.

sbailey:
Thanks for distributing that note.  I also recall a conversation with @forero and/or @geordie666 that concluded with a target selection ingestion step that would
* resolve any TARGETID conflicts, creating a mapping of original ID to final TARGETID as needed
* perform QA (e.g. do the targets actually include non-zero fluxes?  Do they vary?)
* write the final "real" target catalogs with the fixed-forever TARGETID following the form that Adam documented in that note.

We flirted with the idea of letting MTL being the one that assigns the TARGETID, but ended up backing off on that idea and letting a pre-MTL pre-targetcatalog step handle it.  Real targets based on Legacy Survey data releases should be easy in that they have a DR+BRICKID+OBJID way to define the TARGETID already; the complications only arise for mocks and ancillary programs where TARGETIDs could collide.

forero:
Good. I believe that the consensus is having TARGETID somewhere pre-mtl. 

There is a new `mocks` branch. I will start coding the suggested TARGETID schema there in a place outside mtl.

PR #86, opened 2016-10-18
Targets and Truth from different mocks

forero:
This PR refactors code by @apcooper. We can now produce unique ```Targets``` and ```Truth``` files mixing mock files from the Dark and Bright program. ```TARGETID```s do not yet encode any information about target provenance. 

This is the [key script](https://github.com/desihub/desitarget/blob/e81191b06061bed59d2a4e8fbf5e71d8642ddffd/bin/mocks_to_targets_and_truth) doing all the work. It can be tested using [this yaml input file](https://github.com/desihub/desitarget/blob/e81191b06061bed59d2a4e8fbf5e71d8642ddffd/doc/mock_example/input.yaml). All paths refer to standard mock file locations on NERSC. 

A problem is that it takes ages (~50 min) to run on a login node. 90% of the time is spent writing the files to disk (the ```targets.fits``` file weighs 12GB).


moustakas:
Very nice developments.  A couple comments:  First, the magnitude limits in the yaml file (e.g., for BGS) probably shouldn't be hardcoded -- they should be pulled from ```desitarget.cuts``` (or a refactored version of that code).  And second, shouldn't the target catalogs be written out as either bricks or sweeps (sliced by position on the sky) rather than as one gigantic FITS table?

forero:
Hi @moustakas. I will explore the use of ```desitarget.cuts``` for the mocks that include magnitude information. 
I also like the idea of having the targets file written down as bricks/sweeps. The problem is that ```fiberassign``` only runs on single files and we are stuck with that monolithic structure at least in the near term.

apcooper:
Thanks Jaime, this is looking good. I haven't had much time to keep up with this in the last week... Things to figure out, maybe you've already got the covered:

- We need to make sure the TARGETIDS for skies and standards don't clash with those for targets, since they're treated separately from other sources.

-  We're now back to setting TARGETID separately for each source of targets, using a unified scheme (#82), and not re-writing it (#85) to avoid brute-force matches (i) from FA tilemaps to the input target list and (ii) from the input target list back to the mocks. That's great, but it would be nice to keep the no-matching optimization somehow. I'd suggest for (i) this is done by having the tilemaps from FA contain the input MTL rownumber (I'll raise this as an issue on FA), and for (ii) we do something along the lines @geordie666 [suggests]
(https://github.com/desihub/desitarget/issues/82#issuecomment-252732638).

The mocks won't all be split into bricks etc., but the same bit-range structure could be used, provided each mock was careful not to overstep the limits on number of files (almost certainly not a problem) and number of targets per file (4M per brick/file allowed in Adam's scheme). Larger numbers of targets per file could easily happen without some forethought by the providers of the mocks (typical mocks probably contain a much higher density of targets than those that would be eventually selected). Ensuring those limits were respected would boil down to a requirement on those responsible for making mocks.

I suppose we could force the mock TARGETIDs to use the actual legacysurvey brick for their sky positions, and somehow unify all the objectids across all the mocks. That would require a lot of extra processing and I can't see any good reason to do it. It seems most reasonable to allow each mock to interpret both of these bit ranges ('brick' and 'object') for itself when constructing TARGETID. Any thoughts?

Of course the tracing-back-without-matching-TARGETID information (i.e. which row in which mock file does for each row in the MTL correspond to) doesn't have to be stuffed into TARGETID itself. We could add 'MOCKFILE' and 'MOCKROW 'columns to truth.fits.

My earlier implementation kept an ordered list of source file paths in an extension of truth.fits (ordering is not crucial, could easily use key,value pairs instead). Is that still going to be done, to get back to the location on disk of an input mock file from a brick/file ID? The alternative would be to have a hardcoded algorithm for each mock separately (i.e. that knew about the file and directory structure of the particular mock) perhaps with some 'metadata' pulled from the yaml file. Or, to ensure mocks all follow some file/dir name conventions. I like the first option best and the last option least.

Of course, probably not all mock-makers care about going back to the rows in the input files. I think MWS does mainly because at this stage we want access to a lot more properties for the selected targets than it's necessary/practical to carry through the truth.

On John's points:

- I agree it should be an option, probably the default, to take the cuts from code in desitarget (isn't that the hardcoded way? When I made that script the idea was to have a separate yaml files for each run with different permutations of sources and cuts, rather than to hardcode them. (Incidentally, right now there are defaults arguments all the way through the stack down to the actual selection routines, which seems a bit risky...)

- I think the monolithic targets.fits file is a requirement of fiberassign, right @forero?

geordie666:
On this point from @apcooper:

I suppose we could force the mock TARGETIDs to use the actual legacysurvey brick for their sky positions, and somehow unify all the objectids across all the mocks. That would require a lot of extra processing and I can't see any good reason to do it. It seems most reasonable to allow each mock to interpret both of these bit ranges ('brick' and 'object') for itself when constructing TARGETID. Any thoughts?

One reason to do this would be that I'm intending to add real per-brick (or sub-brick-level) fluctuations to the target density. So the target will already know which brick it's in. Another might be to better mimic the flow of real data (in that TARGETID would be set "first"). But, I don't have a very strong feeling on these points.

apcooper:
@geordie666: The thing I was concerned about was how to unify the objectid across **all** the mocks from different sources. 

What you suggest sounds like an intermediate stage combing before target selection, in which we make a 'super-mock' catalogue of all the targets in a given brick (basically mocking the full tractor catalog for each brick). Interesting...

One downside I can see is that it would still require re-writing TARGETID from the raw mocks (since they will have assigned objectids without knowing how many targets were on the same bricks in the other mocks). It might take quite some storage and time. And unless we propagate all the information from the mocks, we (MWS at least) would still need to go back to the raw files, so this would be another layer to go through.

I think even in the real survey it's not the case that 100% of the targets come from the legacysurvey bricks, right? Section 2.1 of the technote seems to suggest using the 'headroom' in brickid (after the 662174 'proper' brickids) to flag targets drawn from those other surveys. That's 386402 'extra' bricks, not enough room for a whole other set from two different 'bricked' target catalogues, so that might work fine in reality, if the extra catalogues are sparse, but it might not be so useful for multiple mocks that are split into bricks. 

Right now we have at least three major separate mocks sources: dark, BGS and MWS. MWS is split into two separate mocks. If we manage to stick to <64 mocks (seems reasonable) maybe we could interpret the DR bits as specifying the source if mock_bit=1?

Sorry to add further noise here, but another thing I'm not sure about is if/how/when the stars from the MWS are going get folded into the galaxy selection as contamination and whether we need to worry about that in this context. Perhaps that's a case in favour of making a unified mock target catalogue...?

forero:
Thanks for all your comments. 

I will merge this PR and create new github-issues to have

1. A unique TARGETID across Targets and StandardStars.
2. Encode mock provenance into TARGETID.
3. Explore the possibility to split the monolithic Targets files into sweeps/bricks (that is going to require work on ```fiberassign```).

I believe those were the main points raised in this discussion. Although, I won't have time to address them in the next couple of weeks (I will delay that to November). My focus will be checking that the loop fiberassign-mtl-quickcat works fine to produce the first batch of results for the datachallenge.

forero:
Ah! Probably one more item would be

4. Including Target contamination.

moustakas:
```
Sorry to add further noise here, but another thing I'm not sure about is if/how/when the stars from the MWS are going get folded into the galaxy selection as contamination and whether we need to worry about that in this context. Perhaps that's a case in favour of making a unified mock target catalogue...?
```
Keep in mind that stars won't end up as contaminants to the other target classes unless we add/include depth information (i.e., photometric uncertainties).  when we ask whether certain objects (e.g., stars) pass particular target-selection criteria (e.g., the ELG criteria), since the cuts are explicitly designed to minimize the selection of contaminants.  

I'm not sure there's even an open issue to address this shortcoming.

PR #92, opened 2016-10-20
Clean version of the quasar Random Forest selection

geordie666:
This PR supersedes https://github.com/desihub/desitarget/pull/80. It is a full version of the quasar Random Forest selection but Python 2/3 changes have been merged self-consistently. It addresses all of the requests in https://github.com/desihub/desitarget/pull/80 but ensures

* Python 2 and 3 compatibility
* Unit tests are passed and new code has coverage
* Non-standard file formats are somewhat circumvented (there are no tar or pickle files, the RF is stored as a numpy instance, which is not ideal but is better than previous options)

Note that the training code has deliberately been left out of this PR, but could be included if @sbailey thinks it prudent. The training code would require some documentation to be useful to a general user.

Once this is merged https://github.com/desihub/desitarget/pull/80 can be closed (without merging).

h/t Christophe Yeche and Etienne Burtin

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ env:
         # - NUMPY_VERSION=1.10
         # - SCIPY_VERSION=0.17
         - ASTROPY_VERSION=1.1.1
-        - DESIUTIL_VERSION=1.4.0
+        - DESIUTIL_VERSION=1.9.0
         - MAIN_CMD='python setup.py'
         # These packages will always be installed.
         - CONDA_DEPENDENCIES='pyyaml qt=4'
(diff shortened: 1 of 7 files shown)


sbailey:
We definitely want more documentation and code for how the random forest is trained and exactly what the pre-cuts are and what variables are used by the RF, but that doesn't have to be a blocking factor for this PR.  Similarly, a future update could move the npz format file to a cross-language format like hdf5 or fits, but at least this has the property of working with python 2.7 and 3.5 and not depending upon scikit-learn.  Merging now.

PR #96, opened 2016-10-25
Add training of Random Forest and BDT

Cyeche:
Add training of Multivariate Analysis (MVA) with stripe82 stars/QSO sample of DR3
Tested against python 2 and python 3
The training samples are on a nersc disk

### DIFF
--- a/bin/qso_training
+++ b/bin/qso_training
@@ -1,0 +1,19 @@
+#!/usr/bin/env python
+
+
+import os, sys
+import numpy as np
+
+from desitarget.train.train_mva_decals import train_mva_decals
+
+from argparse import ArgumentParser
+ap = ArgumentParser()
+ap.add_argument('--step',choices={'test','train','extract_myRF'},help="Training or test of qso TS algorithm")
+ap.add_argument('--debug', action='store_true')
+
+ns = ap.parse_args()
+
+train_mva_decals(ns.step,ns.debug)
+
+
+
--- a/bin/qso_training
+++ b/bin/qso_training
@@ -7,8 +7,8 @@ import numpy as np
 from desitarget.train.train_mva_decals import train_mva_decals
 
 from argparse import ArgumentParser
-ap = ArgumentParser()
-ap.add_argument('--step',choices={'test','train','extract_myRF'},help="Training or test of qso TS algorithm")
+ap = ArgumentParser(usage='Compute four MultiVariate Analyses based on scikit-learn.\n Training and test samples are available on nersc at :\n /global/project/projectdirs/desi/target/qso_training/')
+ap.add_argument('--step',choices={'test','train','extract_myRF'},help="Test or train qso TS algorithm. Store the Random Forest with a DESI persistency in a compressed numpy array")
 ap.add_argument('--debug', action='store_true')
 
 ns = ap.parse_args()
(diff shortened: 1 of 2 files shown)


geordie666:
I'm checking that the mechanics of all of this works (it seems to) but I think that we could really use some extra documentation and examples regarding how to train the MVA. 

I think the basic philosophy (rooted in reproducibility) should be that if we agree to a target selection method as the new default approach, then that method should have documentation with examples. Ideal would be some combination of a technote and some examples for use in the code header so that the training can be reproduced. At the minimum I think we need the exact commands that should be run to reproduce the same RF file that ended up in the desitarget repo.

I'm happy to merge this PR (probably tomorrow) if I get an explicit promise from @Cyeche that more documentation will arrive (and, in particular, some examples of how to use the training code). But, I'll be bugging @Cyeche regularly for that documentation if it doesn't appear soon...

sbailey:
Let's not merge until we've confirmed that

sbailey:
Sorry; meant to cancel comment instead of closing PR.  Reopening.

I would prefer to wait until Adam has confirmed that he can reproduce the RF file for the QSO target selection; this could be just a couple of lines of documentation to cut-and-paste and try.  But ok if Adam wants to merge anyway and bug Christophe until he gets those commands.  If the commands don't work to reproduce the file, then we will have to find some appropriate penance for Christophe...   :)

sbailey:
Actually, backing off on this — we arguably should have required this code, documentation, and independent reproducibility verification before merging in random forests as the baseline QSO selection method in the first place.  But given that we did merge that, it seems silly to hold up merging part of the documentation (this training code) just because we want to get a bit more documentation too.

If @geordie666 is satisfied, go ahead and merge (and keep bugging @Cyeche for the full documentation :)

Cyeche:
I have added more comments  both in desitarget/bin/qso_trainning and desitarget/py/desitarget/train

geordie666:
@Cyeche ...can we get unit test coverage for the training module? So, add something in:

/py/desitarget/test/test_cuts.py

That runs the training on a small subset of objects (which would be placed in the /t directory...

sbailey:
Travis tests were failing due to outage of continuum.io for downloading conda installation; I restarted those.

Are we being too formal with requiring unit tests for the training code before merging?  I would really like to have confirmation that someone other than Christophe can run the training and get the same result, but rerunning a test training as part of every unit test run may be overkill and a barrier to contribution.

geordie666:
I can confirm that all aspects of @Cyeche's code in bin/qso_training work just fine. But Travis tests are still failing because of git check/apt-get type errors. This is ready to merge, but I'd like to check that it passes Travis first. Is there a magic way to trigger Travis tests without an additional commit or opening-and-closing a pull request? I just ask for personal edification.

sbailey:
Thanks for confirming that the training code works for you too.

If you click on the "details" link by the tests listed on the PR page, that will take you to the Travis page.  On the upper right is a link for "restart build" that will rerun all the tests without requiring meaningless git updates to restart the build.  I just restarted it, then it immediately failed with a completely blank log.  Travis is having a bad day; let's try again in a little bit.

geordie666:
I don't see the "restart build" button. Stack Overflow suggests that this
is just reserved for the owner of the project? Anyway, I can always commit
pointless doc changes.

PR #98, opened 2016-10-31
Add "--check" option to select_targets...

geordie666:
... to process Tractor/sweeps files and look for corrupt files without having to run target selection

### DIFF
--- a/bin/select_targets
+++ b/bin/select_targets
@@ -6,7 +6,7 @@ import os, sys
 import numpy as np
 
 from desitarget import io
-from desitarget.cuts import select_targets, qso_selection_options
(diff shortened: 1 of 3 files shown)


bin/select_targets
```diff
@@ -19,6 +19,7 @@ ap = ArgumentParser()
 ap.add_argument("src", help="Tractor file or root directory with tractor files")
 ap.add_argument("dest", help="Output target selection file")
 ap.add_argument('-v', "--verbose", action='store_true')
+ap.add_argument('-c', "--check", action='store_true')
```
sbailey: Please add a help= string to this argument so that someone running `select_targets --help` will be told what the `--check` option will do.  Otherwise fine to merge.

PR #99, opened 2016-10-31
mock truth file TRUETYPE to match SPECTYPE, not sourcetype

sbailey:
This PR updates the truth table output to have TRUETYPE match what a redshift fitter should report in SPECTYPE, i.e. 'GALAXY', 'QSO', or 'STAR' instead of 'ELG, 'LRG', 'MWS_WD', etc.  The original source type is stored as a column 'SOURCETYPE', which we may or may not need.

Other changes that came along for the ride:
* Added example for how to subset by RA,dec boundaries to doc/mock_example/input.yaml
* Updated subsetting to be at the beginning while reading in mocks rather than just before writing the final results out.  This is for speed, to avoid calculating a bunch of things for targets that will be discarded anyway.
* Parallelized reading the 300 input galaxia mock files

I tested this with a quick_survey loop over 8 epochs and verified that all target types are still observed.  It turns out that this did not require any changes to quickcat since that already supported TRUETYPE='GALAXY' (plus DESI_TARGET bits) vs. TRUETYPE='ELG', 'LRG', 'QSO', etc.

### DIFF
--- a/doc/mock_example/input.yaml
+++ b/doc/mock_example/input.yaml
@@ -1,6 +1,12 @@
 #mock target configuration file
 output_dir: /tmp
 run_label:  GX_100PC_MXXL_JT2P4Y_JULY16_B
+subset:
+    ra_dec_cut: False
+    min_ra: 200
+    max_ra: 220
+    min_dec: 0
+    max_dec: 20
 sources:
     SKY: {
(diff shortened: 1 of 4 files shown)


forero:
I also did quick_survey loop tests. Everything runs and the results look reasonable.

Issue #108, opened 2016-11-01
reading MWS mocks v.0.02. (with multiprocessing) crashes


forero:
The following code
```python
import desitarget.io
import desitarget.mock.io
mock_dir='/project/projectdirs/desi/mocks/mws/galaxia/alpha/0.0.2/bricks'
iter_mock_files = desitarget.io.iter_files(mock_dir, '', ext="fits")
import multiprocessing
file_list = list(iter_mock_files)
nfiles = len(file_list)
ncpu = max(1, multiprocessing.cpu_count() // 2)
print('using {} parallel readers'.format(ncpu))
p = multiprocessing.Pool(ncpu)
target_list = p.map(desitarget.mock.io._load_mock_mws_file, file_list)
```
gives a long error message starting with
```
Traceback (most recent call last):
  File "multiproc_bug.py", line 13, in <module>
    target_list = p.map(desitarget.mock.io._load_mock_mws_file, file_list)        
  File "/global/common/edison/contrib/desi/conda/conda_3.5-20160829/lib/python3.5/multiprocessing/pool.py", line 260, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "/global/common/edison/contrib/desi/conda/conda_3.5-20160829/lib/python3.5/multiprocessing/pool.py", line 608, in get
    raise self._value
multiprocessing.pool.MaybeEncodingError: Error sending result: 
````
ending with
```
 Reason: 'error("'i' format requires -2147483648 <= number <= 2147483647",)'
```

This seems to be caused by too large arrays (present in v.0.0.2 of the mocks, not in version v.0.0.1).

Can you reproduce the error?
Any idea for a workaround?

sbailey:
Yes, I can reproduce it, and it appears to be [python issue #17560](https://bugs.python.org/issue17560): `multiprocessing.Pool.map` can't return objects with more than 2**32-1 bytes since it uses a signed 32-bit int for encoding how many bytes it is returning (!).

Ugh.

The short term workaround is in PR #110, which in addition to reading Lyman-alpha mocks also returns to using serial I/O for the galaxia mocks.  I'll think about whether there is another easy way to do the parallel I/O .

sbailey:
If future galaxia mocks are smaller we could return to parallel I/O, but in the meantime I'm closing this ticket.

Issue #107, opened 2016-11-01
include Lyman-alpha mocks in select_mock_targets output


sbailey:
Include Lyman-alpha mocks in select_mock_targets output.

Even just having the (ra,dec,z) would be useful for combining with other target classes for fiberassigment+quickcat based simulation loops.

Second step is including the spectra to simulate for later studies.

forero:
Solved in #110.

PR #110, opened 2016-11-02
reads lya-mocks

forero:
Solves #107. It also temporarily disabled multiprocessing to read galaxia mocks (see bug #108). It also includes changes in the yaml input file to manage different inputs for QSOtracer vs. QSOlya.


sbailey:
I updated the comment about the parallel I/O to link to the python multiprocessing bug/limitation.

New code looks good, but please update `doc/mock_example/input.yaml` so that it matches the new form needed for QSOlya vs. QSOtracer.  i.e. let's keep that file up-to-date as a working example to use.  Otherwise ok to merge.

forero:
Good. I've updated the file `doc/mock_example/input.yaml` to reflect the latest input mocks.

PR #111, opened 2016-11-07
removed hack that checked for zero values of dust as sweeps files now…

geordie666:
… fixed

### DIFF
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -382,18 +382,16 @@ def unextinct_fluxes(objects):
     else:
         result = np.zeros(len(objects), dtype=dtype)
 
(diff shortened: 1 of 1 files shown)


geordie666:
If we merge this, we might consider this code the code for the "official DR3 Target Release" and I'll cut targets later today

sbailey:
Does this require a DR 3.1 release?  That doesn't seem to be at NERSC yet.  i.e. will this change cause select_targets to crash on the current latest DR 3.0 release at NERSC or did Dustin et al fix files in-place?

If this would break the ability to run on DR3 and if DR3.1 isn't happening today, I'd suggest proceeding with the desitarget tag and reference run on the code as-is, and then merge this PR, retag, and rerun whenever 3.1 is publicly released.

geordie666:
Sounds good. It's not a big change to the code, so I'll just close the PR
for now.

PR #115, opened 2016-11-10
Target columns

belaa:
Adds shape vars to specified columns to read when reading tractor files.

### DIFF
--- a/py/desitarget/io.py
+++ b/py/desitarget/io.py
@@ -24,7 +24,8 @@ tscolumns = [
     'DECAM_FRACFLUX', 'DECAM_FLUX_IVAR',
     'WISE_FLUX', 'WISE_MW_TRANSMISSION',
     'WISE_FLUX_IVAR',
-    'SHAPEDEV_R', 'SHAPEEXP_R', 'SHAPEDEV_E1', 'DCHISQ',
+    'SHAPEDEV_R', 'SHAPEEXP_R', 'SHAPEDEV_E1', 'SHAPEDEV_E2',
+    'SHAPEEXP_E1', 'SHAPEEXP_E2', 'DCHISQ',
     ]
 
 def read_tractor(filename, header=False, columns=None):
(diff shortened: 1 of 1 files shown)


dkirkby:
It looks like two tests in `test_io.py` are failing because they read a sweep file in `desitarget/py/desitarget/test/t/` that is missing the new columns.

@sbailey Can you point @belaa to the code you used to create those sweep files? (or just update them if that's easier).

geordie666:
@dkirkby @belaa: I can see the code @sbailey used to make the test files. I'll update them on this branch myself.

sbailey:
For the record, that code was rather inelegantly dumped in `py/desitarget/test/make_testdata.py` with the header comment
```
#- For the record (and future updates):
#- This code was used to generate tractor and sweep file subsets for testing.
#- The hardcoded paths are for Stephen's laptop, but you can swap out any
#- legacy survey data release path as needed.
```

geordie666:
Updating these files is actually a little trickier than I thought. The problem is that many of the unit tests are predicated on a specific file structure from DR1/DR2 but the sweeps/tractor file structures are different for DR3. So, for instance, adding in the `SHAPE` parameters from a DR3 file messes with unit tests that expect, e.g., the `BRICK_PRIMARY` column from the DR2 sweeps. It seems like it will be a lot of work to revamp the 4 or 5 unit tests that need updates, so I'll have to put this on the back-burner for a while.

sbailey:
As a temporary work around until we fix the unit tests, you can still use this code to read a custom set of columns from sweep/tractor files with:
```python
my_columns = desitarget.io.tscolumns.copy()
my_columns.extend(['SHAPEDEV_E1', 'SHAPEDEV_E2'])
data = read_tractor(filename, header=False, columns=my_columns):
```
This is just a mild wrapper around fitsio that also takes care of fixing up some things like 0-length files and whether BRICK_PRIMARY is in the file or not.

The intension of `desitarget.io.tscolumns` is to define exactly which columns are needed for target selection and downstream spectroscopic pipeline use so that we can efficiently process the tractor/sweep files without having to read all 44/69 columns.  I think we will want these additional E1/E2 columns downstream, so ok to add now, but we shouldn't add columns just because we want to use them temporarily for R&D studies or simulations.

geordie666:
I'm closing this. We'll revisit when I've had time to update the unit tests to cope with DR3 files

Issue #90, opened 2016-10-20
Include target contamination


forero:
`select_mock_targets` needs to output contaminants in addition to real targets.
* See #101 for defining input format for contaminant densities
* Make fake file with random variations to incorporate with select_mock_targets
* Update select_mock_targets to output contaminants for each target class with these densities.

Let's start with all contaminants being stars to start with, and update with more realistic mixes of stars vs. galaxies in a later iteration.

sbailey:
Note: desimodel svn trunk now has a desi-tiles.fits file with a STAR_DENSITY column that could be used as a proxy to vary the contaminant rate.  There will be other types of contaminants, but this is a starter.

PR #124, opened 2016-11-26
Mock contaminants

forero:
This PR allows the possibility to have contaminants in the mocks (solution to #90). 

This is the simplest possible implementation. The idea is including a new kind of variable in the input file called `truth_name` that goes together with `target_name`. 

For example `BADQSO` are added as
```
    BADQSO: {
       target_name: QSO,
       truth_name: STD_FSTAR,
       root_mock_dir: /project/projectdirs/desi/mocks/mws/galaxia/alpha/v0.0.2/bricks,
       format: galaxia,
       global_density: True,
       selection: ndens
       }
```
meaning that the inputs are the stars from galaxia, the inputs will be downsampled to match a global number density value (stored in `desimodel/targets.dat`). In the `targets.fits` file the selected objects will be seen as a `QSO` and as a `STD_FSTAR` in the `truth.fits` file.


forero:
Last commit also addresses #125

py/desitarget/mock/selection.py
```diff
@@ -20,8 +20,34 @@
...
+    Parameters:
+    -----------
+    bricks (array): array of bricknames.
+
+    Output:
+    -------
+    Lookup Dictionary. lookup['brickname'] returns a list with all the indices
+    in the input array bricks where bricks=='brickname'
+    """
+    lookup = {}
```
sbailey: FYI, I updated this function to use "lookup" instead of "l" (lowercase ell) as the variable name, since l (ell) can look so much like 1 (one) and I (capital eye).

doc/mock_example/input.yaml
```diff
@@ -69,15 +84,18 @@ sources:
     }
     MWS_MAIN: {
        target_name: MWS_MAIN,
+       truth_name: MWS_MAIN,
        mag_bright: 15.0, 
        mag_faint_filler: 19.0, 
        mag_faintest: 20.0, 
-      root_mock_dir: /project/projectdirs/desi/mocks/mws/galaxia/alpha/v0.0.2/bricks,
+       # root_mock_dir: /project/projectdirs/desi/users/forero/galaxia_small,
```
sbailey: I commented out this testing location (forero/galaxia_small), but a small version of the galaxia catalog would be very useful.  Please move a copy with a version number to the standard /project/projectdirs/desi/mocks/ area, and update this commented entry to point to that.

forero: done.

PR #127, opened 2016-11-29
Mock flux shape

sbailey:
This PR partially addresses the needs of desihub/desisim#191, which needs DECAM_FLUX and OIIFLUX info in the mocks as part of the quickcat redshift efficiency model.  It adds a `select_mock_targets ... --realtargets {targetcatalog}` option to point it to a real data target catalog to use for getting the distributions, e.g. /project/projectdirs/desi/target/catalogs/targets-dr3.1-0.8.1.fits .  

* If provided, it uses those for DECAM_FLUX for ELG, LRG, and QSO targets.
* TODO: It does not yet fill in anything for DECAM_FLUX for BGS or MWS targets, which do provide some magnitudes in their mocks
* It also propagates `SHAPEDEV_R` and `SHAPEEXP_R` since we have those in the real target catalogs even though we don't use them yet.

HACK: the quickcat model also needs [OII] flux for the ELG model, which isn't present in the current mocks or the real data target catalog.  I add this to the truth file for the ELGs following a made up model of the maximum [OII] flux vs. r-band flux.  Below shows the distribution of [OII] flux vs. r flux for templates from desisim.templates.ELG.  The red curve is my toy model of maximum [OII] flux vs. r flux, and I randomly distribute [OII] flux underneath the red curve given the r flux.

oii_vs_r

Or in code:
```python
rflux = targets['DECAM_FLUX'][isELG][:,2]
maxflux = np.clip(3e-16*rflux, 0, 7e-16)
truth['OIIFLUX'][isELG] = maxflux * np.random.uniform(0,1.0,size=nELG)
```
This is wrong but it might not be crazy wrong for the purposes of quick simulations for the OSU meeting, and I'm not sure what else could be done given the short timescale.  Complaints are welcome if they come with code for an alternate way to assign [OII] flux that correlates with photometry that are randomly assigned from real data to a mock catalog that didn't have fluxes in the first place...

### DIFF
--- a/bin/select_mock_targets
+++ b/bin/select_mock_targets
@@ -9,6 +9,7 @@ import yaml
 parser = argparse.ArgumentParser()
 parser.add_argument('--config','-c',default='input.yaml')
 parser.add_argument("--output_dir", "-O", help="Path to write the outputs", type=str, default="./")
+parser.add_argument("--realtargets", "-r", help="Path to real target catalog", type=str)
 args = parser.parse_args()
 
 if not os.path.exists(args.config):
@@ -21,8 +22,15 @@ print('')
 with open(args.config,'r') as pfile:
     params = yaml.load(pfile)
 
+if args.realtargets is not None:
+    from astropy.io import fits
+    print('Loading real targets from {}'.format(args.realtargets))
+    realtargets = fits.getdata(args.realtargets)
+else:
+    realtargets = None
+
 # Construct Targets and Truth files
-desitarget.mock.build.targets_truth(params, args.output_dir)
+desitarget.mock.build.targets_truth(params, args.output_dir, realtargets=realtargets)
 
 print('done!')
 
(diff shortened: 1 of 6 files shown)


julienguy:
We need to do a test run to check efficiencies are reasonable with the oii flux dependent quickcat implementation.

sbailey:
I tested this with quicksurvey_example/master and desisim/master with Christophe's latest quickcat efficiency model, getting 

Target | Obs Efficiency | Redshift Efficiency
--- | --- | ---
LRG      | 0.950 | 0.967
ELG      | 0.779 | 0.977
QSO      | 0.859 | 0.941
BGS_ANY  | 0.166 | 0.978
MWS_ANY  | 0.140 | 0.981

Something is going on with the bright time fiber assignment efficiency (e.g. perhaps the example doesn't have enough bright tiles), but that is separate from this propagation of fluxes, so I'll merge this now and track that as a separate issue.

PR #131, opened 2016-12-05
added galdepth_r, depth_r to mock targets

forero:
Adds two new columns into the mock `targets.fits` file: `GALDEPTH_R` y `DEPTH_R`. It's constant for all galaxies in a brick.


sbailey:
looks good; let's merge as soon as tests finish running (and presumably pass...)

PR #138, opened 2017-01-24
HTM

geordie666:
@sbailey mentioned that we might want HTM pixel information added to the target files for future grouping on the sky (e.g. bundling of fibers). This is a Python-only implementation for constructing HTM IDs.


py/desitarget/htm.py
```diff
@@ -0,0 +1,393 @@
...
+
+All-Python module for performin HTM look-ups
+See here for HTM: http://www.skyserver.org/htm/
+"""
+from __future__ import (absolute_import, division)
+#
+from time import time
+import numpy as np
+from numpy.core.umath_tests import inner1d
+import fitsio
```
sbailey: leftover fitsio import?

py/desitarget/htm.py
```diff
@@ -0,0 +1,393 @@
...
+    t0 = time()
+
+    #ADM convert input spherical coordinates to Cartesian
+    v = SkyCoord(ra=ra*u.degree, dec=dec*u.degree, frame='icrs')
+    v.representation = 'cartesian'
+
+    #ADM convert SkyCoord object to numpy array for speedier manipulation
+    #ADM .T is the transpose attribute
+    #ADM this conversion also allows either single floats or longer arrays to be passed
+    v = np.vstack(np.array([v.x.value,v.y.value,v.z.value])).T
```
sbailey: Is doing this transform via SkyCoord necessary?  Usually unitful astropy stuff is quite slow.  This reminded me of some code from @rainwoodman in a languishing PR in desimodel:
```python
def _embed_sphere(ra, dec):
    """ embed RA DEC to a uniform sphere in three-d """
    phi = np.radians(np.asarray(ra))
    theta = np.radians(90.0 - np.asarray(dec))
    r = np.sin(theta)
    x = r * np.cos(phi)
    y = r * np.sin(phi)
    z = np.cos(theta)
    return np.array((x, y, z)).T
```
With this, we could drop SkyCoord and astropy.units imports

py/desitarget/test/test_htm.py
```diff
@@ -0,0 +1,32 @@
+import unittest
+from pkg_resources import resource_filename
```
sbailey: leftover pkg_resources import?

py/desitarget/htm.py
```diff
@@ -0,0 +1,393 @@
...
+All-Python module for performin HTM look-ups
+See here for HTM: http://www.skyserver.org/htm/
+"""
+from __future__ import (absolute_import, division)
+#
+from time import time
+import numpy as np
+from numpy.core.umath_tests import inner1d
+import fitsio
+from glob import glob
```
sbailey: leftover glob import?

py/desitarget/test/test_htm.py
```diff
@@ -0,0 +1,32 @@
...
+
+from desitarget import htm
+
+class TestHTM(unittest.TestCase):
+
+    def setUp(self):
+        self.ra = np.array([ 287.87029928,  223.95814592,  328.99793884,  127.24251626])
+        self.dec = np.array([ 16.22908364, -53.49629596,  24.40112077, -23.14789754])
+        self.id = np.array(['N023330222213310122202','S233100021230000131322','N031100100010011323332','S131112210233032311131'],dtype='<U22')
+        self.intid = np.array([14015193368226, 12043250829178, 14104942108414, 10813196528989])
```
sbailey: Would be reassuring to add tests for special cases to make sure this code handles edge cases the same as the official HTM definition.  e.g.
  * points on vertices:
    * north and south poles
    * ra=0,90,180,270 dec=0
    * at "inner" vertices at a higher depth
  * points on edges, e.g. ra=0, dec=10

geordie666: I've addressed @sbailey's requested changes. I can't guarantee exact agreement with the official HTM release, because of the infinitely recursive nature of the tree. But, the new "corner-case" unit test demonstrates agreement to a few-millionths of an arcsecond at level 20 (i.e. practically conceivable cases for targets/fiber bundles). If perfect agreement with one of the official HTM releases is desired, then we should just adopt that release as gospel.

sbailey:
I see your corner tests are actually testing 1e-9 off of the corner.  The point of the request for corner and edge tests is that presumably there is some definition for which triangle it should go into, and we should match that, i.e. get our < vs. <= to match the official HTM release.

sbailey:
Passing forward some information from Martin White:  HTM is describe in more detail than the website in a paper here:

https://www.microsoft.com/en-us/research/publication/indexing-the-sphere-with-the-hierarchical-triangular-mesh/

And the associated C# code is available at

http://www.skyserver.org/htm/HtmPrimer/tut_install.html

I've also added some `@unittest.expectedFailure` flagged cases of boundary and trixel corner cases that disagree with the SDSS skyserver implementation at a higher level than just float rounding 20 layers deep.

I'm not pushing for this to be resolved at high priority, but I'm dumping this info here before I forget about it.

sbailey:
Closing this PR and branch since we are moving towards healpix instead of HTM.  If we need to revive this  code it is sitting at 0.8.2-39-gd36cfc3 (i.e. d36cfc3a24d6489f0653ca9d76d834bba93a7ac2) in the git history.

geordie666:
Thanks for remembering to close this!

PR #143, opened 2017-02-01
2016v3 LRG algorithm incorporated into new "sandbox" cuts module

moustakas:
This PR incorporates (hopefully correctly!) one version of the LRG target selection criteria developed by Dawson & Eisenstein in the summer of 2016.  See [here](https://desi.lbl.gov/trac/wiki/TargetSelectionWG/LRGretuning) for extensive documentation.  

Thank you to Daniel Eisenstein for providing me the code to apply the actual cuts in Python.

The algorithm has been placed into a new "sandbox" area (as advocated by @geordie666) for algorithms under consideration.  The cuts are applied via
```
select_targets /path/to/dr3 outfile.fits --sandbox
```
Comments are welcome.  One might object to the fact that the code doesn't apply the "sandbox" cuts *in addition to* the nominal cuts (to make comparisons easy), but in fact this would be a big pain because of the existence of only a ~single targeting bit for each target class.  And in fact, we need to think about how to differentiate *different* algorithms applied to the *same* spectral class (e.g., two different LRG selections), which right now is not supported by the code.  But let's save that for a future PR as the need arises.

Finally, I haven't had time to write any tests -- sorry -- but we need this to not hold up @jaekor91's work on incorporating the ELG XD selection algorithm.


geordie666:
I'd like to see the name of the function that is being run in sandbox written to the header of the output file. Perhaps as one long string. So, e.g., if I'm in the function 

isLRG_2016v3

Then the name of this function can be returned using:

from inspect import stack
selection_name = stack()[0][-3])

Perhaps, in the header, DEPNAM04 could be "sandbox" and DEPVER04 could be a long string listing the sandbox functions that were run.

I won't hold the PR up on that basis, though.

sbailey:
I contributed a test that would have caught that `_is_row` import bug, and also fixed the `_is_row` import.  I didn't find other problems.

Following up on @geordie666's suggestion of stashing the function name: at minimum the output header keywords should indicate that sandbox cuts were applied instead of the regular ones.  Even better to have the explicit function names like @geordie666 suggests.

geordie666:
@sbailey...good catch on the __init__.py. I can't find any other issues, so go ahead and merge if you want.

Should we open an issue about stashing the functions names in the header?

sbailey:
> @sbailey...good catch on the `__init__.py`. I can't find any other issues, so go ahead and merge if you want.

Having said that I didn't find any other problems after adding the tests, Travis auto-testing on both 2.7 and 3.5 caught that one for me.  :)

> Should we open an issue about stashing the functions names in the header?

I went ahead and added a "sandboxcuts" DEPNAM* to the header with values of T/F to provide the minimal provenance for whether the sandbox cuts were applied.  That along with the code/git version which is also stored in the header may be sufficient.

If you agree, go ahead and merge.  Feel free to open a separate issue for getting the full function names in there if you think that is important.

geordie666:
I think the function names might be useful for any TS people who run code to produce files and then lose track of which files they ran using which code. It'd be an enhancement, though. Merging...

moustakas:
Thank you for the rapid review and apologies I wasn't around to fix the issues identified.  I'm deleting the branch.

Issue #136, opened 2017-01-22
refactor and expand targets_truth


moustakas:
```mock.build.targets_truth``` needs to be refactored and expanded to include the metadata necessary to generate spectra.  This issue is directly related to https://github.com/desihub/desisim/issues/203.

I am also planning to address #133 and #130 here.

moustakas:
I am refactoring ```mock.build.targets_truth``` to add identifying spectral information to each mock target.  This will slow things down but I think it's the behavior we want because doing the mapping will a more realistic set of metadata (```DECAM_FLUX```, ```OIIFLUX```, etc.) to each mock object that I think we still want, even if we're not generating spectra (i.e., even if we're just doing a "quicksurvey"-like, catalog to catalog data challenge).  In particular, this should address issues like #127 which right now are being handled in a very ad hoc way.

I'm asking in case this *isn't* the behavior we want.  @sbailey?  @forero?  @dkirkby?

moustakas:
One more question: can I break backwards compatibility (=somewhat annoying)?  We can always use an (older) tagged version of the code if necessary, right?

moustakas:
I decided that since the spectra are being generated anyway in ```mock.build.targets_truth``` (in order to be able to apply the target selection cuts correctly), we might as well write them out.  Any thoughts on what data model I should use to write these out?  I'm parallelizing by ```brickname```, so that's one option, although to use the ```simspec``` input to ```desisim.quickgen``` we actually want the spectra organized into *frames* (which would require a night and exposure ID).

Should I just write the spectra in bricks (across all the spectral classes) and just let the downstream steps (which, e.g., generate the fibermap files) deal with the reorganization?  @sbailey @forero?

forero:
To me the spectra in bricks feels like the most convenient solution.

moustakas:
@forero What are the *required* columns in ```targets.fits``` and ```truth.fits``` in order for ```quicksurvey``` to be run on the output of ```select_mock_targets```?

forero:
In `targets`: TARGETID, RA, DEC, DESI_TARGET, BGS_TARGET, MWS_TARGET, SUBPRIORITY and OBSCONDITIONS. In 'truth': TARGETID, TRUEZ, TRUETYPE and SOURCETYPE.

sbailey:
Replying to a question a few messages back:

[...]

PR #146, opened 2017-02-02
select STD_FSTAR and STD_BRIGHT F-type standards via dedicated function

moustakas:
This refactor creates a new cuts.isFSTD() function which selects both/either STD_FSTAR and STD_BRIGHT standard stars.  The magnitude cuts for STD_FSTAR were set to 16<r<19 (observed r-band) while the STD_BRIGHT cuts were semi-arbitrarily set to 16<r<18, and the documentation on the [target selection page](https://desi.lbl.gov/trac/wiki/TargetSelectionWG/TargetSelection#SpectrophotometricStandardStarsFSTD) was updated.  

I also did a bit of cleanup of the code for style and organization.

This refactor was needed as part of #136.  

Tests are still running but creating the PR now anyway.

### DIFF
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -113,7 +113,7 @@ def isFSTD_colors(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None, 
     return fstd
 
 def isFSTD(gflux=None, rflux=None, zflux=None, primary=None, decam_fracflux=None,
-           objtype=None, decam_snr=None, bright=False):
+           objtype=None, decam_snr=None, obs_rflux=None, bright=False):
     """Select FSTD targets using color cuts and photometric quality cuts (PSF-like
     and fracflux).  See isFSTD_colors() for additional info.
 
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -93,22 +93,71 @@ def isFSTD_colors(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None, 
     Returns:
         mask : boolean array, True if the object has colors like an FSTD
 
+    """
+
+    if primary is None:
+        primary = np.ones_like(gflux, dtype='?')
+    fstd = primary.copy()
+
+    # Clip to avoid warnings from negative numbers.
+    gflux = gflux.clip(0)
+    rflux = rflux.clip(0)
+    zflux = zflux.clip(0)
+    
+    # colors near BD+17
+    grcolor = 2.5 * np.log10(rflux / gflux)
+    rzcolor = 2.5 * np.log10(zflux / rflux)
+    fstd &= (grcolor - 0.32)**2 + (rzcolor - 0.13)**2 < 0.06**2
+
+    return fstd
+
+def isFSTD(gflux=None, rflux=None, zflux=None, primary=None, decam_fracflux=None,
+           objtype=None, decam_snr=None, bright=False):
+    """Select FSTD targets using color cuts and photometric quality cuts (PSF-like
+    and fracflux).  See isFSTD_colors() for additional info.
+
+    Args:
+        gflux, rflux, zflux, w1flux, w2flux: array_like
+          The flux in nano-maggies of g, r, z, w1, and w2 bands.
+        primary: array_like or None
+          If given, the BRICK_PRIMARY column of the catalogue.
+        bright: apply magnitude cuts for "bright" conditions; otherwise, choose
+          "normal" brightness standards.
+
+    Returns:
+        mask : boolean array, True if the object has colors like an FSTD
+
     Notes:
         The full FSTD target selection also includes PSF-like and fracflux
         cuts; this function is only cuts on the colors.
 
     """
-    #----- F-type standard stars
     if primary is None:
         primary = np.ones_like(gflux, dtype='?')
     fstd = primary.copy()
 
-    #- colors near BD+17; ignore warnings about flux<=0
-    with warnings.catch_warnings():
-        warnings.simplefilter('ignore')
-        grcolor = 2.5 * np.log10(rflux / gflux)
-        rzcolor = 2.5 * np.log10(zflux / rflux)
-        fstd &= (grcolor - 0.32)**2 + (rzcolor - 0.13)**2 < 0.06**2
+    # Apply the magnitude and color cuts.
+    fstd &= isFSTD_colors(primary=primary, zflux=zflux, rflux=rflux, gflux=gflux)
+
+    # Apply type=PSF, fracflux, and S/N cuts.
+    fstd &= _psflike(objtype)
+    for j in (1, 2, 4):  # g, r, z
+        fstd &= decam_fracflux[j] < 0.04
+        fstd &= decam_snr[..., j] > 10
+
+    # Observed flux; no Milky Way extinction
+    if obs_rflux is None:
+        obs_rflux = rflux
+
+    if bright:
+        rbright = 15.0
+        rfaint = 18.0
+    else:
+        rbright = 16.0
+        rfaint = 19.0
+        
+    fstd &= obs_rflux < 10**((22.5 - rbright)/2.5)
+    fstd &= obs_rflux > 10**((22.5 - rfaint)/2.5)
 
     return fstd
 
@@ -144,7 +193,7 @@ def isMWSSTAR_colors(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=Non
 
     return mwsstar
 
-def psflike(psftype):
+def _psflike(psftype):
     """ If the object is PSF """
     #- 'PSF' for astropy.io.fits; 'PSF ' for fitsio (sigh)
     #ADM fixed this in I/O.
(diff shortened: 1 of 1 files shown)


py/desitarget/cuts.py
```diff
@@ -94,22 +94,70 @@ def isFSTD_colors(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None,
...
         mask : boolean array, True if the object has colors like an FSTD
 
+    """
+
+    if primary is None:
+        primary = np.ones_like(gflux, dtype='?')
+    fstd = primary.copy()
+
+    # Clip to avoid warnings from negative numbers.
+    gflux = gflux.clip(0)
```
geordie666: Lines 104 to 110 look dangerous. For example, If `rflux < 0` then `rflux = 0` and then `np.log10(rflux / flux)` is -inf and potentially flags a RuntimeWarning. We probably want to clip to a number ever-so-slightly larger than 0? But then `rflux / gflux` is 1 and `grcolor` and `rzcolor` would incorrectly be 0. Is there a way to elegantly catch zero flux without having 0 color?

py/desitarget/cuts.py
```diff
@@ -94,22 +94,70 @@ def isFSTD_colors(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None,
...
-        grcolor = 2.5 * np.log10(rflux / gflux)
-        rzcolor = 2.5 * np.log10(zflux / rflux)
-        fstd &= (grcolor - 0.32)**2 + (rzcolor - 0.13)**2 < 0.06**2
+    # Apply the magnitude and color cuts.
+    fstd &= isFSTD_colors(primary=primary, zflux=zflux, rflux=rflux, gflux=gflux)
+
+    # Apply type=PSF, fracflux, and S/N cuts.
+    fstd &= _psflike(objtype)
+    for j in (1, 2, 4):  # g, r, z
+        fstd &= decam_fracflux[j] < 0.04
```
geordie666: Reviewing `decam_fracflux[j] < 0.04` I found some NaN's for `decam_fracflux`, e.g.:

```
from desitarget import io
import numpy as np
file = '/global/project/projectdirs/cosmo/data/legacysurvey/dr3/sweep/3.0/sweep-170p010-180p015.fits'
objects = io.read_tractor(file)
np.where(objects['DECAM_FRACFLUX'].T != objects['DECAM_FRACFLUX'].T)
(array([1, 1, 1, ..., 4, 4, 4]), array([  15436,   15461,   15467, ..., 1508219, 1508238, 1508379]))
```

Should we catch these NaNs? I'm not sure we can guarantee that `DECAM_FRACFLUX` will always be well-defined in the short-term? Or perhaps we should assume that `DECAM_FRACFLUX` should always be well-defined? In any case, I think this should throw an error, for now...

sbailey:
In general, we should be robust to the case of having NaN / inf / zero-flux / negative-flux in the input files.  In those cases I think the appropriate action for target selection is
* negative or zero-flux : I don't think this is an error condition, it just means that the target shouldn't pass cuts, but we should implement it in a way that doesn't generate a bunch of python RuntimeWarnings about invalid values in `np.log()`.
* NaNs or Infs or other genuinely invalid input values: print an informative warning so that we know that there is something funny in the input files, then don't pass the target (i.e. having a bad input isn't a free pass)
* but go ahead and process the other targets with good inputs

i.e. we shouldn't let a bug in a single target of a single sweep file cause target selection to throw an error and crash, and we should only generate warnings for things that we actually think need to be fixed.

If the current state generates extra RuntimeWarnings but doesn't actually crash, we could proceed with merging and cleanup the warnings later.  @moustakas and @geordie666 can negotiate that.

geordie666:
I _think_ that `RuntimeWarnings` always crash multi-threaded processes? So, e.g., `RuntimeWarning`s won't crash `select_targets` with nproc=1 but would crash `select_targets` with nproc > 1?

If that thought is correct, then we should make sure to not generate `RuntimeWarnings`...

moustakas:
Sorry, I'll try to get to these requested changes as soon as I can.  @forero, if you have any time the help would be welcome.

sbailey:
If a process or thread or the main process *raises* a RuntimeWarning, then it will crash the code, whether it is multithreaded or not.  However, the "RuntimeWarning" that we see from numpy is just a warning, not a raised exception.  I don't know the details of how numpy prints the RuntimeWarning without actually raising it as an exception, but empirically: 
```python
In [1]: import numpy as np
   ...: import multiprocessing.pool
   ...: 
   ...: def blat(x):
   ...:     return np.log(x)
   ...: 

In [2]: #- Multiple processes
   ...: p = multiprocessing.pool.Pool(4)
   ...: p.map(blat, range(-5,5))
   ...: 
/Users/sbailey/anaconda/bin/ipython:5: RuntimeWarning: invalid value encountered in log
  
/Users/sbailey/anaconda/bin/ipython:5: RuntimeWarning: invalid value encountered in log
  
/Users/sbailey/anaconda/bin/ipython:5: RuntimeWarning: invalid value encountered in log
  
/Users/sbailey/anaconda/bin/ipython:5: RuntimeWarning: invalid value encountered in log
  
/Users/sbailey/anaconda/bin/ipython:5: RuntimeWarning: divide by zero encountered in log
  
Out[2]: 
[nan,
 nan,
 nan,
 nan,
 nan,
 -inf,
 0.0,
 0.69314718055994529,
 1.0986122886681098,
 1.3862943611198906]

In [3]: #- Multiple Threads
   ...: t = multiprocessing.pool.ThreadPool(4)
   ...: t.map(blat, range(-5,5))
   ...: 
/Users/sbailey/anaconda/bin/ipython:5: RuntimeWarning: invalid value encountered in log
  
/Users/sbailey/anaconda/bin/ipython:5: RuntimeWarning: divide by zero encountered in log
  
Out[3]: 
[nan,
 nan,
 nan,
 nan,
 nan,
 -inf,
 0.0,
 0.69314718055994529,
 1.0986122886681098,
 1.3862943611198906]

In [4]: 
```
i.e. messages get printed, but the code doesn't actually crash and `nan` and `-inf` are returned for negative and 0 inputs to `np.log`.

geordie666:
Let's test this once NERSC is available again and I'll merge the PR if nothing is crashing `select_targets`. Perhaps I'm just wrong...but I'm fairly sure that something related to the flagging of `Runtime warnings` was killing `select_targets` on the `stdcuts` branch.

moustakas:
Thanks for the additional tests and input.  I think there's no reason to not catch these (known) invalid inputs, especially (obviously) if they cause a crash.  I removed all the suppression of the warnings (see the 'Files Changed' tab) in this PR because I wanted to expose these errors and to ensure we were being explicit about what errors were annoying and which were mission-critical.

sbailey:
NERSC isn't fully back, but I'm going to go ahead and merge this so that it doesn't get too out of sync with other in progress work on desitarget.  If we find problems when NERSC is back, we can re-open.

geordie666:
After this merge, I continue to find that `select_targets` is crashing, now on `master` seemingly due to `Runtime warnings` raising exceptions related to `np.log10(0)`. For example:

`select_targets /global/project/projectdirs/cosmo/data/legacysurvey/dr3/sweep/3.0/sweep-170p010-180p015.fits blah.fits -v`

Produces

```
During handling of the above exception, another exception occurred:

  File "/global/homes/a/adamyers/git/desitarget/bin/select_targets", line 45, in <module>
    qso_selection=ns.qsoselection, sandbox=ns.sandbox)
  File "/global/homes/a/adamyers/git/desitarget/py/desitarget/cuts.py", line 815, in select_targets
    targets = pool.map(_select_targets_file, infiles, reduce=_update_status)
  File "/global/homes/a/adamyers/git/desitarget/py/desitarget/internal/sharedmem.py", line 694, in map
    pg.join()
  File "/global/homes/a/adamyers/git/desitarget/py/desitarget/internal/sharedmem.py", line 407, in join
    raise SlaveException(*self.Errors.get())
desitarget.internal.sharedmem.SlaveException: divide by zero encountered in true_divide
  File "/global/homes/a/adamyers/git/desitarget/py/desitarget/internal/sharedmem.py", line 262, in _slaveMain
    self.main(self, *self.args)
  File "/global/homes/a/adamyers/git/desitarget/py/desitarget/internal/sharedmem.py", line 560, in _main
    r = realfunc(work)
  File "/global/homes/a/adamyers/git/desitarget/py/desitarget/internal/sharedmem.py", line 626, in realfunc
    else: return func(i)
  File "/global/homes/a/adamyers/git/desitarget/py/desitarget/cuts.py", line 780, in _select_targets_file
    desi_target, bgs_target, mws_target = apply_cuts(objects, qso_selection)
  File "/global/homes/a/adamyers/git/desitarget/py/desitarget/cuts.py", line 578, in apply_cuts
    obs_rflux=obs_rflux)
  File "/global/homes/a/adamyers/git/desitarget/py/desitarget/cuts.py", line 141, in isFSTD
    fstd &= isFSTD_colors(primary=primary, zflux=zflux, rflux=rflux, gflux=gflux)
  File "/global/homes/a/adamyers/git/desitarget/py/desitarget/cuts.py", line 109, in isFSTD_colors
    grcolor = 2.5 * np.log10(rflux / gflux)
RuntimeWarning: divide by zero encountered in true_divide
```
Am I the only one that raises this exception?

geordie666:
To somewhat answer my own question, I could explicitly fix this using a

```
with np.errstate(invalid='ignore',divide='ignore'):
```

but I don't understand why only _I_ appear to be concerned about being explicit about catching this `RuntimeWarning`...?

PR #147, opened 2017-02-08
Added ELG XD selection code for the first time. Potentially requires …

jaekor91:
…a lot of edits. The function (and only one) added is called apply_XD_globalerror.


geordie666:
@jaekor91: can we add these changes in `py/desitarget/sandbox/cuts.py` instead of `py/desitarget/cuts.py` until we're sure that the code is stable?

jaekor91:
Yes. In fact, I believe that is the area I was supposed to put the code. I will make the change. Thank you.

PR #148, opened 2017-02-08
Added ELG XD selection code for the first time. Potentially requires …

jaekor91:
…a lot of edits. The function (and only one) added is called apply_XD_globalerror.

### DIFF
--- a/bin/select_targets
+++ b/bin/select_targets
@@ -29,7 +29,6 @@ ap.add_argument("--numproc", type=int,
     help='number of concurrent processes to use [{}]'.format(nproc),
     default=nproc)
 
-
 ns = ap.parse_args()
 infiles = io.list_sweepfiles(ns.src)
 if len(infiles) == 0:
@@ -44,7 +43,7 @@ if ns.check:
     print('{} potentially corrupt files'.format(nbadfiles))
 else:
     targets = select_targets(infiles, numproc=ns.numproc, verbose=ns.verbose,
-                             qso_selection=ns.qsoselection, sandbox=ns.sandbox, FoMthresh=FoMthresh)
+                             qso_selection=ns.qsoselection, sandbox=ns.sandbox, FoMthresh=ns.FoMthresh)
 
     io.write_targets(ns.dest, targets, indir=ns.src,
         qso_selection=ns.qsoselection, sandboxcuts=ns.sandbox)
--- a/bin/select_targets
+++ b/bin/select_targets
@@ -21,12 +21,14 @@ ap.add_argument("dest", help="Output target selection file")
 ap.add_argument('-v', "--verbose", action='store_true')
 ap.add_argument('-c', "--check", action='store_true',help="Process tractor/sweeps files to check for corruption, without running full target selection")
 ap.add_argument("--sandbox", action='store_true',help="Apply the sandbox target selection algorithms.")
+ap.add_argument("--FoMthresh", type=float,help='XD Figure of Merit Threshold for an ELG (sandbox)')
 ap.add_argument('--qsoselection',choices=qso_selection_options,default='randomforest',
                 help="QSO target selection method")
 ### ap.add_argument('-b', "--bricklist", help='filename with list of bricknames to include')
 ap.add_argument("--numproc", type=int,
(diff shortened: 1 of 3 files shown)


geordie666:
"params" is a pretty large dictionary. Could it instead be a recarray that we store as a fits file in the directory:

/desitarget/py/desitarget/data/

Also, is params created from a training procedure? If so, we'd need an example of how to run that training procedure to create params in case we need to re-train and @jaekor91 isn't around...

jaekor91:
"params" is indeed a long array. Within the function is another way of accessing all the parameters by importing them from .npy files. There are 28 such files, four corresponding to each of seven classes (GMM properties + dNdm). The reason why we didn't use this was we were concerned about overhead associated with opening files every time we call the function. (Right now, it takes about 70 ms to run the selection over two square degrees of sky, not taking into account table load overhead. So, perhaps, this extra overhead is not a concern.) 

At the moment, I don't know of a simple way to store all these parameters in a single recarray, as the parameters of different classes (amplitudes, covariance matrices, and means of GMM and dNdm parameters) have different dimensions. Is there a simple way to do this?

From your previous e-mail, I see that Christophe saved a python script that demonstrates how he trained his model in "desitarget/py/desitarget/train". Would it be an option to simply store instead several html version of IPython notebooks, which are what I used to get the model parameters? I also have an accompanying document that describes a step-by-step process of how the parameters were obtained.

geordie666:
@jaekor91 Let's leave `params` as it is for now...when we have the final version of the code we can clean that up outside of `sandbox`.

Also, for now, an IPython notebook is fine, although I think we'd ultimately prefer an example Python script, again when we shift the code out of `sandbox` and make it "official".

I'm trying to get your code working inside the `apply_cuts` formalism. It's obvious to me what I pass to `apply_XD_globalerror` as `objs` but what should I pass as `last_FoM`?

geordie666:
I edited this in sandbox so that it works the way I think will be most useful for @jaekor91. At the command line you can write things like
```
select_targets /global/project/projectdirs/cosmo/data/legacysurvey/dr3/sweep/3.0/sweep-190p000-200p005.fits blah.fits -v --sandbox --numproc 1 --FoMthresh 0.2
```
and two files will be written, the standard targets file (as called "blah.fits" here) and a file called "FoM.fits" which is written to the working directory. FoM.fits contains the targets **_and_** the XD Figure of Merit for all of the ELG targets. 

Note that you can also pass the FoM threshold (what is called last_FoM in @jaekor91's code) using the --FoMthresh 0.2 flag, so that @jaekor91 can experiment with different cuts to optimize the ELG selection.

The code itself is still a little messy (hey, that's what the sandbox is for) but I think it's good enough to merge. I'd appreciate @moustakas having a quick look at this branch (and maybe issuing the above command) to check that I didn't break anything. If I didn't break anything, let's merge at this point.

jaekor91:
@geordie666 I can provide a trainer script that produces the parameters required for the selection algorithm given input truth catalogs. I will do this once we decide on how the function ultimately imports the parameters. 

For last_FoM, the user would provide an appropriate threshold value, which can be computed using another function given the same set of parameters as the selection. For the fiducial (or the default) set of parameters, it is 0.502. Perhaps, I should have made it an optional parameter and given this value. But then, we are still exploring what the "optimal" threshold value should be so perhaps this is better.

My initial reason for returning FoM array was so that I could adjust FoM threshold and observe its effect on the selection without having to run the code again. The current code requires the user to re-run the code every time FoM threshold is changed. Now that I think about it again, my original idea is not feasible as it would require importing all catalogs into RAM simultaneously! (However, I could still set the threshold artifically low and increase it to study the resulting variation on the selection.)

By the way, is it possible to pass other parameters as input such as f_i and reg_r?

@geordie666 Thank you for doing the most of the work integrating my code to desitarget. I'm looking forward to studying further the selection on NERSC once @moustakas has a chance to review the code.

moustakas:
I can confirm that the code runs for me as advertised and that the output files look good.  I'd like to complain about the nonconformity to [PEP8](https://www.python.org/dev/peps/pep-0008), but I won't.  :)

moustakas:
I'm leaving the branch open, @jaekor91, so you can add more optional input parameters as needed.

PR #153, opened 2017-02-22
First mature code for masking bright stars

geordie666:
Code to:

- collect bright stars from the sweeps in any set of bands
- create a bright star mask
- write a bright star mask
- apply a bright star mask to targets
- update targeting bits to indicate that something is `IN_BRIGHT_OBJECT`
- run all of the above functionality from `bin/select_targets`
- apply unit tests

Still to do in the short term:

- update MTL to prevent anything with `IN_BRIGHT_OBJECT` from ever being observed
- set `BRIGHT_OBJECT` as well as `IN_BRIGHT_OBJECT` by matching on `TARGETID`

Still to do in the long term:

- determine the correct star-radius relation for the masks (the code currently uses the SDSS relationship for Tycho stars). The "correct" relationship could be measured during commissioning?

But, this is probably already good enough for realistic masking for the data challenge (modulo setting priorities in MTL)


bin/select_targets
```diff
@@ -20,7 +21,18 @@ ap.add_argument("src", help="Tractor/sweeps file or root directory with tractor/
...
+                help='File containing bright object mask if -s (--starmask) is sent. If not sent, star mask is built from files in src directory')
+ap.add_argument("--starmaskbands", 
+                help='Bands to use to build the star mask if -s (--starmask) is sent but --starmaskfile is not (e.g. GRZ)',
+                default="GRZ")
+ap.add_argument("--starmaskmaglim",
+                help='Magnitude limits for building the star mask if -s (--starmask) is sent but --starmaskfile is not (e.g. 10,10,10)',
+                default="10,10,10")
+ap.add_argument("--starmaskoutfile",
+                help='Output file name to which to write star mask if -s (--starmask) is sent but --starmaskfile is not')
+ap.add_argument("--sandbox", action='store_true',help="Apply the sandbox target selection algorithms")
```
sbailey: This set of options seems to overload what `select_targets` is supposed to do, conflating selecting targets vs building bright star masks.  I'd prefer a separate executable focused on building the star mask and then have that mask file used as an input to select_targets.  Building the mask might grow in complexity and number of options and may eventually not even use the same inputs as `select_targets`.

py/desitarget/brightstar.py
```diff
@@ -0,0 +1,458 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# -*- coding: utf-8 -*-
+"""
+=====================
+desitarget.brightstar
+=====================
```
sbailey: Did you forget to add and commit a unit test for `desitarget.brightstar`?  You mention unit tests in the PR but the current unit tests never even import this file.

py/desitarget/cuts.py
```diff
@@ -6,6 +6,7 @@
 import sys
 
 import numpy as np
+np.seterr(all="ignore")
```
sbailey: Do we really want to turn all of them off here?  We seem to have come full circle on numpy error masking.  I originally caught and hid the warnings when I knew they were harmless (negative flux is legal but shouldn't pass cuts, so don't jabber about log of a negative number...).  That then got removed to expose all the warnings, and then re-added on a case-by-case basis.  But I think we want to keep the warnings trapping to isolated pieces of code and not a generic module level filter.  Thoughts?

py/desitarget/brightstar.py
```diff
@@ -0,0 +1,458 @@
...
+    starmask : :class:`recarray`
+        A recarray containing a bright star mask as made by desitarget.brightstar.make_bright_star_mask
+
+    Returns
+    -------
+        an ndarray of the updated desi_target bit that includes bright star information
+                                                                                                                                                      
+    To Do
+    -----
+        Currently sets IN_BRIGHT_OBJECT but should also match on the TARGETID to set BRIGHT_OBJECT bit
```
sbailey: Should also set `NEAR_BRIGHT_OBJECT` for a larger radius, or add that to the TODO list.

geordie666:
Thanks. I did forget to "git add" the unit test (oops)...will fix these as
soon as I get time today.

geordie666:
OK, I addressed all of @sbailey's requests, and added some extra code for plotting masks. I also fixed a DeprecationWarning in cuts.py that was crashing the 2.7 unit tests.

The main reason that the region @sbailey flagged contained so many bright stars was because I wasn't pre-cutting the mask on sources with `NOBS==0` (h/t Dustin Lang). The mask in that region now looks better:

maskregion

I verified that the masks fall on bright stars by uploading a mask file to the viewer at http://legacysurvey.org/viewer

The large radii for some of the remaining masks is caused by my current adoption of the SDSS radius-magnitude relationship for Tycho stars. We'll need some discussion of what is the "correct" radius, which may be tricky to ascertain without commissioning observations near bright stars.

PR #160, opened 2017-03-15
Fix overflow in LRG cuts in sandbox/cuts.py

npadmana:
The fluxes are all float32 arrays, and the cuts (as written below) raise these fluxes to the 3rd and 4th power. Since float32 has a maximum value of ~5e38, this can overflow if the flux is a few times 1.e9, raising a RuntimeWarning, which causes the code to abort.

We temporarily ignore these warnings, to let the code run to the end.

Tested on the DR3.1 sweeps on Edison --
```
select_targets --numproc 24 -v --sandbox /project/projectdirs/cosmo/data/legacysurvey/dr3.1/sweep/3.1 test.fits
```

### DIFF
--- a/py/desitarget/sandbox/cuts.py
+++ b/py/desitarget/sandbox/cuts.py
@@ -81,22 +81,26 @@ def isLRG_2016v3_colors(gflux=None, rflux=None, zflux=None, w1flux=None,
     lrg &= (zflux < 10**(0.4*2.5)*rflux) # r-z<2.5
     lrg &= (zflux > 10**(0.4*0.8)*rflux) # r-z>0.8
 
-    # This is the star-galaxy separation cut
-    # Wlrg = (z-W)-(r-z)/3 + 0.3 >0 , which is equiv to r+3*W < 4*z+0.9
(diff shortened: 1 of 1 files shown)


geordie666:
Ignoring warnings like this is fine for the `sandbox` but let's try to make sure that statements like `with np.errstate(...'ignore')` never propagate into `py/desitarget/sandbox/cuts.py`

npadmana:
@geordie666 --- Agreed in general. Although I'll note that, in this case, this just allows the infs to propagate. Do we have a preferred way of handling these? In the past, I've clipped fluxes both at the low and high end, but I'm not sure that is any better than letting the infs through.

geordie666:
Fair point...formally we should probably try and determine what are "bad values" and what are just "low values" and try to capture the bad values using an `np.where` instead of circumventing the `RuntimeWarning`. Then we could have a list of which warnings were allowed to be ignored and which should be captured. This is probably a wider discussion to have with Data Systems, though. 

For now, it's fine to do anything in the `sandbox`

Issue #169, opened 2017-04-05
connecting mocks to spectra: outstanding issues


moustakas:
PR #156 merged the first version of the code which connects mock catalogs to spectra (and the corresponding target and truth catalogs).  However, there were a number of outstanding issues (in no particular order):

- [X] Track down why no standard stars are being selected and written out.  *Comment* This bug was identified and fixed before #156 was merged.

- [X] Include velocity dispersion broadening in the galaxy spectra.  *Comment* After the speed-up to `desisim.templates` in https://github.com/desihub/desisim/pull/229 velocity dispersion broadening is now being included.

- [ ] Add contaminants to all the spectral classes.

- [ ] Validate the space-densities of the targets after target selection.  *Comment* We need some QA to ensure that the targeting criteria being applied in `desitarget.mock.selection` are correct, and that the resulting target catalogs have the right space-density.

- [ ] Add some basic unit tests.

- [ ] Resolve desihub/desisim#228 about lya_spectra not being generated at the correct redshift due to `z_wind` logic

- [X] file format fixes (fixed in PR #170)

- [ ] #159 : parallelize `select_mock_targets`

- [ ] Add shapes and sizes to the target catalogs.  *Comment*: Since morphology isn't going to be used for the 2% sprint (and possibly even the 20% challenge) this could be de-prioritized.

I'm certain I'm missing other issues so let's expand this list.

moustakas:
@sbailey and @forero have independently reported (off-list) a problem with generating the mock spectra with the current version of the code.  I've been able to reproduce the problem on edison with the setup and commands below.


```bash
source /project/projectdirs/desi/software/desi_environment.sh
module swap desiutil/1.9.3
module swap desispec/0.13.2
module swap desisim/master
module swap desitarget/master

mkdir -p $SCRATCH/datachallenge2017 && cd $SCRATCH/datachallenge2017
cp -i /scratch1/scratchdirs/ioannis/datachallenge2017/mock_input_newtest.yaml .

srun -N 1 -n 1 -c 24 -p debug -L SCRATCH,project \
time select_mock_targets -c mock_input_newtest.yaml -s 444 --nproc 24 --output_dir newtest
```

For the record the problem seems to be when the ELG spectra are generated, so it's perhaps related to the `desisim.templates` speed-up, but I'm looking into it.  Here are the last few lines of the log output:
```
[snip]
INFO:io.py:129:load_all_mocks: Loaded 9 mock catalog(s).

[...]

PR #173, opened 2017-04-13
add contaminants to mock.build.targets_truth and fix lots of small issues

moustakas:
This is an incremental PR to address some of the issues raised in #169.  It includes a first  (preliminary) effort of including contaminants and also adds many small improvements and fixes to `mock.build.targets_truth` (and the code it calls).

Contaminants identified using a new set of targeting bits, which I found to be necessary because of the extensive overlap between various target classes and contaminants.  

There's also a preliminary [notebook](https://github.com/desihub/desitarget/blob/qamock/doc/nb/mock-qa.ipynb) which generates some much-needed QA.  There are still many issues we'll want to fix (and which I'll document elsewhere) but this should be sufficient for the 2% sprint.

Still ToDo before merging:

* This depends on https://github.com/desihub/desisim/pull/234 so we should merge that first.
* I still need to test this at NERSC and check the time / memory usage, but I think it has improved.


moustakas:
Following up on the conversation on the telecon -- this shouldn't be merged quite yet until I've tested and timed at NERSC.  There's also some memory savings I'd like to try to implement.

sbailey:
desihub/desisim#234 has been merged, so I think dependencies are in place for this.  This code looks fine.  Go ahead and merge when you are satisfied with NERSC tests.

sbailey:
Also, please update doc/changes.rst before merging.

moustakas:
This is now reasonably functional so I'm merging now.

PR #174, opened 2017-04-15
Add Random Forest for ELG in sandbox

Cyeche:


py/desitarget/sandbox/cuts.py
```diff
@@ -512,9 +614,13 @@ def apply_sandbox_cuts(objects,FoMthresh=None):
                        primary=primary)
 
     if FoMthresh is not None:
-        elg, FoM = apply_XD_globalerror(objects, FoMthresh, glim=23.8, rlim=23.4, zlim=22.4, gr_ref=0.5,
+        if (FoMthresh!=2.0) :
```
moustakas: What is special about `FoMthresh=2.0`?  It looks like you're trying to distinguish between @jaekor91's XD algorithm and your new RF algorithm, but these should probably be totally separate, right?

Cyeche:
I have changed the strategy, if the threshold is negative, we use RF and the pcut is abs(threshold). It's a fast implementation. See my mail to know  how we may implement better it.

geordie666:
`select_targets --sandbox --FoMthresh 0.2` and `select_targets --sandbox --FoMthresh -0.2` both work for me on this branch without breaking anything, so in the spirit of the sandbox not having to adhere to any strict rules, @Cyeche can go ahead and merge this if ready.

geordie666:
Oh, but before merging you should add a one-line description of what you did to the `changes.rst` file in `desitarget/doc`

moustakas:
This PR was merged even though unit tests failed.  Looks like the problem is in the `changes.rst` file, which caused my tests to fail as well after I merged master.  I'll fix it on my branch (since I'm editing it anyway), but @Cyeche please be sure to wait for the tests to complete before merging!

Cyeche:
Ok I'm sorry about that! 

It would be useful that someone write a short recipe of what we have to do in DESI wiki. Otherwise only a few persons will contribute to the code development which is not the spirit of a scientific collaboration….

geordie666:
Another option is that we don't record updates to the sandbox code in
changes.rst, in the spirit that the sandbox code is just a testing area
that will not be publicized as an "official" part of the code. We also
don't, e.g., need unit tests for the sandbox code.

Ad.

Cyeche:
This is an option that we may consider. 

But anyway, for a non-expert like me, it would be very useful to have a recipe: what we have to do step by step, including very trivial information such as which kind of editor we have to use to edit a changes.rst file… 

Christophe. 


> Another option is that we don't record updates to the sandbox code in
> changes.rst, in the spirit that the sandbox code is just a testing area
> that will not be publicized as an "official" part of the code. We also
> don't, e.g., need unit tests for the sandbox code.
> 
> Ad.
>

PR #175, opened 2017-04-17
script to extract metadata from Lyman-alpha mocks

moustakas:
I wasn't sure where this script belonged, if anywhere, so I put it in `desitarget/etc/lyamock-metadata.py`. Having the metadata of the Lyman-alpha mocks without the spectra themselves is needed for #173.

Note that I didn't bother storing the parent simpleSpec file (for each object) nor the row--I just grab the coordinates, redshift, and g-band magnitude.

The output metadata table has been transferred to NERSC:
```
In [1]: import fitsio
In [2]: ff = fitsio.FITS('metadata-simpleSpec.fits')
In [3]: ff[1]
Out[3]: 
  file: metadata-simpleSpec.fits
  extension: 1
  type: BINARY_TBL
  extname: METADATA
  rows: 1526512
  column info:
    RA                  f8  
    DEC                 f8  
    Z                   f4  
    MAG_G               f4  
```

### DIFF
--- a/etc/lyamock-metadata.py
+++ b/etc/lyamock-metadata.py
@@ -1,0 +1,71 @@
+#!/usr/bin/env python
+
+"""Extract
+
+"""
+import os
+import numpy as np
+from glob import glob
+import fitsio
+
+from astropy.table import Table, Column, vstack
+
+from desispec.io.util import fitsheader, write_bintable
+from desiutil.log import get_logger
+log = get_logger()
+
+import multiprocessing
+nproc = max(1, multiprocessing.cpu_count() // 2)
+    
+def _lyapath():
+    return os.path.join(os.getenv('DESI_ROOT'), 'mocks', 'lya_forest', 'v0.0.2')
+
+def _read_lya(lyafile):
+    """Read the metadata from a single Lya file."""
+    log.info('Reading {}'.format(lyafile))
+    ra, dec, z, mag = [], [], [], []
+    ff = fitsio.FITS(lyafile)
+    ff = ff[1:len(ff)]
+    for h in ff:
+        head = h.read_header()
+        z.append(head['ZQSO'])
+        ra.append(head['RA'])
+        dec.append(head['DEC'])
+        mag.append(head['MAG_G'])
+
+    ra = np.array(ra)
+    dec = np.array(dec)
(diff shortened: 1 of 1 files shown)


moustakas:
Oops, I did mean to include EXTNAME -- I'll do so.

As for the other suggestions, if https://github.com/desihub/desisim/pull/234 is approved then we will never actually touch the Lyman-alpha forest spectra themselves -- we'll generate them on the fly.  The only thing we actually need from this mock is RA, DEC, and z since this mock has a BAO signal.  Or at least that's my understanding and the assumption I've been operating under in uncommitted changes I've been making to https://github.com/desihub/desitarget/pull/173.

We can have a conversation offline if that'd be more helpful, too.

sbailey:
I think we definitely will be using the Lyman-alpha mock spectra themselves and thus want traceability back to them.  Andreu's 1D code produces the right per-object power but not cross-object correlations.  That is useful for many studies, but ideally we should also generate a QSO + Lyman-alpha sample with cross-object Lyman-alpha correlations including BAO, i.e. using the skewers from Jean-Marc Le Goff's mocks.

i.e. one or both of us is misunderstanding the Lyman-alpha BAO content of J-M vs. Andreu's mocks and their intended use.  I'm booked for the rest of today but we can followup in person tomorrow if this remains unclear.

sbailey:
You already had EXTNAME=METADATA in there (as the original post showed, I just missed it, apologies).

I added a new EXTNAME=FILEMAP HDU with the mapping of MOCKFILE -> MOCKFILEID and added two columns to the original HDU: MOCKFILEID [0-27] and MOCKHDUNUM [1-60k].

I also added some safety features: `python lyamock-metadata.py -h` will print a usage message and exit instead of trying to overwrite the one-and-only `/global/project/projectdirs/desi/mocks/lya_forest/v0.0.2/metadata-simpleSpec.fits` output file, and input and output directories can be specified too.

I updated the original metadata-simpleSpec.fits file, while caching your original as metadata-simpleSpec.orig.fits in case I messed something up.

I'll merge after tests pass for Travis / github cleanliness, though this new file isn't covered by tests at all.

moustakas:
Thank you for the cleanup, @sbailey.

PR #176, opened 2017-04-24
Bright Star Masking

geordie666:
Full functionality to mask circular regions of the sky, including:

- generating SAFE/BADSKY locations around the mask perimeter
- correctly setting the target bits (including TARGETID) for SAFE/BADSKY sky locations
- adding TARGETID-relevant bits to targetmask.yaml
- setting a NEAR_RADIUS warning for objects close to (but not in) a mask
- corrected code for plotting mask shapes using ellipses at Decs far from 0
- functions for calculating the area of a mask and retrieving the BRICKID (now from desipec)
- complete unit test and documentation


py/desitarget/brightstar.py
```diff
@@ -453,22 +660,208 @@ def is_in_bright_star(targs,starmask):
...
+    """Append targets at SAFE (BADSKY) locations to target list, set bits in TARGETID and DESI_TARGET
+
+    Parameters
+    ----------
+    targs : :class:`recarray`
+        A recarray of targets as made by desitarget.cuts.select_targets
+    starmask : :class:`recarray`
+        A recarray containing a bright star mask as made by desitarget.brightstar.make_bright_star_mask
+    drstring : :class:`str`, optional
+        The imaging data release for the targets as a 9-string (i.e. 'dr3      ') to update TARGETID
```
sbailey: Is trailing space padding required, and if so why?  Should this be a string or an integer?  Is it used other than for turning into an integer for including in a targetid ?

py/desitarget/brightstar.py
```diff
@@ -453,22 +660,208 @@ def is_in_bright_star(targs,starmask):
...
+        The original recarray of targets (targs) is returned with additional SAFE (BADSKY) targets appended to it
+
+    Notes
+    -----
+        - See the Tech Note at https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=2346 for more details
+          on the SAFE (BADSKY) locations
+        - See the Tech Note at https://desi.lbl.gov/DocDB/cgi-bin/private/RetrieveFile?docid=2348 for more details
+          on setting the SKY bit in TARGETID
+        - Currently hard-coded to create an additional 10,000 safe locations per sq. deg. of mask. What is the 
+          correct number per sq. deg. (Npersqdeg) for DESI is an open question.
```
sbailey: Can you explain the concept behind Npersqdeg more?  I thought every mask would get a set of safe locations at its perimeter and that if a density was specified it would be a linear density (i.e. how far apart on the perimeter can they be?) rather than an area density.

py/desitarget/data/targetmask.yaml
```diff
@@ -24,14 +24,14 @@ desi_mask:
     - [STD_WD,      34, "White Dwarf stars", {obsconditions: DARK|GRAY}]
     - [STD_BRIGHT,  35, "F-type standard for BRIGHT conditions",
         {obsconditions: BRIGHT}]
+    - [BADSKY,      36, "Blank sky locations that are imperfect but still useable",
+        {obsconditions: DARK|GRAY|BRIGHT|POOR|TWILIGHT12|TWILIGHT18}]
```
sbailey: I know that mask bits aren't supposed to have meaningful numbers, but was there a specific reason for moving this from bit 53 near the other BRIGHT_OBJECT bits to bit 36 near the standard star bits?  I expect our standards bits might expand somewhat and it will seem odd to have a BADSKY bit sitting in the middle of them.

py/desitarget/data/targetmask.yaml
```diff
@@ -144,6 +144,14 @@ obsmask:
     - [MORE_ZGOOD,  4, "redshift known; need more observations"]
     - [DONOTOBSERVE,  16, "Do not observe this target (possibly temporarily)"]
 
+#- Bits that can be set in TARGETID. See the schema at
+#- https://desi.lbl.gov/DocDB/cgi-bin/private/RetrieveFile?docid=2348
+targetid_mask:
+    - [BRICKID,    22, "RIGHTMOST bit that indicates BRICKID (BRICKID encompasses bits 22-41 inclusive"]
+    - [DR,         42, "RIGHTMOST bit that indicates Data Release (DR encompasses bits 42-47 inclusive"]
+    - [MOCK,       48, "this target is a mock target"]
+    - [SKY,        49, "this target is a sky fiber target"]
```
sbailey: Might as well include `OBJID` in here as well to give the complete specification of targetid bits.  Clarify whether you are counting from 0 or 1 (I hope zero).  Should this be encoded as `[NAME, FIRSTBIT, NBITS, "Description"]` ?  That over constrains the problem, but might be easier for code to deal with (e.g. it doesn't have to know that `DR` comes after `BRICKID` and use the combination of them to figure out how many bits are reserved for `BRICKID`.  A unit test could confirm that all bits are accounted for and non-overlapping (especially if you also add a `RESERVED` section at the end).

py/desitarget/brightstar.py
geordie666: This is purely because we specified this as a 9-string here:

https://desi.lbl.gov/DocDB/cgi-bin/private/RetrieveFile?docid=2348;filename=targetID.pdf;version=1

(see the final section). We could pass this as an integer here, if you'd prefer, and convert it to a string/pad it for other purposes? We'll need both, though (one for the `TARGETCAT` hash and one for bit-encoding.

py/desitarget/brightstar.py
geordie666: This is probably a "choice". Is the minimum fiber separation (really) constrained by an _angular distance_ (then how-far-apart-on-the-perimeter makes sense), or by a _fiber density_ (in which case Npersqdeg-area-of-the mask makes sense)?

py/desitarget/data/targetmask.yaml
geordie666: I moved this because I started to think of the meaning of the bit more as a (bad) sky fiber location rather than as a masked location. I believe that we will need to have other BADSKY locations that are unrelated to masks, according to the argument laid out in this TechNote:

https://desi.lbl.gov/DocDB/cgi-bin/private/RetrieveFile?docid=2345;filename=blank-skies.pdf;version=3

I'm certainly happy to return this to bit 53, if you think it makes more logical sense, though.

py/desitarget/data/targetmask.yaml
geordie666: I agree. I think it makes more sense to code this as `[NAME, FIRSTBIT, NBITS, "Description"]`. I'll make this change once we agree on the other changes.

geordie666: Oh, and do you have a preferred form for documented examples? Jupyter notebook?

geordie666: And, finally, I could add a mask_randoms in /bin, or split out mask_targets from select_targets in /bin if you think that would be more transparent?

py/desitarget/brightstar.py
sbailey: Let's either use an integer or a non-space-padded string, but *not* require exactly 9 characters at the python function API level.  A string could give us slightly more flexibility, e.g. for testing with non-'dr{n}' values.

py/desitarget/brightstar.py
sbailey: I'm just confused about *where* the BADSKY targets are placed when specifying them as a density and whether that implies that you are thinking about their distribution differently than I am. Presumably they are not distributed interior to the mask, but I also don't think we need them anywhere other than the perimeter outside of the mask (except for extended bright objects which is a different use of this bit).  The main point is that any fiber should be able to reach a BADSKY location if it has to.

py/desitarget/data/targetmask.yaml
sbailey: Jupyter is fine, or rst with example code.

It seems premature to add a `bin/mask_randoms` executable.  Examples showing how to call the underlying python functions would probably be more useful at this point.

Using the mask within `select_targets` is fine, as long as it is sufficiently structured and documented such that the mask can also be used outside of `select_targets`.  i.e. no need to make a separate `mask_targets` executable script.

py/desitarget/brightstar.py
geordie666: OK, sounds good...I'll rewrite this for a non-space-padded string or the form 'CCI' and/or 'CCII' (C for character, I for integer)

py/desitarget/brightstar.py
geordie666: The BADSKY targets are definitely placed on the perimeter of the mask. How many you want to be generated is calculated as 

Npersqdeg x area_of_the_mask

So, if I were to specify the DESI SKY fiber density to be 100 fibers per sq. deg., then a mask that was 0.1 sq. deg. in area would have 10 (equally spaced) locations generated around its perimeter.

py/desitarget/data/targetmask.yaml
geordie666: Sounds good...I'll document the code accordingly

sbailey:
Changes look fine, merging now, with a few comments:

* I had completely forgotten about the "extra" dictionary feature of bitmasks that you used for `targetid_mask.OBJID.firstbit`, `targetid_mask.OBJID.nbits`, etc.  Note that `targetid_mask.OBJID.firstbit` is identical to `targetid_mask.OBJID.bitnum` so is a bit redundant, but ok.
* The mask examples notebook is great.  Thanks for adding that.
* Cell 125 illustrates the issue I was trying to describe about parameterizing a linear density vs. an area density of BADSKY locations around a mask: the largest mask seems to have way more than it needs compared to the smallest mask if you think about it as a positioner at the edge of a mask needing to be about to reach something.  This may not matter in practice, but if we find ourselves with too many for some masks but too few for others, we could revisit this parameterization.

PR #177, opened 2017-04-26
reduce the memory usage of the mock-->spectra code

moustakas:
This is a follow-up to #173 in which `desitarget.mock.targets_truth` is now approximately 5%-15% faster than what's in master and should use significantly less memory.

This speed-up is accomplished largely by being smarter about (not) simulating the FAINTSTAR contaminants until the end.  Specifically, I use the stellar templates themselves to figure out which faint stars are going to pass target selection cuts (on a given brick with the simulated noisy photometry) and only generate spectra for those that do.  

In addition, I've reduced the memory usage *significantly* by applying target selection on each individual brick rather than keeping all the targets (and spectra!) until all the bricks were done. I also don't bother creating any (empty!) sky spectra, and I only load one mock at a time rather than holding them all in memory simultaneously.  This should allow us to scale over a much larger chunk of sky than what we're currently doing.  

Some other notes:
* The output directory is required to be empty and the user must explicitly set `--clobber` in `select_mock_targets` if it isn't.  This is to make sure old files don't mix in with new ones, especially since we're calling this code over different patches of sky.
* I toyed with the idea of writing out intermediate files/spectra and then gathering them up at the end but in the end this wasn't necessary.
* `map_id_filename.txt` isn't being written because the connection back to the mocks is broken anyway.  I'll fix this in a separate PR (or here, if requested).

### DIFF
--- a/bin/select_mock_targets
+++ b/bin/select_mock_targets
@@ -22,6 +22,7 @@ parser.add_argument('--bricksize', '-b', help='Size of the imaging bricks (deg)'
 parser.add_argument('--outbricksize', '-o', help='Desired size of the output bricks', type=float, default=0.25)
 parser.add_argument('--nproc', type=int, help='number of concurrent processes to use [{}]'.format(nproc), default=nproc)
 
+parser.add_argument('--clobber', action='store_true', help='Remove intermediate files.')
 parser.add_argument('-v','--verbose', action='store_true', help='Enable verbose output.')
 args = parser.parse_args()
 
@@ -50,7 +51,7 @@ else:
 # Construct Targets and Truth files
 targets_truth(params, args.output_dir, realtargets=realtargets, seed=args.seed,
               verbose=args.verbose, nproc=args.nproc, bricksize=args.bricksize,
-              outbricksize=args.outbricksize)
+              outbricksize=args.outbricksize, clobber=args.clobber)
 
 log.info('All done!')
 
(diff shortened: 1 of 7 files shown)


moustakas:
For the record using this branch I ran an ~11.5 deg2 patch of sky on edison with 24 cores in approximately 10m30s.

```bash
source /project/projectdirs/desi/software/desi_environment.sh
module swap desiutil/1.9.3
module swap desispec/0.14.0
module swap desisim/master
module unload desitarget

mkdir -p $SCRATCH/dc2017 && cd $SCRATCH/dc2017

git clone git@github.com:desihub/desitarget
cd desitarget && git checkout memorymock && cd ..

export PYTHONPATH=$PWD/desitarget/py:$PYTHONPATH
export PATH=$PWD/desitarget/bin:$PATH

cp -i /scratch1/scratchdirs/ioannis/dc2017/qatargets_input.yaml .

srun -N 1 -n 1 -c 24 -p debug -L SCRATCH,project time select_mock_targets \
  -c qatargets_input.yaml -s 222 --nproc 24 --output_dir qatargets
```

sbailey:
Looks good, thanks.  Please add a comment to changes.rst and then go ahead and merge.

Did you want to leave the disabled "temporary debugging plot" code in mocks/build.py?  OK to leave it there if you think you will reenable it during future debugging, but if you are completely done with it let's remove that.

PR #179, opened 2017-05-24
Update to LRG target selection code

geordie666:
Moving the most recent Eisenstein/Dawson LRG code over from the sandbox so it's our default until the TS WG indicates otherwise.


geordie666:
Resulting LRG selection has about 590 targets per sq. deg. when run on DR3.1

moustakas:
Can you point me to the output files of running this new selection on DR3.1?  I need to make sure the current set of LRG templates span the color-space of this new sample.

geordie666:
Sure...I'll send you that link over email.

moustakas:
With apologies for the delay, I'm looking at this now.  @geordie666 could you fix the conflict(s) in the `doc/changes.rst` file?  

Also note that we (someone not me!) will need to update the nominal redshift distribution in `desimodel/data/targets/nz_lrg.dat` with the one [here](https://desi.lbl.gov/trac/wiki/TargetSelectionWG/LRGretuning).

moustakas:
@geordie666 I resolved the conflict in `changes.rst` and I added the following notebook for our own edification:
  https://github.com/desihub/desitarget/blob/LRGupdatesADM/doc/nb/lrg-color-color.ipynb

I don't think that any shortcomings in the LRG templates (and there are some..) should hold up this PR, but these comparisons are good to see and will help inform the next set of template updates.  

I'll also share the notebook on-list for input and comments.

moustakas:
@geordie666 Can you or Kyle update the n(z) for this selection?
  https://github.com/desihub/desitarget/pull/179#issuecomment-306181711

Are we ready to merge this?  I'm working on updated templates but what I'm using are good enough for now.

geordie666:
If all you need is an update to the N(z) in desimodel in SVN, then I can do
this as soon as we merge (or at least, within a day)...

Ad.

PR #181, opened 2017-05-26
various bug fixes and enhancements in the mocks-->spectra, including healpix outputs and new QA

moustakas:
Opening this PR now although it's not ready to merge.  This branch fixes a variety of bugs and issues which were leading to lower-than-expected ELG and standard-star target densities and it also adds a new QA module for visualizing the results.  We can likely reuse a lot of this code for 
#21.  

Output files are also now grouped by healpix pixels using the same hierarchical structure @tskisner is experimenting with (thanks to @sbailey for the code snippets).  

This link is obviously a throw-away and still work in progress based on a small 9 deg2 test area on my laptop but comments on the kinds of plots you'd like to see are welcome:
  http://www.sos.siena.edu/~jmoustakas/research/datachallenge/qa

Note that the lower-than-expected ELG and QSO target densities shown is a known/understood issue related to the *contaminants* not the science targets themselves.  I'll show this explicitly in the next iteration of plots.


py/desitarget/mock/QA.py
```diff
@@ -0,0 +1,220 @@
...
+Generate QA figures from the output of desitarget.mock.join_targets_truth.
+
+"""
+from __future__ import (absolute_import, division, print_function)
+
+import os
+import numpy as np
+import warnings
+
+def target_density(cat):
```
sbailey: pls update docstrings to define inputs.  e.g. here, that `cat` is a table with columns RA and DEC.

py/desitarget/mock/spectra.py
```diff
@@ -251,6 +330,89 @@ def elg(self, data, index=None, mockformat='gaussianfield'):
         else:
             raise ValueError('Unrecognized mockformat {}!'.format(mockformat))
 
+        if False:
```
sbailey: If you are done with this test code, let's remove it.

Issue #183, opened 2017-06-01
observing priorities not initialized


forero:
After reading the targetmask [yaml file](https://github.com/desihub/desitarget/blob/master/py/desitarget/data/targetmask.yaml) the [target mask initialization](https://github.com/desihub/desitarget/blob/master/py/desitarget/targetmask.py#L51)  sets
the priorities for targets 'DONE', 'OBS' and 'DONOTOBSERVE' to `1`:

```python
In [1]: from desitarget import desi_mask, bgs_mask, mws_mask

In [2]: desi_mask['ELG'].priorities
Out[2]: 
{'DONE': 1,
 'DONOTOBSERVE': 1,
 'MORE_ZGOOD': 3000,
 'MORE_ZWARN': 3000,
 'OBS': 1,
 'UNOBS': 3000}

In [3]: desi_mask['LRG'].priorities
Out[3]: 
{'DONE': 1,
 'DONOTOBSERVE': 1,
 'MORE_ZGOOD': 3200,
 'MORE_ZWARN': 3200,
 'OBS': 1,
 'UNOBS': 3200}

In [4]: desi_mask['QSO'].priorities
Out[4]: 
{'DONE': 1,
 'DONOTOBSERVE': 1,
 'MORE_ZGOOD': 3500,
 'MORE_ZWARN': 3400,
 'OBS': 1,
 'UNOBS': 3400}
```
We should explicitly set those priorities to other values different from `1`. What should we do?

sbailey:
Thanks for tracking this down.  It will somewhat clutter the yaml file, but it probably would be best to be explicit about these defaults.  Let's use:
```
DONOTOBSERVE: 0
OBS: 1
DONE: 2
```
That keeps all those states at low priority while still distinguishing them and allowing low-priority reobservation of observed targets that haven't been flagged as DONOTOBSERVE.

moustakas:
Fixed in #184.

PR #184, opened 2017-06-01
explicit initialization of observing priorities

forero:
Solves #183


sbailey:
Does MWS use `MORE_ZGOOD` and `MORE_ZWARN`, i.e. should those be added too as explicit values instead of implied defaults?

Tests are failing; please check why.
```
FAIL: test_zcat (desitarget.test.test_mtl.TestMTL)
  File "/home/travis/build/desihub/desitarget/py/desitarget/test/test_mtl.py", line 52, in test_zcat
    self.assertTrue(np.all(mtl['NUMOBS_MORE'] == [0, 1, 0, 3, 1]))
AssertionError: False is not true
```

forero:
The test pass now. It was a problem in mtl with the inclusion of  new `priority` values of 2. 
I don't know about MORE_ZGOOD and MORE_ZWARN usage in MWS. Perhaps @apcooper can help us with that question?

apcooper:
Right now we don't have plans for either MORE_ZWARN or MORE_ZGOOD, but thanks for asking, it's another thing we need to discuss at the collaboration meeting. 

Personal opinion: I doubt we'd reobserve MWS_MAIN targets under any circumstances. If we're going to reobserve by default it would most likely be for MWS_WD and MWS_NEARBY.

PR #186, opened 2017-06-06
more debugging and support for healpix input to select_mock_targets

moustakas:
This PR is a follow-up to #181 which ~fully implements processing of the mock spectra based on an input set of healpixels rather than RA,Dec bounds and fixes a number of small I/O bugs.  The code here is also needed by https://github.com/desihub/two_percent_DESI/pull/2.

Outstanding issues I'd like to address (~today) before merging this:
* Fix failing Travis tests, which I don't understand.  [Here](https://travis-ci.org/desihub/desitarget/jobs/239869890) is a recent failure which complains that `astropy.units` is not found when loading `desimodel.focalplane`!
* Test and time the results at NERSC over a reasonably sized chunk of sky and inspect the QA for gross problems.


weaverba137:
Add `'astropy.units'` to `autodoc_mock_imports` in doc/conf.py.  Other import errors can be handled similarly.

sbailey:
Documenting some timing information:

At NERSC, it is important to set `OMP_NUM_THREADS=1` otherwise you can end up with 24 processes all trying to use 24 cores each for some underlying MKL operations and it bogs down.

Trying a test case with just ELGs and nside=32 pixels (3.36 deg^2):
* single pixel: 3m47s
* 2 pixels: 5m59s
* 4 pixels: 10m53s

I added a few timing printouts and found that it is spending ~75% of its time in sharedmem (i.e. multiprocessing) parallelized code, with end-to-end effective CPU utilization of 20/24 from /usr/bin/time.

Timing on cori is faster (2m10 vs. 3m47), though oddly using ncpu=32 and ncpu=16 had about the same timing.  On edison, ncpu=12 was 5m52 vs. ncpu=24 at 3m47.

There are 1021 nside=32 pixels that touch a 2% survey tile, so to process all of them one pixel at a time would take ~3.8 * 1021 / 60 = 65 edison node hours, or ~130 nodes in a debug job for 30 minutes if the I/O can keep up with that many processes reading in mocks, templates, etc.

Going to nside=64, there are 3216 pixels touching the 2% survey (i.e. fewer than 4*1021), but I haven't timed 4 nside=64 pixels vs. 1 nside=32 pixel for overhead comparison.

Example setup and command to run:
```
module use /global/common/edison/contrib/desi/modulefiles
module load desiconda/20170510-1.1.3-spectro
module load desimodules/test-master
module swap desiutil/1.9.4
module swap desispec/0.14.0
module swap desisim/master

export HEALTEST=$SCRATCH/healtest
mkdir -p $HEALTEST && cd $HEALTEST

git clone git@github.com:desihub/desitarget
cd desitarget && git checkout sky-and-QA && cd ..

export PYTHONPATH=`pwd`/desitarget/py:$PYTHONPATH
export PATH=`pwd`/desitarget/bin:$PATH
export OMP_NUM_THREADS=1

cp /scratch1/scratchdirs/ioannis/healtest/healtest.yaml .
srun -N 1 -n 1 -c 24 -p debug -L SCRATCH,project time
select_mock_targets -c healtest.yaml --nproc 24 --verbose --seed 1
--healpixels 6507 --nside 32
```

sbailey:
paranoia: have you explicitly checked the random number seed distribution, i.e. that different pixels get different spectra when run separately but with the same seed; and if you process multiple pixels within the same call to `select_mock_targets` that you also get different spectra?

moustakas:
I confirmed that the behavior of the random seed is as expected.  

Also note that I made a copy of the mocks into my $SCRATCH space on cori
  `/global/cscratch1/sd/ioannis/desi/mocks` 
as well as on edison 
  `/scratch1/scratchdirs/ioannis/desi/mocks`

moustakas:
With the same basic environment setup on cori (but using the mocks on $SCRATCH), running just ELGs with a single nside=32 healpixel with
```
time srun -N 1 -n 1 -c 24 -p debug -L SCRATCH -C haswell select_mock_targets -c healtest.yaml --nproc 24 --verbose --seed 1 --healpixels 6507 --nside 32
```
takes 6m23s vs your 2m10s.  So something is amiss that I don't understand.

moustakas:
OK, I think we should merge this branch with #187 or vice versa and then merge to master.  I've added the Lya skewers back in and a minimal amount of velocity broadening for the ELG, LRG, and BGS target classes.  Both these changes have added computational overhead (not sure how much but perhaps as much as 10%-20%  for each class, although at the expense of more realistic spectra), so our timing tests should be realistic as far as the Data Challenge is concerned.

@sbailey I have a ~hour window this afternoon if you want to coordinate.  Ping me on slack or email.

moustakas:
With this branch I ran all target classes through one nside=64 pixel in 10m10s with 4 cores on my laptop, which is kinda long...  But at least it works! 

This is ready to merge --

PR #187, opened 2017-06-08
MPI version of select_mock_targets

sbailey:
Note: this PR is from the "sjb" branch into John's "sky-and-QA" branch, not into master itself.

This PR adds `desitarget/bin/mpi_select_mock_targets` which is an MPI-parallelized version of `select_mock_targets`.  It takes a file with TILEIDs as input (e.g. twopct.ecsv), calculates which pixels overlap those tiles, checks the output directory structure for any pixels that have already been processed and removes those, and then distributes the remaining pixels among the MPI ranks.  Each rank writes out pixels in groups of 4 as it finishes them such that if the job times out before completing, you can just resubmit the same command and it will pick up where it left off.

This also reorganizes the output structure to `{outdir}/{group}/{pix}/*-{nside}-{pix}.fits` where `group = pix//100`.  That choice is isolated into `mock.io.get_healpix_dir()` and `mock.io.findfile()` if we want to change it.

Example usage on 12 nodes, running 2 ranks per node (the underlying `targets_truth` function that this wraps is additionally multiprocessing parallelized over targets):
```
export OMP_NUM_THREADS=1
export MPICH_GNI_FORK_MODE=FULLCOPY
export KMP_AFFINITY=disabled
srun -N 12 -n 24 -c 16 python $DESITARGET/bin/mpi_select_mock_targets \
    --tiles twopct.ecsv -c healtest.yaml \
    --nproc 16 --seed 1 --nside 64
```

Timing on Cori is a little over a minute per nside=64 pixel per half-node, such that doing the 3216 nside=64 pixels that cover the 2% survey footprint should take ~64 node-hours, plus some additional I/O overhead when 128 processes are working in parallel.

github is showing some commits in this PR that came from John rather than me.  I don't understand that: I intended to cleanly branch off of John's branch, make my changes, and then merge back into John's branch.  Before merging, please confirm if the changes shown here to `select_mock_targets` (not mpi_...) and `selection.py` are supposed to be there, or accidentally reverting something to the wrong state.


moustakas:
This looks great.  I'm anxious to test but I'm not quite done committing to my `sky-and-QA` branch (still adding the Lyman-alpha skewers back in) so I'm not sure what effect merging this branch to master will have, since most of the changes are from my branch.  And what happens as I continue to commit to my branch?

moustakas:
@sbailey Could you either paste an example SLURM script that calls the new `mpi_select_mock_targets` script or point me to one on-disk?

sbailey:
Example configuration and usage via an interactive job on cori:
```
#- Load a basic environment; drop desitarget
module use /global/common/cori/contrib/desi/modulefiles/
module load desimodules/test-master
module unload desitarget

#- Get a copy of desitarget branch sjb
cd $SCRATCH/test/
git clone https://github.com/desihub/desitarget
cd desitarget && git checkout sjb && cd ..
export PYTHONPATH=$(pwd)/desitarget/py:$PYTHONPATH
export PATH=$(pwd)/desitarget/bin:$PATH

#- Grab config files
cp /global/cscratch1/sd/sjbailey/desi/dc17b/twopct.yaml .
cp /global/cscratch1/sd/sjbailey/desi/dc17b/twopct.ecsv .

#- Run job
export OMP_NUM_THREADS=1
export MPICH_GNI_FORK_MODE=FULLCOPY
export KMP_AFFINITY=disabled
salloc -N 1 -t 00:30:00 -C haswell --qos interactive

srun -N 1 -n 2 -c 16 mpi_select_mock_targets \
    --tiles twopct.ecsv -c twopct.yaml \
    --nproc 16 --seed 1 --nside 64
```
First output should appear in ~5 minutes.  The code currently doesn't have an option for limiting how many tiles or pixels are processed, so this will just run until you ^C or the job times out.  We could add `--ntiles` or `--npix` options to help with testing.

sbailey:
Don't merge yet...  My example ran with lots of log messages, but didn't produce any output within ~5 minutes.  I'm tied up in a meeting but want to double check that before merging.

moustakas:
I won't merge this until we've touched base.  But should we merge #186 first?  I'm a *teensy* bit worried about how the two branches are interacting (or not).

Lyman-alpha skewers being read and generated by default, which has added I/O overhead (although I think it's reasonably optimized given the awkward format of the Lya mocks!).

sbailey:
Two updates:
* when starting from the beginning, I'm hitting memory problems with the example as posted; previously this had worked because I had first some some single node tests that did big bricks, and then my 2/node tests were working on smaller bricks.
* I accidentally ended up with this PR aiming at master instead of John's branch.  I will close this and reopen a new PR with the correct target (I hope...)

PR #189, opened 2017-06-12
Update to Data Model for DR4 style files

geordie666:
Significant updated to handle transition from pre-DR4 to post-DR4 data model. Includes:
   
- desitarget.io can now read old DR3-style and new DR4-style tractor and sweeps files
- desitarget.cuts now always uses DR4-style column names and formats
- desitarget.brightstar builds masks on DR4-style data, and uses `RELEASE` to set the Data Release instead of passing around the DR as a string
- HEALPix pixel number (currently nside=64) is added to output target files and header information is included indicating the pixels side and that `NEST` is `True`
- `select_targets` now passes around information related to `HEALPix`
- any unit tests that explicitly used columns and formats from the data model have been updated to the DR4+ data model, as have the test files that the unit tests rely on


moustakas:
@geordie666 How high priority is it to merge this sooner rather than later?  I'm trying to gauge what I need to get done before the collaboration meeting.

geordie666:
We can discuss that on the telecon today. As the final DR4 files aren't yet in place, I suspect it can wait awhile.

sbailey:
On the telecon we agreed to wait until after the collaboration meeting to merge this, to avoid breaking things.  Prior to merging we should also test desitargets.mocks, desisim, and desispec with this branch to know what if anything it will break.  If it does break those codes, we should try to do a coordinated update of all packages.

geordie666:
One thing I didn't add, which we could during a coordinated effort, is the mapping of imaging RELEASE to the _SOUTH and _NORTH targeting bits. I don't think that's much of an extension to this branch (a RELEASE -> N/S mapping table in desitarget.cuts)...but it might be worth seeing what I've broken downstream before breaking yet more.

geordie666:
I included some additional functionality while unit test issues were being examined. I added a `RELEASE` to `PHOTSYS` ('N' or 'S') mapping mechanism, and updated the `TARGETID` to incorporate the new DR4(+) `RELEASE` column.

weaverba137:
I have just copied the latest changes from desihub/desispec#416 to see if that fixes the Travis errors.

weaverba137:
Your Travis problems are fixed.  Merge when ready.

py/desitarget/io.py
```diff
@@ -27,18 +34,107 @@
...
+    ('MW_TRANSMISSION_G', '>f4'), ('MW_TRANSMISSION_R', '>f4'), ('MW_TRANSMISSION_Z', '>f4'), 
+    ('FRACFLUX_G', '>f4'), ('FRACFLUX_R', '>f4'), ('FRACFLUX_Z', '>f4'), 
+    ('NOBS_G', '>i2'), ('NOBS_R', '>i2'), ('NOBS_Z', '>i2'), 
+    ('PSFDEPTH_G', '>f4'), ('PSFDEPTH_R', '>f4'), ('PSFDEPTH_Z', '>f4'), 
+    ('GALDEPTH_G', '>f4'), ('GALDEPTH_R', '>f4'), ('GALDEPTH_Z', '>f4'), 
+    ('FLUX_W1', '>f4'), ('FLUX_W2', '>f4'), ('FLUX_W3', '>f4'), ('FLUX_W4', '>f4'), 
+    ('FLUX_IVAR_W1', '>f4'), ('FLUX_IVAR_W2', '>f4'), ('FLUX_IVAR_W3', '>f4'), ('FLUX_IVAR_W4', '>f4'), 
+    ('MW_TRANSMISSION_W1', '>f4'), ('MW_TRANSMISSION_W2', '>f4'), 
+    ('MW_TRANSMISSION_W3', '>f4'), ('MW_TRANSMISSION_W4', '>f4'), 
+    ('SHAPEDEV_R', '>f4'), ('SHAPEEXP_R', '>f4'), ('DCHISQ', '>f4', (5,))
```
moustakas: Would we ever need the inverse variances on the sizes (`shapedev_r_ivar` and `shapeexp_r_ivar`, I think)?  What about `fracdev`?  

Also, what do we think about removing the `nobs_[g,r,z]` columns, which I added about six months ago to help identify regions with >2 passes.  Presumably the *depth* columns contain this information and more.

geordie666:
I'll refactor for `desiutil.log` at a later date, but let's not have it hold up this PR for now (it's already quite a lot for one PR).

I personally would _definitely_ retain the `nobs` columns. People in the working groups have been cutting on `nobs` to retain a "more representative" set of the final imaging to test target densities. So, I think that they're useful columns.

I can't see why we'd ever need the `ivar` columns for shape. I can see potential uses for `fracdev`. Let's leave those columns in for now but with a view to dropping them in a year if they aren't in use downstream.

I'll open issues about trimming columns from the target files, and about using `desiutil.log` instead of `print`, but then let's merge so I can move onto other things.

moustakas:
Agreed on all fronts -- merging now.

PR #202, opened 2017-07-17
 add RF for ELG slection in Sandbox with a new training

Cyeche:
- Add a new RF trained with HSC data for the selection of ELG in the sandbox
- A possible issue with the size of the RF file stored in sandbox/data


sbailey:
The new file is `py/desitarget/sandbox/data/rf_model_dr3_elg_HSC.npz` at 65MB.  Christophe says this was big enough to generate a warning when he pushed it, but github did allow it.  They do have a [strict upper limit of 100MB](https://help.github.com/articles/what-is-my-disk-quota/) and a recommended max repo size including all history of 1GB.  This case might be ok, but we really should try to catch this before it enters the repo instead of after, and question whether the extra trees / extra training data that made this larger were really necessary for the performance of the algorithm or just semi-random choices (or worse, overtraining).

Sorry to be out of the loop, but does this commit come with documentation about its performance, training sample, cross checks, etc?  Could you link to that (docdb, collab meeting or targeting telecon presentation...)?  I know this is the sandbox, but big files in the repo can become problematic for everyone including out of the sandbox, so we should check this before merging into master (or otherwise cleaning up with [BFG repo cleaner](https://rtyley.github.io/bfg-repo-cleaner/) before the history gets complicated).

Cyeche:
Hi,

The point is that we need a well defined procedure to store this kind of files.  This issue happened with the sandbox, but we can imagine that for the official code, we retrain the Random Forest RF) algorithm for QSO selection and that we will have several versions of RF. I think that the mechanism to download these files is similar to calibrations files for spectra. If a method exists for spectra calibrations we should put in place for the RF too.

Christophe.

geordie666:
We could have a mechanism similar to `desimodel`, where we have a separate directory in SVN specifically for large-ish files. That requires an environment variable to be set but isn't _too_ awkward?

sbailey:
If we can keep the RF definition files in github, that is way more convenient than having to separately download.  And even though this file is big, if it *needs* to be that big then ok.  My only pushback is about whether it needs to be this way, or whether half as many trees or double the leaf size would have worked just as well.  Has that been studied?  It's a pain that file size has to be a consideration, but let's not switch too quickly to an externally tracked set of data files until we've verified that we really do need a RF definition file this large.

sbailey:
If we can keep the RF definition files in github, that is way more convenient than having to separately download.  And even though this file is big, if it *needs* to be that big then ok.  My only pushback is about whether it needs to be this way, or whether half as many trees or double the leaf size would have worked just as well.  Has that been studied?  It's a pain that file size has to be a consideration, but let's not switch too quickly to an externally tracked set of data files until we've verified that we really do need a RF definition file this large.

sbailey:
If we can keep the RF definition files in github, that is way more convenient than having to separately download.  And even though this file is big, if it *needs* to be that big then ok.  My only pushback is about whether it needs to be this way, or whether half as many trees or double the leaf size would have worked just as well.  Has that been studied?  It's a pain that file size has to be a consideration, but let's not switch too quickly to an externally tracked set of data files until we've verified that we really do need a RF definition file this large.

moustakas:
Another option would be to bump this to a new `desidevtarget` (or whatever) repository in the `cosmodesi` organization as the next level of a sandbox repo.  I'm not saying this is the best solution but I wanted to mention it.

sbailey:
If we can keep the RF definition files in github, that is way more convenient than having to separately download.  And even though this file is big, if it *needs* to be that big then ok.  My only pushback is about whether it needs to be this way, or whether half as many trees or double the leaf size would have worked just as well.  Has that been studied?  It's a pain that file size has to be a consideration, but let's not switch too quickly to an externally tracked set of data files until we've verified that we really do need a RF definition file this large.

sbailey:
Sorry for the repeats.  github kept giving me a 500 error but apparently the messages actually were getting through; I wasn't trying to overemphasize my point...

geordie666:
I agree that it would be convenient to not have to have large external files. So, let's ask @cyeche to investigate whether it's possible to reduce the file size without loss of purity/efficiency (e.g., force a limit on the number of trees) and we'll merge when that investigation is complete.

Cyeche:
Hello,

Event If I succeed in reducing the size of the file by a factor 2, it will not really address this issue. For sure we will have new versions of the RF in the coming years and  I think that we need a mechanism that synchronize the code and the RF stored somewhere. It is not urgent to design it but I don't think that it will be possible to continue without.

Cheers, Christophe.

> I agree that it would be convenient to not have to have large external files. So, let's ask @Cyeche to investigate whether it's possible to reduce the file size without loss of purity/efficiency (e.g., force a limit on the number of trees) and we'll merge when that investigation is complete.
> 
>

sbailey:
Sounds good.  Note that if we do determine that a smaller file will work, we'll still need to BFG repo clean this file out of the way before replacing it since it has already been pushed to github.

geordie666:
@cyeche: If you can reduce the file size by a factor of say ~5, then I think all of our problems would be solved. Git could support 50 RF files of size ~10MB without complaining too much. And, we could delete obsolete versions from the sandbox. Would a factor of ~5 be plausible?

Cyeche:
A factor 5 is difficult, a factor 2 doable… Actually the size is two times bigger than what I was expecting naively, This means that I have to understand this point anyway….

> @Cyeche: If you can reduce the file size by a factor of say ~5, then I think all of our problems would be solved. Git could support 50 RF files of size ~10MB without complaining too much. And, we could delete obsolete versions from the sandbox. Would a factor of ~5 be plausible?

geordie666:
Great! Let's try and get the file size as small as we can and then we'll work out how to proceed. At that point myself or @sbailey can use the BFG on the larger file in the repo. If we end up with 10 RF files at 30MB over the next few years, then that might be OK.

If it's possible to use a smaller RF until DR6+, then I'd still prefer that..._if_ it only impacts the targets at the level of a few %. We could then use a larger tree on the final (and the SV) data to retrieve that few %.

geordie666:
In the end, we decided that the easiest way to keep large files out of github was to simple have @Cyeche not commit them. If people want to experiment with the RF method for ELGs, they'll have to obtain the files directly from @Cyeche. If enough people need these files, we'll investigate an SVN option for the data.

PR #203, opened 2017-07-20
adds mpi_merge_mock_targets; EXTNAME cleanup in select_mock_targets

sbailey:
Adds `mpi_merge_mock_targets` to do an MPI-parallelized merging of the individual per-healpix targets, truth, sky, and standards files into global targets.fits, truth.fits, etc.  This was formerly in two_percent_DESI with details hardcoded for the 2% sprint.

At the individual healpixel level, `select_mock_targets` (`desitarget.mock.build.targets_truth`) now puts the truth metadata table and the truth spectra into a single file per pixel instead of separate files.  I found that if the TRUTH table is in the file before the spectra, this is still efficient for reading and merging the TRUTH table, but it isn't consistently efficient if the TRUTH table comes after WAVE and FLUX.  Even though the fits libraries should be able to fseek over the data without actually reading them, it appears that there is still some I/O overhead for large fseeks if the file isn't fresh on disk.

This is part of the 2% redo cleanup.


bin/mpi_merge_mock_targets
```diff
@@ -0,0 +1,124 @@
...
+    todo['targets'] = not os.path.exists(args.outdir+'/targets.fits') or args.force
+    todo['truth'] = not os.path.exists(args.outdir+'/truth.fits') or args.force
+    todo['mtl'] = not os.path.exists(args.outdir+'/mtl.fits') or args.force
+else:
+    todo = None
+
+todo = comm.bcast(todo, root=0)    
+
+if todo['sky']:
+    merge_files(comm, args.mockdir+'/*/*/sky-*.fits', 'SKY', args.outdir+'/sky.fits')
```
moustakas: It's unlike this code will be run anywhere other than at NERSC but usually I try to use `os.path.join` and `.format()`.  I'm just being unhelpfully picky, though, so no changes needed.

py/desitarget/mock/build.py
```diff
@@ -855,47 +855,40 @@ def _update_spectra_status(result):
                 log.info('No {} standards on healpix {}, {} not written.'.format(stdsuffix, pixnum, stdfile))
 
         # Finally write out the rest of the targets.
-        # targetsfile = os.path.join(output_dir, 'targets-{}.fits'.format(healsuffix))
-        # truthfile = os.path.join(output_dir, 'truth-{}.fits'.format(healsuffix))
-        # truthspecfile = os.path.join(output_dir, 'spectra-truth-{}.fits'.format(healsuffix))
-
         targetsfile = mockio.findfile('targets', nside, pixnum, basedir=output_dir)
         truthfile = mockio.findfile('truth', nside, pixnum, basedir=output_dir)
```
moustakas: Doesn't `desispec.io.util.findfile` have the functionality to find these files, given a different top-level path?

py/desitarget/mock/build.py
```diff
@@ -855,47 +855,40 @@ def _update_spectra_status(result):
...
             log.info('Writing {}'.format(truthfile))
-            try:
-                truth[inpixel].write(truthfile+'.tmp', format='fits', overwrite=True)
-            except:
-                truth[inpixel].write(truthfile+'.tmp', format='fits', clobber=True)
-            os.rename(truthfile+'.tmp', truthfile)
-        
-            log.info('Writing {}'.format(truthspecfile))
             hx = fits.HDUList()
+            hdu = fits.convenience.table_to_hdu(truth[inpixel])
```
moustakas: Didn't know about this method -- nice.

py/desitarget/mock/build.py
```diff
@@ -855,47 +855,40 @@ def _update_spectra_status(result):
...
-                hx.writeto(truthspecfile+'.tmp', overwrite=True)
+                hx.writeto(truthfile+'.tmp', overwrite=True)
             except:
-                hx.writeto(truthspecfile+'.tmp', clobber=True)
-            os.rename(truthspecfile+'.tmp', truthspecfile)
+                hx.writeto(truthfile+'.tmp', clobber=True)
+            os.rename(truthfile+'.tmp', truthfile)
 
 def join_targets_truth(output_dir='.', nside=64, healpixels=None, verbose=True):
     """Combine all the target and truth catalogs generated by targets_truth into a
```
moustakas: This function (`mock.build.join_targets_truth`) is a serialized version of `mpi_merge_mock_targets`, although it's now obsolete with the new data model.  We should nevertheless update it as we'll still want to be able to produce small test datasets (e.g., on a laptop).  I leave it up to you to decide whether to do this in this PR or another one.

py/desitarget/mock/build.py
sbailey: Since these files are specific to the mock target building, `desispec.io.findfile` doesn't know about these (it only knows about real data spectro pipeline stuff).  thus the `desitarget.mock.io.findfile()` for things that are specific to mock I/O files.

bin/mpi_merge_mock_targets
sbailey: You are formally correct; sometimes I do that and sometimes I favor what seems more readable while still working on Linux and Mac.

py/desitarget/mock/build.py
sbailey: Thanks for the quick review and for catching the replication.  I refactored the script functionality into this module and structured it so that MPI is now optional.  

Normal serial:
```
join_mock_targets [options]
```
With 4 MPI ranks reading in parallel:
```
srun -n 4 join_mock_targets [options] --mpi
```

PR #206, opened 2017-08-03
Mocks without spectra

forero:
This PR implements the option `--no-spectra` into `mpi_build_mock_targets` to avoid generating spectra on the fly. 

The code works but I think I found some problems. 
It seems to me that helpixels trace can be seen in the catalogs. See [this notebook](https://github.com/desihub/quicksurvey_example/blob/without_spectra/nospectra.ipynb). The first two plots there come from the `--no-spectra` option, the next two plots come from the version generating the full spectra. In both cases there are some overdensities piling up on the healpix boundaries. 
This is more evident with `n_side=32` or `n_side=16`. I will post those plots later. In the meantime I will start looking into what is causing the problem.

You can use [this `quicksurvey_example` branch](https://github.com/desihub/quicksurvey_example/tree/without_spectra) to test the `--no-spectra` option. All the required input files are there.

### DIFF
--- a/bin/mpi_select_mock_targets
+++ b/bin/mpi_select_mock_targets
@@ -18,7 +18,7 @@ import argparse
 from astropy.table import Table
 import desimodel.footprint
 from desiutil.log import get_logger, DEBUG
-from desitarget.mock.build import targets_truth
+from desitarget.mock.build import targets_truth, targets_truth_no_spectra
 import desitarget.mock.io as mockio
 
 import multiprocessing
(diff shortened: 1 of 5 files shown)


moustakas:
Thanks for this update, @forero.  I'll try to review the code soon but note that this is going to create some ~significant conflicts with my branch, which is updating the mock code to deal with the DR4 data model (which @sbailey is waiting on).  I would definitely like to hold off before merging this until we figure out the best path forward.

forero:
Good. I will wait. What is the timescale for you to merge your branch with master?

sbailey:
#207 for new data model has been merged, but that generated conflicts here.  @forero can you fix the conflicts here and then post when it is ready for review again?

forero:
@moustakas Sure. I will also wait for @geordie666's BFG update today to resubmit this PR.

geordie666:
@forero I'm sure this is obvious, but make sure you've pushed all of your current work to desitarget. Once I update the repo, you'll have to clone a clean copy, and anything that hasn't been pushed will then be gone.

moustakas:
I have just one quick comment--

I personally think that the right way to approach this---originally suggested by @sbailey---is to generate GMMs based on a complete spectral simulation of a large chunk of the survey (whether 2%, 10%, or something different), and then to sample from those GMMs to (quickly) generate target and truth catalogs (without spectra) for the whole survey footprint (many times, if necessary).

That way all the correlations and distributions between, e.g., target class, [OII] flux, magnitudes, colors, etc. will be properly captured.

@forero also please keep in mind (oops, I guess that's two comments) that my next refactor of `mock.build.targets_truth` will significantly change the driver function `get_spectra_onebrick`.  I note that in this PR the function `get_magnitudes_onebrick` contains a lot of cut-and-paste code from that code which will need to be rewritten as well.

Issue #192, opened 2017-06-20
ensure that the QSO target densities aren't too low


moustakas:
Opening an issue based on an off-list conversation with various QSO-aware team members.

The basic issue is that the QSOs target densities in the data challenge code are too low, especially for Lya QSOs.

Here's what currently happens in the code:

1. Tracer QSOs are read from the Gaussian random field mock and assigned a g-band magnitudes (but not colors) from the Gaussian Mixture Models derived by @belaa from the DR2 imaging.  The Lya QSOs are read from Jean-Marc's mocks and come with g-band magnitudes (although Jean-Marc's new mocks don't).  

2. I then use the `desisim.templates.QSO` class to generate spectra at the input redshift and normalized to the desired apparent magnitude.  One of the outputs of this code is the synthesized DECaLS grz fluxes (but not WISE fluxes since the template spectra don't go red enough).  

3. Next, I add scatter to the (noiseless) photometry based on the brick that the QSO falls on and then apply the *optical* color cuts to select bona fide QSO targets.  This is the point where we're loosing too many QSOs since targets (and thereby spectra) that don't pass these cuts are simply discarded.

4. Finally the resulting target densities are compared to the desired target densities specified in the parameter file (120/deg2 and 50/deg2 for tracers and Lya, resp.) plus or minus the target density fluctuations model that @geordie666 has coded up.  Targets are randomly subsampled if the "measured" density is above the desired density.  If the "measured" density is too low then a warning is issued and the code keeps going (retaining all the targets, obviously).

It's obvious now that this sequence is both inefficient and insufficient because it leads to too few targets (especially Lya QSOs).

So what I'm going to do is to make steps 2-4 iterative and aware of one another until the desired target density is either achieved or exceeded.

moustakas:
BTW here are the QSO target densities from the input mocks.

<img width="817" alt="screen shot 2017-06-19 at 3 05 28 pm" src="https://user-images.githubusercontent.com/1431820/27350196-8a664a6e-55ae-11e7-9cf6-514de21de3bc.png">

moustakas:
Going to track this issue in #252.

PR #207, opened 2017-08-04
update mock.build.targets_truth to use the latest imaging data model

moustakas:
This PR updates `mock.build.targets_truth` (and related functions) to use the latest imaging (>=DR4) data model.  I've tested this on my laptop with a single nside=64 healpixel (26370) but not at NERSC.  

@sbailey:  If you could generate ~10 deg2 or so of simulated spectra then we should check the QA output (which I've documented elsewhere but will do so again) before proceeding, to make sure there aren't any issues I missed.

@geordie666:  It's quite likely I broke the fluctuations code in `desitarget.QA` (maybe just backwards-compatibility), but that code is due for a refactor anyway and doesn't need to be looked at now.

@forero: Once this is merged your #206 can proceed, although there are going to be a number of annoying conflicts which will need to be resolved.  Sorry!

Also note that this PR depends on https://github.com/desihub/desisim/pull/252 being merged first (and subsequently `desitarget` will depend on `desisim/master`).

Finally note that this does not yet include the promised speed-up of the code.


geordie666:
Don't worry about breaking `desitarget.QA` I didn't update it for the new data model, anticipating that at some point I'd get around to rewriting it for HEALPix.

sbailey:
desihub/desisim#252 has a units issue to resolve before it can be merged and thus before this can be merged.

In the meantime, here is the list of columns in select_mock_targets -> targets*.fits output that are *not* in `/project/projectdirs//desi/target/catalogs/targets-dr4-0.14.0.fits`:
```
#- should be MW_TRANSMISSION* instead?
EBV

#- shouldn't be in here?  targetmask.yaml has mapping of what bits
#- can be observed under what conditions, but I don't think we should
#- fix that choice into the target catalog at the time it is written.
OBSCONDITIONS

#- should be in real targets file, i.e. ok to be here?
SHAPEDEV_E1
SHAPEDEV_E2
SHAPEEXP_E1
SHAPEEXP_E2
SUBPRIORITY
```
and here are the columns in targets-dr4-0.14.0.fits that are not in select_mock_targets output:
```
BRICKID
BRICK_OBJID
DCHISQ
DEC_IVAR
FLUX_IVAR_G
FLUX_IVAR_R
FLUX_IVAR_W1
FLUX_IVAR_W2
FLUX_IVAR_W3
FLUX_IVAR_W4
FLUX_IVAR_Z
FLUX_W3
FLUX_W4
FRACFLUX_G
FRACFLUX_R
FRACFLUX_Z
HPXPIXEL
MW_TRANSMISSION_G
MW_TRANSMISSION_R
MW_TRANSMISSION_W1
MW_TRANSMISSION_W2
MW_TRANSMISSION_W3
MW_TRANSMISSION_W4
MW_TRANSMISSION_Z
NOBS_G
NOBS_R
NOBS_Z
PHOTSYS
RA_IVAR
RELEASE
TYPE
```

Please sync up what you can, and open a ticket documenting other items to address later.

moustakas:
[Note: Blocked out text is a quote from the preceding comment from @sbailey and my responses are below each block.]
```
#- should be MW_TRANSMISSION* instead?
EBV
MW_TRANSMISSION_G
MW_TRANSMISSION_R
MW_TRANSMISSION_W1
MW_TRANSMISSION_W2
MW_TRANSMISSION_W3
MW_TRANSMISSION_W4
MW_TRANSMISSION_Z
```
This is unsatisfactorily messy right now.  Basically, none of the photometric fluxes *nor* the spectra include the effects of Galactic reddening.  I can include this but I need some input: (1) do we have a MW reddening curve code up anywhere, e.g., in desiutil?  (2) does the pipeline use the *EBV* in the targets catalog (or in whatever downstream catalog is used as input to the spectral processing, e.g., fibermap?) to deredden the spectra.  If not we'll want to be a little careful, right?; and (3) should I include reddened fluxes in the *targets* catalog and dereddened fluxes in the *truth* catalog, or reddened fluxes in both (in which case MW_TRANSMISSION_? will need to be in the *truth* catalog as well).

```
#- shouldn't be in here?  targetmask.yaml has mapping of what bits
#- can be observed under what conditions, but I don't think we should
#- fix that choice into the target catalog at the time it is written.
OBSCONDITIONS
SUBPRIORITY
```
I included these because @forero mentioned that they were needed for fiberassignment. Please confirm before I remove them.

```
#- should be in real targets file, i.e. ok to be here?
SHAPEDEV_E1
SHAPEDEV_E2
SHAPEEXP_E1
SHAPEEXP_E2
```
I can't find the thread right now but @dkirkby mentioned that we will want these in our "real" target catalog at some point in the future, so keeping them here I think makes sense.

```
and here are the columns in targets-dr4-0.14.0.fits that are not in select_mock_targets 
output. Please sync up what you can, and open a ticket documenting other items to 
address later.
```

```
FLUX_IVAR_G
FLUX_IVAR_R
FLUX_IVAR_W1
FLUX_IVAR_W2
FLUX_IVAR_W3
FLUX_IVAR_W4
FLUX_IVAR_Z
FLUX_W3
FLUX_W4
```
They'll be simplistic, but adding the inverse variances shouldn't be a problem.  I actually hadn't realized that we were doing *W3,W4* forced photometry.  Unfortunately, the issues are that (1) some of the templates don't go red enough to synthesize photometry at these wavelengths and (2) the galaxy templates don't include dust emission, so the fluxes wouldn't be useful even if they did. Should I include these as dummy (empty) columns or not include them at all?  

And note that we don't use W3/W4 for target selection because that imaging is very shallow.

```
BRICKID
BRICK_OBJID
DCHISQ
DEC_IVAR
FRACFLUX_G
FRACFLUX_R
FRACFLUX_Z
HPXPIXEL
NOBS_G
NOBS_R
NOBS_Z
PHOTSYS
RA_IVAR
RELEASE
TYPE
```
I could use some input from @geordie666 about the rest of these.  I think I can include `BRICKID` easily enough.  `RA_IVAR` and `DEC_IVAR` requires an astrometric error model and many of the rest of these are related to the actual survey data that *could* be mocked up in the output catalog, but I'm not sure how useful they would be...  Thoughts?

forero:
@moustakas You can remove `OBSCONDITIONS` and `SUBPRIORITY` provided that we modify `mtl` to compute those two columns before feeding the info into `fiberassign`. I think we should also have a random number column attached to the `targets` file which can be used used later as `SUBPRIORITY`.

moustakas:
The random seed is stored in the truth catalog (which makes sense). Is this
sufficient?

forero:
It makes sense for the simulations, but in real life we will only have the `targets` file and we will need the random number there as well. My vote goes to having the random number number in the `targets` file.

geordie666:
Some comments on these (none of which, I would say, really **need** to be passed through):

`BRICKID` - very easy to include now that we have correct code in `desiutil.brick`. So, probably just included it.
`BRICK_OBJID - probably not needed as long as we're passing around a rigorous `TARGETID` value.  In particular, `TARGETID` should use the `encode_targetid` function in `desitarget.targets` https://github.com/desihub/desitarget/blob/master/py/desitarget/targets.py#L124
`DCHISQ` - might ultimately be useful as a spectroscopic prior on galaxy/star? 
`DEC_IVAR` - might be useful for fiber assignation?
`FRACFLUX_G` - probably not needed
`FRACFLUX_R` - probably not needed
`FRACFLUX_Z` - probably not needed
`HPXPIXEL` - easy enough to generate, so we may as well keep it. The `NSIDE` is in the header of the targets file.
`NOBS_G` - probably not needed (although useful for debugging why a target might have been "bad")
`NOBS_R` - probably not needed (although useful for debugging why a target might have been "bad")
`NOBS_Z` - probably not needed (although useful for debugging why a target might have been "bad")
`PHOTSYS` - probably not needed
`RA_IVAR` - might be useful for fiber assignation?
`RELEASE` - probably not needed as long as we're passing around a rigorous `TARGETID`. In particular, `TARGETID` should use the `encode_targetid` function in `desitarget.targets` https://github.com/desihub/desitarget/blob/master/py/desitarget/targets.py#L124
`TYPE` - might ultimately be useful as a spectroscopic prior on galaxy/star?

sbailey:
Punting on the reddening question for the moment (need to read more about what is downstream; I don't recall...)

SUBPRIORITY: this accidentally moved categories during the replies.  Looking back at my original comment, it was in the set of "should be in real targets file, i.e. ok to be here?" (which should have been a statement, not a question)  i.e. SUBPRIORITY should be in the real target file too and it is an oversight that it isn't (and shows that we haven't run fiberassign on a real target catalog in a long time).  i.e. please leave it here (and not just a random seed).

OBSCONDITIONS, on the other hand, I think should be calculated by the MTL code based upon the targeting bits and targetmask.yaml info, and not put into the target catalog from the very start.

Other missing columns: please add the simple ones or spin to a separate ticket, but let's *not* add a blank or meaningless column just to match the data model.  I'd rather crash later due to a missing column than silently do the wrong thing later because it is using meaningless numbers in a column.

geordie666:
@moustakas: To be slightly clearer, you EITHER need to:

1. Set `BRICK_OBJID` to a unique value and use `targets.encode_targetid` OR
2. Set `TARGETID` to a "dummy" unique value, and ignore `BRICK_OBJID`

My preference is 1. but my understanding is that you're currently doing 2.

Oh, and, yes, please, set `mock=1`! That's exactly what that's for...

I'm not 100% sure what we'd do with `RELEASE`. You could start using it to encode mock releases according to some schema. Whatever you do won't clash with the actual data releases provided you send `mock`=1.

moustakas:
I've implemented Option 1, @geordie666, which is the right thing to do, I think (with `mock=1`). I'm leaving `RELEASE` off for now but documenting this in a separate issue (to be posted shortly).

moustakas:
OK, this I think is ready to merge.  Note that this PR only includes data model changes and not the promised speed-up / memory refactor (which will also address a handful of other outstanding issues like #192).

moustakas:
Tests are failing with 
```
[snip]
    raise BadZipfile("Bad CRC-32 for file %r" % self.name)
BadZipfile: Bad CRC-32 for file 'arr_0.npy'
```
on code that this PR didn't touch.  Not sure how to fix this --

sbailey:
I'm trying a rerun of the travis test in case it was "just" a corrupted file when it did the git clone.  We haven't BFG removed the big random forest file in PR #202, so the git clone is pretty heavyweight.

For the record, the remaining mock vs. real targets file discrepancies are:

In mocks but not in real targets, but should be in real targets:
```
SHAPEDEV_E1
SHAPEDEV_E2
SHAPEEXP_E1
SHAPEEXP_E2
SUBPRIORITY
```

In real targets but not yet in mocks; being tracked in #209:
```
DCHISQ
FLUX_IVAR_*
FLUX_W3, FLUX_W4
MW_TRANSMISSION_W3, MW_TRANSMISSION_W4
FRACFLUX_G, FRACFLUX_R, FRACFLUX_Z
NOBS_G, NOBS_R, NOBS_Z
PHOTSYS
RA_IVAR, DEC_IVAR
RELEASE
TYPE
```

I'll merge after we sort out the apparently unrelated test failures.

geordie666:
After digging around, I was just going to suggest (and hope) that it's file corruption in the clone.

I wanted to hold off BFG'ing until we clean up a couple of these PRs and merge the `ADMplotQA` branch (which is close to ready). I'll have to warn everyone to step back for a few hours and then to re-clone desitarget, to avoid the large file being re-committed by git history.

sbailey:
Rerunning on Travis without changing anything didn't have the bad CRC-32 error, nor did a fresh pull from my laptop.  Merging now.

Issue #204, opened 2017-07-26
STD_WD definition in dark and bright time not consistent between targetmask and mock.build


apcooper:
Carried over from #169 

White dwarfs are being written out as (bright- and dark-time) standards:
https://github.com/desihub/desitarget/blob/master/py/desitarget/mock/build.py#L845-L847
but this looks inconsistent with https://github.com/desihub/desitarget/blob/master/py/desitarget/data/targetmask.yaml#L24, where the obsconditions for STD_WD are (DARK|GRAY).

apcooper:
This is easy to fix in `targetmask.yaml` (7a69904). The extra bit of logic on `STD_WD` [here](https://github.com/desihub/desitarget/blob/master/py/desitarget/mock/build.py#L845-L847) is probably unnecessary now?

apcooper:
Fixed in #208.

PR #208, opened 2017-08-09
No explicit test on STD_WD when writing standards

apcooper:
Treats WDs in the same way as other standards, i.e.relying on them having the observing conditions set correctly in targetmask.  Re. #169, #204.

### DIFF
--- a/py/desitarget/mock/build.py
+++ b/py/desitarget/mock/build.py
@@ -838,13 +838,11 @@ def targets_truth(params, output_dir='.', realtargets=None, seed=None, verbose=F
                 os.rename(skyfile+'.tmp', skyfile)
 
         # Write out the dark- and bright-time standard stars.
-        for stdsuffix, stdbit in zip(('dark', 'bright'), ('STD_FSTAR', 'STD_BRIGHT')):
+        for stdsuffix, stdbit in zip(('dark', 'bright'), ('STD_FSTAR', 'STD_BRIGHT', 'STD_WD')):
             # stdfile = os.path.join(output_dir, 'standards-{}-{}.fits'.format(stdsuffix, healsuffix))
             stdfile = mockio.findfile('standards-{}'.format(stdsuffix), nside, pixnum, basedir=output_dir)
 
-            istd = (pixnum == targpix) * ( (
-                (targets['DESI_TARGET'] & desi_mask.mask(stdbit)) |
-                (targets['DESI_TARGET'] & desi_mask.mask('STD_WD')) ) != 0)
+            istd = (pixnum == targpix) * ( ( (targets['DESI_TARGET'] & desi_mask.mask(stdbit)) ) != 0)
             #istd = (targets['DESI_TARGET'] & desi_mask.mask(stdbit)) != 0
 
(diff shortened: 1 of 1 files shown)


apcooper:
I think this is simple enough, I just don't have time now to test it so a second look would help. Is the `*` in that line rather than `&` for a particular reason?

py/desitarget/mock/build.py
```diff
@@ -838,13 +838,11 @@ def _update_spectra_status(result):
...
         # Write out the dark- and bright-time standard stars.
-        for stdsuffix, stdbit in zip(('dark', 'bright'), ('STD_FSTAR', 'STD_BRIGHT')):
+        for stdsuffix, stdbit in zip(('dark', 'bright'), ('STD_FSTAR', 'STD_BRIGHT', 'STD_WD')):
             # stdfile = os.path.join(output_dir, 'standards-{}-{}.fits'.format(stdsuffix, healsuffix))
             stdfile = mockio.findfile('standards-{}'.format(stdsuffix), nside, pixnum, basedir=output_dir)
 
-            istd = (pixnum == targpix) * ( (
-                (targets['DESI_TARGET'] & desi_mask.mask(stdbit)) |
-                (targets['DESI_TARGET'] & desi_mask.mask('STD_WD')) ) != 0)
+            istd = (pixnum == targpix) * ( ( (targets['DESI_TARGET'] & desi_mask.mask(stdbit)) ) != 0)
```
apcooper: Is there a reason for `*` rather than `&` here?

py/desitarget/mock/build.py
```diff
@@ -838,13 +838,11 @@ def _update_spectra_status(result):
                 os.rename(skyfile+'.tmp', skyfile)
 
         # Write out the dark- and bright-time standard stars.
-        for stdsuffix, stdbit in zip(('dark', 'bright'), ('STD_FSTAR', 'STD_BRIGHT')):
+        for stdsuffix, stdbit in zip(('dark', 'bright'), ('STD_FSTAR', 'STD_BRIGHT', 'STD_WD')):
```
sbailey: This zip has two elements in the first list and 3 in the second list, and thus will drop the 3rd element as currently written.  This logic may need a slightly larger refactor to have more than one target bit mapping to the "dark" or "bright" standards.

py/desitarget/mock/build.py
sbailey: Either `*` or `&` will work for logically joining two boolean arrays.
```
In [3]: np.array([True, False]) & np.array([True, True])
Out[3]: array([ True, False], dtype=bool)

In [4]: np.array([True, False]) * np.array([True, True])
Out[4]: array([ True, False], dtype=bool)
```

moustakas:
Can we wait until #207 is in before merging this.  I'm nearly done with my PR!

apcooper:
Thanks, I'll fix the comments and update the PR.

apcooper:
OK to merge this now?

sbailey:
Thanks for the ping; forgot about this one.  Looks good.  Merging now.

PR #210, opened 2017-08-09
Software for producing (initial) QA plots/webpage

geordie666:
This PR produces the first set of QA webpages of target files. It:

- Has a simple binary `run_target_qa` that runs the plot-making software in full
- Creates 1-D and 2-D density plots that can be weighted by the DESI footprint
- Makes color-color plots of targets
- Produces a .html page that wraps the plots
   (see, e.g., http://portal.nersc.gov/project/desi/users/adamyers/desitargetQA/ )

@sbailey: I factored this PR off a rebase of master to simplify history, as lots of PRs were ongoing; please check the overall sequence of commit logs

@moustakas: This PR contains an example of how to deal with the second part of https://github.com/desihub/desiutil/issues/68 . For instance, the creation of pixel weights based on footprint coverage in `desimodel` is here:

https://github.com/desihub/desitarget/blob/ADMplotQArebase/py/desitarget/QA.py#L1330-L1341 

and those weights can be passed to `desutil.plots.plot_sky_binned` as here:

https://github.com/desihub/desitarget/blob/ADMplotQArebase/py/desitarget/QA.py#L1089-L1091


sbailey:
[Network graph](https://github.com/desihub/desitarget/network) looks fine so rebase + branch + push looks like to worked out.

I was surprised at how long this took (13m40s) but otherwise fine.  I'll save minor quibbles for future requests.  Merging now.

PR #211, opened 2017-08-12
Mocks without spectra in the new target model.

forero:
(After PR #207 it was easier to create a new branch an a new PR than fix #206)
This PR implements the option --no-spectra into mpi_build_mock_targets to avoid generating spectra on the fly.

The code works but I think I found one problem and one bug

1. It seems to me that helpixels trace can be seen in the catalogs. [See this notebook](https://github.com/desihub/quicksurvey_example/blob/without_spectra/nospectra.ipynb). The first two plots there come from the --no-spectra option, the next two plots come from the version generating the full spectra. In both cases there are some overdensities piling up on the healpix boundaries.
This is more evident with n_side=32 or n_side=16. I will post those plots later. In the meantime I will start looking into what is causing the problem.

2. `mpi_select_mock_targets` produces non-unique `TARGETIDs` for `targets` and `sky`. This bug will make `fiberassign` crash. This bug was already in master. I tried to fix it [with this](https://github.com/desihub/desitarget/blob/2dfd9bdde8f3aabaa72d99cd312d31f88bbe851a/py/desitarget/mock/build.py#L1396-L1399), but it doesn't cover all cases. It seems that some bricks are covered by two different processors.

You can use this [quicksurvey_example branch](https://github.com/desihub/quicksurvey_example/tree/without_spectra) to test the --no-spectra option. All the required input files are there.

### DIFF
--- a/py/desitarget/mock/build.py
+++ b/py/desitarget/mock/build.py
@@ -1042,3 +1042,151 @@ def join_targets_truth(mockdir, outdir=None, force=False, comm=None):
         mtl.meta['EXTNAME'] = 'MTL'
         mtl.write(tmpout, overwrite=True, format='fits')
         os.rename(tmpout, out_mtl)
+
+def _get_magnitudes_onebrick(specargs):
+    """Filler function for the multiprocessing."""
+    return get_spectra_onebrick(*specargs)
+
+def get_magnitudes_onebrick(target_name, mockformat, thisbrick, brick_info, Spectra, dust_dir,
+                         select_targets_function, source_data, rand, log):
+    """Wrapper function to generate magnitudes for all the objects on a single brick."""
+        
+def targets_truth_no_spectra(params, output_dir='.', realtargets=None, seed=None, verbose=False,
+                  bricksize=0.25, nproc=1, nside=16, healpixels=None):
+    """Generate a catalog of targets, spectra, and the corresponding "truth" catalog
+    (with, e.g., the true redshift) for use in simulations.
+
+    Args:
+        params : dict
+            Source parameters.
+        output_dir : str
+            Output directory (default '.').
+        realtargets : astropy.table
+            Real target catalog table, e.g. from DR3 (deprecated!).
+        bricksize : float
+            Brick size for assigning bricknames, which should match the imaging
+            team value (default 0.25 deg).
+        nproc : int
+            Number of parallel processes to use (default 1).
+        nside : int
+            Healpix resolution corresponding to healpixels (default 16).
+        healpixels : numpy.ndarray or int
+            Restrict the sample of mock targets analyzed to those lying inside
+            this set (array) of healpix pixels.
+
+    Returns:
+        A variety of fits files are written to output_dir.
+
+    """
+    import healpy as hp
+    from time import time
+
+    from astropy.io import fits
+
+    from desiutil import depend
+    from desispec.io.util import fitsheader, write_bintable
+    from desitarget.mock.selection import SelectTargets
+    from desitarget.mock.spectra import MockMagnitudes
+    from desitarget.internal import sharedmem
+    from desimodel.footprint import radec2pix
+    
+    if verbose:
+        log = get_logger(DEBUG)
+    else:
+        log = get_logger()
+
+    if params is None:
+        log.fatal('Required params input not given!')
+        raise ValueError
+
+    # Initialize the random seed
+    rand = np.random.RandomState(seed)
+    # Create the output directories and clean them up if necessary.
+    if os.path.exists(output_dir):
+        if os.listdir(output_dir):
+            log.warning('Output directory {} is not empty.'.format(output_dir))
+    else:
+        log.info('Creating directory {}'.format(output_dir))
+        os.makedirs(output_dir)
+        
+    log.info('Writing to output directory {}'.format(output_dir))
+    print()
+
+    # Default set of healpixels is the whole sky (yikes!)
+    if healpixels is None:
+        healpixels = np.arange(hp.nside2npix(nside))
+
+    areaperpix = hp.nside2pixarea(nside, degrees=True)
+    skyarea = len(healpixels) * areaperpix
+    log.info('Grouping into {} healpixel(s) (nside = {}, {:.3f} deg2/pixel) spanning {:.3f} deg2.'.format(
+        len(healpixels), nside, areaperpix, skyarea))
+
+    brick_info = BrickInfo(random_state=rand, dust_dir=params['dust_dir'], bricksize=bricksize,
+                           decals_brick_info=params['decals_brick_info'], log=log,
+                           target_names=list(params['sources'].keys())).build_brickinfo()
+
+    # Initialize the Classes used to assign spectra and select targets.  Note:
+    # The default wavelength array gets initialized here, too.
+    log.info('Initializing the MockMagnitudes and SelectTargets Classes.')
+    Spectra = MockMagnitudes(rand=rand, verbose=verbose, nproc=nproc)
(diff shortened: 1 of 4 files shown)


forero:
This solves now the problem with the duplicate TARGETIDs. With the current parallelization (pixels in MPI, bricks in multiprocessing, looping over target types) the easiest way to ensure unique global TARGETID is to use a `pixel_objid` and a `pixel_id` to build the `TARGETID`.

moustakas:
@forero I think the issue is actually with `objid`, which is [set to a running integer](https://github.com/desihub/desitarget/blob/08fed3c2af46217cbd33b7db0ed9948d2f5598ed/py/desitarget/mock/build.py#L388) rather than with `brickid` (which is what you've changed), which by definition is unique.  

I *think* what's happening -- although you should confirm -- is that the way the code is set up right now the same brick can fall on two different, adjacent healpixels.  The objects on these bricks then get assigned the same `brickid` *and* the same `objid`, resulting in duplicate `TARGETID`. 

This behavior will be fixed when we refactor the fluctuations code (rather, when @geordie666 refactors it!) to work on the scale of healpixels rather than bricks.

That said, the solution you've implemented should be OK for now, but we should create a new issue to be sure we don't lose track of what we want to do in the future.

py/desitarget/mock/build.py
```diff
@@ -760,6 +761,12 @@ def _update_spectra_status(result):
                 if target_name.upper() != 'SKY':
                     trueflux = trueflux[keep, :]
 
+        #rewrite the target ID to be sure that it is unique in this brick
+        if len(targets):
+            new_pixel_ID = current_pixel_ID + len(targets)
+            targets['BRICK_OBJID'][:]= np.arange(current_pixel_ID,new_pixel_ID)
+            current_pixel_ID = new_pixel_ID
```
moustakas: This code is obsolete now, right?

py/desitarget/mock/build.py
```diff
@@ -818,15 +826,20 @@ def _update_spectra_status(result):
     except:
         nsky = 0
 
+    #compute target healpixel number
+    targpix = radec2pix(nside, targets['RA'], targets['DEC'])
     targetid = encode_targetid(objid=targets['BRICK_OBJID'],
-                               brickid=targets['BRICKID'], mock=1)
+                               brickid=targpix, mock=1)
```
moustakas: This is OK for now but in the future we will want to keep `BRICKID` to what it gets from the `brick_info` dictionary and ensure earlier in the code that `OBJID` is unique, thereby ensuring that `TARGETID` is always unique.  

We should also really refactor this piece of the code out and write a unit test to ensure `TARGETID` is never duplicated.

py/desitarget/mock/build.py
```diff
@@ -966,6 +978,12 @@ def merge_file_tables(fileglob, ext, outfile=None, comm=None):
         log.info('Writing {}'.format(outfile))
         header = fitsio.read_header(infiles[0], ext)
         tmpout = outfile + '.tmp'
+        
+        # Remove duplicates
+        #vals, idx_start, count = np.unique(data['TARGETID'], return_index=True, return_counts=True)
+        #if len(vals)!=len(data):
+        #    data = data[idx_start[count==1]]
+
```
moustakas: Maybe change this code block to an `assert` statement?

py/desitarget/mock/build.py
```diff
@@ -1042,3 +1060,509 @@ def join_targets_truth(mockdir, outdir=None, force=False, comm=None):
...
         os.rename(tmpout, out_mtl)
+
+def _get_magnitudes_onebrick(specargs):
+    """Filler function for the multiprocessing."""
+    return get_magnitudes_onebrick(*specargs)
+
+def get_magnitudes_onebrick(target_name, mockformat, thisbrick, brick_info, Magnitudes, dust_dir,
+                         select_targets_function, source_data, rand, log):
+    """Wrapper function to generate magnitudes for all the objects on a single brick."""
+
```
moustakas: This function (`get_magnitudes_onebrick`) contains a significant amount of cut-and-pasted code from the corresponding `get_spectra_onebrick` function (including the debugging plots, extinction coefficients, etc.).   Could you please refactor the common code so that the two functions just contain the code they need.  This will make it *significantly* easier for me to implement the speed-up I have in mind without having to change the same code in two places.

py/desitarget/mock/build.py
```diff
@@ -1042,3 +1060,509 @@ def join_targets_truth(mockdir, outdir=None, force=False, comm=None):
...
+    else:
+        log.debug('Selected {} {}s on brick {}.'.format(nobj, target_name.upper(), thisbrick))
+        targets = targets[keep]
+        truth = truth[keep]
+
+    return [targets, truth]
+
+def targets_truth_no_spectra(params, output_dir='.', realtargets=None, seed=None, verbose=False,
+                  bricksize=0.25, nproc=1, nside=16, healpixels=None):
+    """Generate a catalog of targets, spectra, and the corresponding "truth" catalog
```
moustakas: A significant fraction of this code is also cut-and-paste....

py/desitarget/mock/build.py
forero: I would prefer to keep the `magnitudes` vs. `spectra` functions completely separate. The `magnitudes` function doesn't actually need any speed-up.

py/desitarget/mock/build.py
forero: It's not. This is how I get the unique ID on a pixel.

py/desitarget/mock/build.py
forero: Will do.

py/desitarget/mock/build.py
forero: I would like to keep this as an independent function to use `--no_spectra` regardless of modifications in `targets_truth`.

forero:
@moustakas 
> I think the issue is actually with objid, which is set to a running integer rather than with brickid (which is what you've changed), which by definition is unique.

In the current implementation in `master` the problem is worse than that because `get_spectra_onebrick` is called multiple times **on the same brick** for different target classes, you get a huge number of repeating `objid`. The implementation on master gives ~35 % of repeated TARGETID.

> I think what's happening -- although you should confirm -- is that the way the code is set up right now the same brick can fall on two different, adjacent healpixels. The objects on these bricks then get assigned the same brickid and the same objid, resulting in duplicate TARGETID.

That's a different problem. We get that problem if we rewrite `objid` from `get_spectra_onebrick` to be unique across a brick and different target classes (i.e. the implementation in this branch)  and keep the `brickid` to build `TARGETID`. In that case we get ~0.25% of repeted TARGETID, all of them located at the intersection between pixels.

That's why I rewrite `objid` and used `pixid` to get unique `TARGETID`s once the pixels are merged.

py/desitarget/mock/build.py
moustakas: Speed-up is only one issue.  The bigger issue that my (future) refactor will address is that the target densities don't end up being correct.  For the QSOs the issue is described in #192 in detail.  I don't know whether your *no-spectra* version of `targets_truth` suffers from this problem.

On principle I also think we shouldn't be cutting and pasting dozens of lines of code.  

I should also mention that I don't expect any other major changes to the "generate spectra" version of `targets_truth` after the refactor which fixes the target densities.

forero: I agree we shouldn't be replicating tons of code. Once all the big current and future issues are solved in the full spectra version I will refactor the `--no-spectra` option into the latest version. 

Admittedly, this is a quick hack to avoid the previous state where we lost the option of running fast catalog only end-to-end simulations because of major ongoing work in `select_targets_truth`.

PR #212, opened 2017-08-22
Changes BGS reobservation scheme to match BGS spec

apcooper:
Changes to calc_numobs in targets.py to implement the following:

- BGS targets can be reobserved on every epoch, there is no maximum
  number of reobservations (although priorities change on ZGOOD).

- BGS targets should be observed no more than once on a single epoch
  (which can happen in tile overlaps).

### DIFF
--- a/py/desitarget/targets.py
+++ b/py/desitarget/targets.py
@@ -458,14 +458,22 @@ def calc_numobs(targets):
     isqso = (targets['DESI_TARGET'] & desi_mask.QSO) != 0
     nobs[isqso] = 4
 
-    # FIXME (APC): Better not to hardcode all this here? Took out the following
-    # for compatibility with earlier MWS tests
-    # SJB: better to not hardcode (for BGS and LRGs too), but until that is
-    # refactored we still need to request 2 observations for BGS_FAINT
-    #- TBD: BGS Faint = 2 observations
+    # BGS: observe both BGS target classes once (and once only) on every epoch,
+    # regardless of how many times it has been observed on previous epochs.
+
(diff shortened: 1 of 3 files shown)


apcooper:
These changes were made by Jianhua He (@JianhuaHe) and @forero, I just got nominated to write some comments and issue the PR. @JianhuaHe has tested the changes.

apcooper:
A couple of tests fail. One is a simple bug in cases where the target table is missing the `NUMOBS` column.

The other failing test checks for a `BGS_BRIGHT` target with one observation having priority `DONE`. In this new scheme BGS targets are never `DONE`.

forero:
@apcooper Can you also update the corresponding tests to match the current expected behavior?

JianhuaHe:
Hi, Andrew:

Current BGS strategy does not have DONE. Targets will always be re-observed if fibres are available.
Could you elaborate more on what is the first bug?

Cheers,
jianhua

forero:
The `NUMOBS` problem is a bug in the test code, not in the `calc_numobs` code itself. The problem is that in the test code there isn't a `NUMOBS` column in the `target` data structure, which is later required in the current implementation of `calc_numobs` https://github.com/desihub/desitarget/blob/master/py/desitarget/test/test_numobs.py

apcooper:
@forero  -- yes, I'll update the tests.

@jianhuahe --  the lack of detail in my comment is because this is a thread on github (see the link at at the bottom of the email). All the information is in the test logs linked to automatically from the github PR, so there's no need for me to repeat it. Moreover, the bug is simple, so I fixed it immediately -- but because I haven't yet updated the PR it's still worth noting for others why the tests are failing.

PR #215, opened 2017-09-15
Changes in mock targets --no-spectra

forero:
This PR:
* Allows for a restart from completed pixels.
* Solves a minor bug in completing in lyaqso generation when there is no lyaqso.
* Solves a major bug in the case of negative numbers for `ndens`.

### DIFF
--- a/py/desitarget/mock/build.py
+++ b/py/desitarget/mock/build.py
@@ -1254,11 +1254,30 @@ def targets_truth_no_spectra(params, output_dir='.', realtargets=None, seed=None
         
     log.info('Writing to output directory {}'.format(output_dir))
     print()
+    
 
+    
     # Default set of healpixels is the whole sky (yikes!)
     if healpixels is None:
         healpixels = np.arange(hp.nside2npix(nside))
 
+        
+    #shortening pixel list if some of them were already computed
+    log.info('pixels to do [before file testing] {}'.format(healpixels))
+    donepix = []
+    for i in range(len(healpixels)):
+        pixnum = healpixels[i]
+        targetsfile = mockio.findfile('targets', nside, pixnum, basedir=output_dir)
+        truthfile = mockio.findfile('truth', nside, pixnum, basedir=output_dir)
+        if os.path.exists(targetsfile) and os.path.exists(truthfile):
+            log.warning('Removing pixel {} from list as files {} and {} already exist'.format(healpixels[i], targetsfile, truthfile))
+            donepix.append(i)
+    if len(donepix)>0:
+        healpixels = np.delete(healpixels, donepix)
+    log.info('pixels to do [after file testing] {}'.format(healpixels))
+    if len(healpixels)==0:
+        return
+        
     areaperpix = hp.nside2pixarea(nside, degrees=True)
     skyarea = len(healpixels) * areaperpix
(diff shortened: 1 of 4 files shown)


py/desitarget/mock/build.py
```diff
@@ -1254,11 +1254,30 @@ def targets_truth_no_spectra(params, output_dir='.', realtargets=None, seed=None
...
+        if os.path.exists(targetsfile) and os.path.exists(truthfile):
+            log.warning('Removing pixel {} from list as files {} and {} already exist'.format(healpixels[i], targetsfile, truthfile))
+            donepix.append(i)
+    if len(donepix)>0:
+        healpixels = np.delete(healpixels, donepix)
+    log.info('pixels to do [after file testing] {}'.format(healpixels))
+    if len(healpixels)==0:
+        return
+        
     areaperpix = hp.nside2pixarea(nside, degrees=True)
```
moustakas: This check happens in `mpi_select_mock_targets` and I think we should add it to `select_mock_targets` as well; not here.  In particular, the only time I call `mock.build.targets_truth` directly is when I'm testing (usually with a single healpixel), in which case I don't want that healpixel to be skipped.

py/desitarget/mock/io.py
```diff
@@ -642,7 +642,7 @@ def read_gaussianfield(mock_dir_name, target_name, rand=None, bricksize=0.25,
             lyafiles = lyafiles_qso
             lyahdu = lyahdu_qso
             templatetype = templatetype_qso
-            templatesubtype = templatesubtype_qso
+            templatesubtype = templatesubtype
         else:
```
moustakas: This is definitely a bug but it should be fixed on line 611, not here.  This line (645) should still be `templatesubtype = templatesubtype_qso` but then line 611 should be `templatesubtype_qso = np.repeat('', nobj_qso)`.

py/desitarget/mock/selection.py
```diff
@@ -362,6 +362,8 @@ def density_select(self, targets, truth, source_name, target_name, density=None,
             if density and n_in_brick > 0:
                 mock_density = n_in_brick / brick_area
                 desired_density = float(self.brick_info['FLUC_EBV_{}'.format(source_name)][brickindx] * density)
+                if desired_density < 0.0:
+                    desired_density = 0.0
```
moustakas: Why would `desired_density` ever be negative?  It seems to me that this edge case should be captured much earlier..

py/desitarget/mock/io.py
forero: All right. I will make that change.

py/desitarget/mock/selection.py
forero: I agree this should be captured earlier. I believe that `desired_densities` are being generated with a normal distribution around the mean desired density, that might be the source of the problem. But in the way density_select is written we weren't protected against that case.

py/desitarget/mock/build.py
forero: OK. I will remove it from here.

forero:
@moustakas I made the changes you requested but the tests are failing because there is a problem with `speclite`. Do you know how to solve that problem?

sbailey:
The default version of sphinx recently increased to 1.6, but that is incompatible with astropy <2.  Using SPHINX_VERSION=1.5 has worked in other repos; I just pushed that change here.

PR #216, opened 2017-09-21
stdstar selection without obsconditions

sbailey:
It appears that we messed up our implementation of stdstar selection in PR #208. The code in master doesn't select or output any targets as standard stars (!).  This PR reverts the logic to just select (STD_FSTAR or STD_WD) for dark/gray time and (STD_BRIGHT or STD_WD) for bright time.  It doesn't explicitly check the target OBSCONDITIONS flag, since it appears that *all* mock standards are flagged as MWS targets anyway and thus are flagged for dark/gray/bright conditions.

Adding BRIGHT to the list of allowable conditions for STD_WD in desitargets.yaml in commit [7a6990](https://github.com/desihub/desitarget/commit/7a69904a1ee2055d0206952288caccb267095d86#diff-0f465bc9b95c5e21ca142c26fcfd2fa8) was still correct for bookkeeping, but for the purposes of stdstar selection, it seems we should just use the targetbits without also checking obsconditions.

Please double check me, @apcooper.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,7 @@ env:
         # - NUMPY_VERSION=1.10
         # - SCIPY_VERSION=0.17
         - ASTROPY_VERSION=1.2.1
-        # - SPHINX_VERSION=1.3
+        - SPHINX_VERSION=1.5
         - DESIUTIL_VERSION=1.9.5
         - DESISPEC_VERSION=0.11.0
         - DESISIM_VERSION=master
--- a/py/desitarget/mock/build.py
+++ b/py/desitarget/mock/build.py
@@ -621,6 +621,7 @@ def targets_truth(params, output_dir='.', realtargets=None, seed=None, verbose=F
     alltargets = list()
     alltruth = list()
     alltrueflux = list()
+
     for source_name in sorted(params['sources'].keys()):
         # Read the mock catalog.
         target_name = params['sources'][source_name]['target_name'] # Target type (e.g., ELG)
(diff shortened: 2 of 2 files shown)


apcooper:
The revision looks good and indeed the changes in #208 mean no STDs are written, but I'm slightly confused about the reason, specifically the connection to `OBSCONDITIONS` and 
> all mock standards are flagged as MWS targets anyway and thus are flagged for dark/gray/bright conditions.

Let me check I've understood: this statement applies only to the standards (rather than all MWS targets), because they have *both* `DESI_TARGET` and `MWS_TARGET` set, which means that somewhere their `OBSCONDITIONS` are set to the `&` of the `OBSCONDITIONS`  for their MWS class (e.g. `MWS_MAIN`) and those for their 'STD' class (e.g. `STD_FSTAR`). Correct? Doesn't that mean that every standard will be selected for dark and bright time rather than none of them being selected for either? Maybe you meant these were two separate bugs, which I agree with.

Let's go ahead with this PR, since it fixes both.

I worry that that ambiguity (extra `OBSCONDITIONS` for targets that get a flag in `DESI_TARGET` because they're special for some reason) might cause problems somewhere else, but I've no idea how to fix it.

For the record, to me it looks like the obvious problem here (nothing being written) is in the way I was accumulating the results from successive `stdbits` in the loop with `&=` here: https://github.com/desihub/desitarget/blob/35b3fe3b547fe8057ac8b70896d56f5305aa420a/py/desitarget/mock/build.py#L863
given that
https://github.com/desihub/desitarget/blob/35b3fe3b547fe8057ac8b70896d56f5305aa420a/py/desitarget/mock/build.py#L848
i.e. `istd` starts off as all `True` (for the targets in the pixel anyway).

So I think the following would also fix this in the spirit of #208:
```python
    for stdsuffix in ('dark', 'bright'):
        # Standards taken from targets
        istd = np.zeros_like(inpixel)

        # Select all targets with each of the standard star types and
        # matching obsconditions.
        for stdbit in ('STD_FSTAR', 'STD_BRIGHT', 'STD_WD'):
            is_std_type = (targets['DESI_TARGET'] & desi_mask.mask(stdbit)) != 0

            if stdsuffix == 'dark':
                has_right_obsconditions = (targets['OBSCONDITIONS'] & (obsconditions.DARK | obsconditions.GRAY) != 0)
            elif stdsuffix == 'bright':
                has_right_obsconditions = (targets['OBSCONDITIONS'] & (obsconditions.BRIGHT) != 0)
            istd |= (is_std_type & has_right_obsconditions)
        istd &= inpixel
```
although that keeps the slightly ugly copying of `inpixel` and relies on sensible `OBSCONDITIONS`.

sbailey:
Yes, there are two separate issues at play here:
1. the code wasn't outputting any standards at all due to & vs. | order of accumulations.
2. whether OBSCONDITIONS could/should be used for standard star selection vs. just using the targeting bits.

Regarding (2): my first attempt at fixing this was to select all targets with *any* STD bit set, and then pick the subset of those with the obsconditions matching dark/gray or bright.  But imagine a target that is faint enough to be useful as a dark standard (STD_FSTAR) but not a bright standard (not STD_BRIGHT), but is still an interesting MWS science target (MWS_ANY).  Because it is a MWS_ANY target, it gets OBSCONDITIONS BRIGHT set, even though it can't be used as a bright standard.

Thinking about it a bit more, I think the fully correct solution would be to check the obsconditions of the target bit definitions instead of the obsconditions of the targets themselves.  If we start having N>>1 standard star bits and subtleties of which ones go with dark vs. gray vs. bright, let's revisit that.  In the meantime, thanks for the check and I'll merge this as-is (once travis sphinx tests pass...)

PR #217, opened 2017-09-23
Refactoring density fluctuation code used by mocks...

geordie666:
...to calculate statistics in HEALPixels instead of in bricks.


geordie666:
Good point...let me add NSIDE and the nesting scheme to the header of the output file, and then I'll re-push the code and merge this.

PR #218, opened 2017-09-25
Modify morphology cut for QSO selection for DR5

Cyeche:
New morphology definition for DR5 and for the next data releases

### DIFF
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -364,7 +364,7 @@ def isQSO_colors(gflux, rflux, zflux, w1flux, w2flux, optical=False):
 
     return qso
 
-def isQSO_cuts(gflux, rflux, zflux, w1flux, w2flux, w1snr, w2snr, deltaChi2,
+def isQSO_cuts(gflux, rflux, zflux, w1flux, w2flux, w1snr, w2snr, deltaChi2, release,
                objtype=None, primary=None):
     """Cuts based QSO target selection
 
@@ -397,7 +397,7 @@ def isQSO_cuts(gflux, rflux, zflux, w1flux, w2flux, w1snr, w2snr, deltaChi2,
     qso &= w1snr > 4
     qso &= w2snr > 2
 
-    qso &= deltaChi2>40.
+    qso &= ((deltaChi2>40.) | (release>=5000) )
 
     if primary is not None:
         qso &= primary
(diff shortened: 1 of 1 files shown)


geordie666:
@Cyeche : I'm happy to merge this once unit tests pass. At a minimum, it looks like you'll need to add the `RELEASE` keyword to unit test calls to the isQSO... functions.

Cyeche:
In the future (once DR6 will replace DR4) we may remove this dependency to « release". Even I guess that we may have different algorithms for north and south.

sbailey:
Requiring `release` in `isQSO_cuts` is an API change that would break any other code that was calling it (as happened with the unit tests).  I don't see any calls to it in desisim, so we may not have any cases outside of desitarget.

`release` is an optional parameter to `isQSO_randomforest`, but it looks like if you use the default (None) it will crash the code trying to do the `(release>=5000)` test.

So: is the intension that `release` is required or not?  I'd rather not make it temporarily required if you expect to deprecate it in the future.  But it seems likely that we will need to have cuts tuned for North vs. South, and it seems that requiring `release` is the way to enable that.

Cyeche:
Currently, we have 3 configurations:

DR3: DECaLS (south hemisphere) with PSF/SIMP morphological models
DR4:  MzLS+BASS (North hemisphere) with PSF/SIMP morphological models
DR5: DECaLS (south hemisphere) with PSF/REX morphological models


Basically, we want to be able to run the 3 configurations. I was thinking that a test over the release number will allow us to run the 3 configurations. If you have another method I’m ready to use it. 

> Requiring release in isQSO_cuts is an API change that would break any other code that was calling it (as happened with the unit tests). I don't see any calls to it in desisim, so we may not have any cases outside of desitarget.
> 
> release is an optional parameter to isQSO_randomforest, but it looks like if you use the default (None) it will crash the code trying to do the (release>=5000) test.
> 
> So: is the intension that release is required or not? I'd rather not make it temporarily required if you expect to deprecate it in the future. But it seems likely that we will need to have cuts tuned for North vs. South, and it seems that requiring release is the way to enable that.
> 
>

geordie666:
I think we'll have to break the API at some point to have a North versus South set of selections, whether it is done explicitly on "North" v "South" or by `RELEASE`. So, I'm happy for @Cyeche to start passing `RELEASE` around. I was waiting for this to be done in a way that doesn't break unit tests before merging.

One option would be to default to, say, `RELEASE=5000` for a particular target selection function if `RELEASE` is `None` or if `RELEASE` is not passed to that function. Then downstream code for the data challenges would default to the DR5 selection unless they specified a `RELEASE` number. But, again, I'd want to see @CYeche's code written such that it defaults to `RELEASE=5000` in a way that doesn't break the current unit tests (in particular, in  a manner doesn't specifically doesn't break `cuts.isQSO_cuts`, for now).

sbailey:
Superseded by PR #239 which has been merged.  Closing this one.

PR #219, opened 2017-09-27
add more uniform than random sky mocks

sbailey:
We are currently using /project/projectdirs/desi/mocks/GaussianRandomField/v0.0.1/2048/random.fits for the sky locations.  These have an appropriate average density, but due to random fluctuations it can happen that a positioner has no science targets, standard stars, or sky locations within its reach.  This PR provides code for generating "more uniform than random" sky locations that are random smaller than some scale but have a guaranteed uniformity above that scale.  The default is healpix nside=2048 which corresponds to about half of a fiber patrol radius, and it guarantees that every nside=2048 pixel that covers the DESI footprint will have one sky location such that every fiber should have two sky options.

The final sky catalog will be more complicated than this and based upon actually avoiding objects while still guaranteeing enough sky locations for every fiber and not just on average, but this code can generate mocks closer to that.

The following plots show a comparison of the current sky locations (left) with the proposed new ones (right)

newsky

A histogram of the densities of nside=2048 pixels covering the DESI footprint shows that the current scheme (blue) often has no targets within a pixel, while the new scheme (orange) guarantees one sky fiber per pixel.

sky-coverage

Implementation notes:
* due to the odd shapes of healpixels, the targets aren't exactly random within a pixel and use an iterative algorithm to fill in points that randomly scattered outside of the pixel to which they were assigned.  A more clever algorithm could replace this.
* example file at /project/projectdirs/desi/users/sjbailey/mocks/sky/0.1/uniformsky.fits.  Once vetted this would be moved into the standard DESI project mock directory structure.
* this new catalog needs desihub/desitarget#82 to be addressed to fully make use of the extra uniformity

And to state the obvious (once you've thought about it for a moment): these "more uniform than random" sky locations should *not* be used for random mocks for LSS studies, even for quick-and-dirty checks.  They specifically *aren't* random.

Usage:
```python
from desitarget.mock.sky import random_sky
skyra, skydec = random_sky(nside=2048)
```
There are also options for passing in a subset of the DESI tiles to cover, and a `maxiter` parameter to make sure that the iterative algorithm doesn't go on forever.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -17,11 +17,13 @@ desitarget Change Log
    * Correct random seeds when processing pix in parallel
    * Misc other small bug fixes
 * Added ``mpi_select_mock_targets``
+* Added ``desitarget.mock.sky.random_sky`` [`PR #219`_]
 
 
 .. _`PR #200`: https://github.com/desihub/desitarget/pull/200
 .. _`PR #210`: https://github.com/desihub/desitarget/pull/210
 .. _`PR #214`: https://github.com/desihub/desitarget/pull/214
+.. _`PR #219`: https://github.com/desihub/desitarget/pull/219
 
 0.14.0 (2017-07-10)
 -------------------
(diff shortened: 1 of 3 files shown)


dkirkby:
Nice!

PR #220, opened 2017-09-28
First (simplistic) code for generating blank sky positions.

geordie666:
We'll eventually move to using the imaging surveys blob files, but this "each object is an avoidance zone on the sky" approach will work in a pinch, and also might be useful for mocks.


py/desitarget/skies.py
```diff
@@ -0,0 +1,486 @@
...
+            if len(w[0]) > 0:
+                goodskies[idskies[w]] = False
+
+        #ADM guard against the case where there are no objects with large radii
+        if len(bigsepw) > 0:
+            #ADM match the large-separation objects and flag any skies that match such an object
+            log.info('Matching additional positions out to {:.1f} arcsec...t = {:.1f}s'
+                     .format(maxrad,time()-start))
+            idskies, idobjs, d2d, _ = cobjs[bigsepw].search_around_sky(cskies,maxrad*u.arcsec)
+            import pdb ; pdb.set_trace()
```
sbailey: leftover `pdb.set_trace()`.  I hit this when testing on DR4 sweep-160p035-170p040.fits.  Please confirm that this case is working as expected and remove `pdb.set_trace()` or otherwise further debug why this was needed.

bin/select_targets
```diff
@@ -50,8 +52,18 @@ if ns.check:
...
+
+    if ns.skies:
+        skies = select_skies(infiles,numproc=ns.numproc)
+        ntargs = len(targets)
+        nskies = len(skies)
+        nobjs = ntargs + nskies
+        targs = np.empty(nobjs, dtype=targets.dtype)
+        targs[0:ntargs] = targets
+        targs[ntargs:nobjs] = skies
+        targets = targs.copy()
```
sbailey: I don't think this copy is necessary.  Simply `targets = targs` would allow the I/O to proceed without having 2 copies in memory (python garbage collection might not get around to cleaning it up anyway before finishing, but at least this would give it the option of doing so if needed).

sbailey:
Below is a plot for a piece of DR4 sweep-160p035-170p040.fits showing tractor object (blue) and identified sky locations (orange) -- i.e. not nearly enough sky locations.
sky

Zooming in on a region shows blank areas where we would want a sky target that are much larger than typical separations between the sky targets that do exist and the nearest sweep objects:
sky-zoom

Please debug these cases to see why we aren't getting more sky and adjust parameters to get a good sky density over an entire sweep.

py/desitarget/skies.py
geordie666: Thanks, this highlighted a bug in my unit test coverage where none of the test objects flagged this clause in the code. I've removed this `pdb.set_trace()` and improved the unit test coverage.

bin/select_targets
geordie666: Agreed. I've removed the `copy()`

geordie666:
I tracked the issue of the coverage of the good sky positions to a handful of galaxy-like objects in the sweeps with very large half-light radii. There are ~300 objects of type `EXP` or `DEV` that have radii of > 100 arcseconds in a typical sweeps file. Here's a plot of what those objects look like (assuming they're circular, which is actually reasonable for many of them). The "large" objects are in blue, with the "good" sky positions in black and the "bad" ones in red:

skies

I've been bugging @dstndstn for a way to cull these objects, and it could just be a case of ignoring anything with a large half-light-radius above a certain threshold in flux (i.e. this could be a feature where bright objects have unrealistically large wings). In any case, I'm looking into a good way to fix this.

These objects probably shouldn't be a problem when we ultimately move to a more sophisticated algorithm based on the "blob" files.

geordie666:
@sbailey: I implemented an elliptical mask formalism to help improve the available space for sky fibers. For your problem area, here's a plot of the new "avoidance zones" due to resolved objects (ellipses) and the existing unresolved (PSF objects):
objects
and, here's the resulting sky positions (black circles are "good" SKY positions and red circles are BADSKY positions):
skies
There are a few things to note:

- The process for dealing with elliptical objects is slow (although the sweeps would still all run in a day or so even with the code in this PR). I think I can speed things up considerably by performing a circular positional match first, to reduce the number of objects that need tested against an ellipse. I want to move onto DR5 target stuff, though, so let's leave that for later.

- I stupidly merged master into this branch earlier in the week when I wasn't paying close attention. I think that history was preserved in a reasonable way, but if you disagree let me know and I'll delete this branch, build a new one of master, and then add all of the new code as a single commit. Otherwise, please merge, so I can move on to other things.

sbailey:
OK, let's merge this after tests pass Travis and spin other updates / improvements to new tickets and PRs.

When you run target selection on DR5, please also make a matching set of DR4+DR5 sky and badsky locations using this code.  Thanks.

PR #222, opened 2017-10-04
fixes installs missing GMM mock data

sbailey:
This PR fixes an installation bug that wasn't copying the mock GMM data into the install location.


sbailey:
Tests are failing because DESI svn server is unavailable for fetching desimodel data.  Will retry when power is back at LBL.

Issue #227, opened 2017-10-11
Healpix pattern in the mock outputs 


forero:
The current version of `mpi_select_mock_targets --no-spectra` produces an evident healpix pattern in the outputs that were downsampled to have a mean number density

![image](https://user-images.githubusercontent.com/1848618/31419902-64dc7dd0-ae79-11e7-9da6-451e5ff98c12.png)

![image](https://user-images.githubusercontent.com/1848618/31419914-6d221f9a-ae79-11e7-92a7-d77f7f62635d.png)

![image](https://user-images.githubusercontent.com/1848618/31419919-7240d5e8-ae79-11e7-919a-83b04c7fcfe2.png)

![image](https://user-images.githubusercontent.com/1848618/31419926-7c4fb220-ae79-11e7-9eff-a19e6ef2a723.png)

The reason for this is that the parallelization is done on healpixels and the number density is estimated using bricks, for the bricks that are on the boundaries between healpixels the number density estimation is wrong. In other words, the pattern comes from the bricks being cut between two neighboring healpixels.

I am working on a solution and should have a PR ready during the day.

moustakas:
Closed in #228

PR #228, opened 2017-10-11
New select mock targets --no-spectra without bricks

forero:
This PR refactors `targets_truth_no_spectra` and solves bug #227.


forero:
Some documentation and additional tests are still pending, but I leave this open for others to start reviewing the code.

forero:
Results look much better now. The healpix boundaries are not there anymore. But there is still a glitch at the boundary for the QSOs

image
image
image
image

bin/mpi_select_mock_targets
```diff
@@ -36,7 +36,6 @@ parser.add_argument('--realtargets', '-r', help='Path to real target catalog', t
 parser.add_argument('-v','--verbose', action='store_true', help='Enable verbose output.')
 parser.add_argument('--no-check-env', action='store_true', help="Don't check NERSC environment variables")
 parser.add_argument('--sort-pixels', action='store_true', help="Sort pixels by galactic latitude")
-parser.add_argument('--no-spectra', action='store_true')
```
sbailey: For the record, from an email with @forero: `select_mock_targets --no-spectra` is now fast enough that we don't necessarily need an MPI wrapper for it.  We may want to re-enable this feature in the future to enable running the full survey in a few minutes, but ok for this to be disabled for now.

py/desitarget/mock/build.py
```diff
@@ -1073,436 +1078,290 @@ def join_targets_truth(mockdir, outdir=None, force=False, comm=None):
...
             Output directory (default '.').
-        realtargets : astropy.table
-            Real target catalog table, e.g. from DR3 (deprecated!).
-        bricksize : float
-            Brick size for assigning bricknames, which should match the imaging
-            team value (default 0.25 deg).
         nproc : int
             Number of parallel processes to use (default 1).
-        nside : int
+        healpix_nside : int
```
sbailey: minor: everywhere else throughout our code we just use `nside` and not `healpix_nside`

py/desitarget/mock/build.py
```diff
@@ -1073,436 +1078,290 @@ def join_targets_truth(mockdir, outdir=None, force=False, comm=None):
...
-    del brick_info # memory clean-up
+    return [targets, truth]
 
-    # Consolidate across all the mocks.  Note that the code quits if alltargets
-    #is zero-length, even if skytargets is non-zero length. In other words, if
-    #the parameter file only contains SKY the code will (wrongly) quit anyway.
-    if len(alltargets) == 0:
-        log.info('No targets; all done.')
-        return
+def target_selection(Selection, target_name, targets, truth, healpix_nside, healpix_id, seed, rand, log,output_dir):
```
sbailey: please add docstring (here and elsewhere for newly added functions)

py/desitarget/mock/build.py
```diff
@@ -1073,436 +1078,290 @@ def join_targets_truth(mockdir, outdir=None, force=False, comm=None):
...
+        healpix_nside : int
             Healpix resolution corresponding to healpixels (default 16).
         healpixels : numpy.ndarray or int
             Restrict the sample of mock targets analyzed to those lying inside
-            this set (array) of healpix pixels.
-
+            this set (array) of healpix pixels. Default (None)
+        
     Returns:
-        A variety of fits files are written to output_dir.
```
sbailey: What does initialize do?  "Initializes variables..." what variables?  Does it write any files anymore?  Create directories?  Change some global variables?

forero:
@sbailey that error message was a real bug. If you run the same command it works. It will take ~1 hour to run. If you run it with `--nside 4` it takes 5 minutes. Running time is proportional to the amount of pixels it has to process.

py/desitarget/mock/build.py
forero: Changed.

forero:
Latest full sky results look good
image
image
image
image
image

py/desitarget/mock/build.py
forero: I have updated all the docstrings in the new functions.

forero:
All comments from @sbailey are addressed/solved. This looks good to me.

moustakas:
Will try to review this by today---thanks.

forero:
@moustakas thanks! 
Integration tests are failing with
>/home/travis/build/desihub/desitarget/py/desitarget/mock/build.py:docstring of desitarget.mock.build.downsample_pixel:3: ERROR: Unexpected indentation.

I don't see the problem in the docstring. Do you understand why is this happening?

sbailey:
Alas, sphinx/rst are very picky.  I think you need a blank line before "Args" and "Options" and a colon instead of a period.  e.g.
```
function description.

Args:
    blat: array
       stuff about blat...
```
instead of
```
function description
Args.
    blat: array
        stuff about blat....
```

This formatting change applies to many of your new docstrings.

To test this locally without going via travis, you can use `python setup.py build_sphinx`.  You'll need sphinx 1.5 (not 1.6) for that.

forero:
Sphinx build is working. 
@moustakas @sbailey any new suggestions?

forero:
> Most of the new code / refactoring in mock.build is used for the new targets_truth_nospectra function even though targets_truth needs / uses a lot of the same machinery. I guess this can all stay separate although it makes it hard for others to figure out what's going on and makes for a lot of duplicated code.

I made the refactor to allow the possibility for the next iteration in `targets_truth` to explicitly reuse those functions and move to a merge where the only thing we need in `targets_truth` is an `spectra=False` option.

> You loop on healpixels (outer loop) and then target types (inner loop) but since not all the mock catalogs are organized by healpixels it means that there's more I/O overhead than there could be. This is probably fine though since we still only read the ra,dec boundaries we need, so it's not as bad as it used to be.

It runs just fine, much better than it used to.

> Why don't you use the density fluctuations results (#217)?

Because it needs significantly new code. It will come in a new PR.

forero:
Merging now.

forero:
@moustakas BTW, the option for contaminants was also removed from `--no-spectra` because it uses bricks and not pixels. This will have to be reintroduced again.

Issue #223, opened 2017-10-06
General clean-up of target selection code


geordie666:
Some outstanding irritating features of `desitarget` that need to be corrected

- `HPXNSIDE` and `HPXNEST` should be keywords but not dependencies
- LRG targeting needs to be split into bits on the basis of magnitude cuts
- `RELEASE` should be an option for all sub-functions for region-dependent cuts. Default to DR5

geordie666:
Closed in https://github.com/desihub/desitarget/pull/229 and by spinning off individual items to https://github.com/desihub/desitarget/issues/230 and https://github.com/desihub/desitarget/issues/231

Issue #198, opened 2017-07-07
Use `desiutil.log` instead of `print`


geordie666:
Use `desiutil.log` instead of `print` throughout the `desitarget` code

geordie666:
Closed in https://github.com/desihub/desitarget/pull/229

Issue #120, opened 2016-11-17
io.py recursively descends directories to look for sweeps and tractor files


geordie666:
This is dangerous. For example, now we have dr3.0 and dr3.1 sweeps directories in the dr3 release, io.iter_files could collect _both_ sets of sweeps instead of just one set. Since we made dr3.1 I've been using this with care to only retrieve one set of sweeps files, but we should probably fix it to insist on better specification of the exact directory of interest.

sbailey:
I'm not sure what there is to do here other than make it clear to users to specify a sweep version subdirectory and not just a top level sweep or data release directory.  It is useful for development to be able to copy an arbitrary set of sweeps or tractor files to another location and use them as-is without select_targets imposing a specific directory structure.

We could cache the filenames that it finds while traversing the directory tree and raise an error if it finds the same filename in two different directories.  That isn't bulletproof, but could help.  Even then, I'd offer an override option to force select_targets to take whatever it finds, including repeats.

geordie666:
I like the option of flagging an error if you traverse two different directories unless the user sends an override flag. Leave that with me. I don't think this is critical...it's just very easy at the moment to process targets twice given our dr3/dr3.1 directory structure.

sbailey:
To be clear: I think the error should be if you find the same filename in two different directories, not just that you were traversing two different directories.  The sweeps happen to be in a single directory per release, but select_targets should also work on tractor files that are grouped into many directories (but without repeated filenames).

geordie666:
Yes...agreed. I worded my response badly.

geordie666:
Closed in https://github.com/desihub/desitarget/pull/229

PR #229, opened 2017-11-01
General clean-up prior to running DR5 official targets

geordie666:
Includes some of the following changes
   - Use :mod:`desiutil.log` instead of verbose (everywhere except mocks)
   - Change HEALPix references to header keywords instead of dependencies
   - Include SUBPRIORITY and shape parameter IVARs in target outputs

I didn't complete all of the issues in this milestone:

https://github.com/desihub/desitarget/issues/223

So some of them will have to get pushed into a new issue. On the other hand, I did fix additional issues that weren't mentioned in that milestone such as:

https://github.com/desihub/desitarget/issues/120
https://github.com/desihub/desitarget/issues/198

I've checked that the code runs to completion on DR4 and DR5 sweeps files.


sbailey:
Looks good.  Thanks.  Please close the related issues and open new ones as needed.  Merging now.

Issue #231, opened 2017-11-01
RELEASE should be a keyword option for all sub-functions for region-dependent cuts. 


geordie666:
In particular to allow different Target Selection cuts using the Northern and the Southern imaging. We should default to DR5, for now.

Spinning this off from https://github.com/desihub/desitarget/issues/223 so that it's its own thing.

geordie666:
Region-dependent cuts have been possible for some time, now. In a few cases (for the QSO RF) we actually use `RELEASE`, but for the most part we just pass a `south=True` kwarg instead.

PR #239, opened 2017-11-14
Rebase for Christophe Yeche's updates to the desitarget quasar selection

geordie666:
This PR rebases @Cyeche's PR:

https://github.com/desihub/desitarget/pull/218

With the rebase, unit tests pass more easily because master contains new test files that include the `RELEASE` column. Passing that column to `isQSO_cuts` was sufficient to pass the unit tests.

This PR therefore:

- updates `cuts.py` to allow different star-galaxy separation for different `RELEASE` numbers

We'll likely need a better refactor to add this capability for *all* target classes (as noted in https://github.com/desihub/desitarget/issues/231)

### DIFF
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -364,7 +364,7 @@ def isQSO_colors(gflux, rflux, zflux, w1flux, w2flux, optical=False):
 
     return qso
 
-def isQSO_cuts(gflux, rflux, zflux, w1flux, w2flux, w1snr, w2snr, deltaChi2, release,
+def isQSO_cuts(gflux, rflux, zflux, w1flux, w2flux, w1snr, w2snr, deltaChi2, release=None,
                objtype=None, primary=None):
     """Cuts based QSO target selection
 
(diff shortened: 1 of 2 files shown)


geordie666:
@sbailey: I didn't update the changes file for this PR, as I new you were going to make a tag. You could add:

- allows different star-galaxy separations for quasar targets for different release numbers

(or some-such)

to `changes.rst`

sbailey:
`isQSO_cuts` now *requires* `release` as an input.  I don't think this breaks anything for desisim or desitarget.mocks, which uses `isQSO_colors`, but just highlighting that change of API because it might break something.  Is making `release` required really necessary, or could a default allow it to work with backwards compatibility?

In contrast, `isQSO_randomforest` does not require `release` since it has a default as `None`, but it looks like if it isn't provided the code will choke on `qso &= ((deltaChi2>30.) | (release>=5000) )`.  It should either be required in the function signature, or have a default+code that won't choke on that default.

geordie666:
Good catch on the `release=None`, I'll remove that as a default.

Note that if people are using `io.read_tractor` to read sweeps or Tractor files, then `RELEASE` is automatically added as a column. So, there is no backwards-compatability issue with `RELEASE`. If people are *not* using `io.read_tractor` to read sweeps or Tractor files, and are running on pre-DR4 files, then they're going to have far more problems than missing the `RELEASE` column.

geordie666:
In the end, I decided to default to `RELEASE=5000` in the event of `release=None` for both `isQSO_cuts` and `isQSO_randomforest`. That way we don't kill anything downstream, in the short-term, just to make a modification for the quasar targets. Ultimately, though, we'll need a wider refactor so that, e.g., the mocks are aware of a dependency on `RELEASE` for different target classes.

@sbailey: Please merge unless you think that this is going to break your mock-based workflow.

moustakas:
I did follow this thread and everything looks good from my perspective -- thank you.

PR #248, opened 2017-12-15
add skyfibers.py script, to (re-)generate possible sky fiber locations from LegacySurvey DR outputs

dstndstn:


geordie666:
This code is currently divorced from the overall machinery in desitarget, but it's going to be very useful for assigning sky fibers (thanks @dstndstn).

I need to write some unit tests for coverage. I also need to merge the overall structure of the script into the `select_targets` formalism. I won't get around to hooking things together for a few weeks, though, and this *is* mature and working code, so I'm merging this (standalone) module to prevent it lagging master.

Issue #252, opened 2017-12-21
select_mock_targets -- desired updated features and bug fixes


moustakas:
Current high-priority punch list.

- [ ] Ensure all targets -- including contaminants -- have the correct / desired target density (see #192 for QSOs).

- [ ] Use simqso to generate QSO spectra with the correct underlying QSO luminosity function (see https://github.com/desihub/desisim/pull/293)

- [ ] The MWS photometry written out in the truth catalog may have an indexing bug (see #251).

- [ ] Fix duplicate TARGETIDs (#178).

- [ ] Add units to target and truth files (#237).

- [ ] Validate we can use the `MOCKID` in the output target catalog to get back to the correct row in the original mock catalog (#88).

- [ ] Use GMMs based on DR5 and deconvolved for photometric scatter (see #250).

- [ ] Compare columns in mock and data target catalogs and assess (related to #209). For example, I've already added `PSFDEPTH_[W1,W2]` to the mock target catalog.

moustakas:
Most of these issues were addressed in #264, @sbailey has opened new outstanding issues, and I'll open some more.  Closing this for now.

PR #254, opened 2017-12-29
new density fluctuations model

moustakas:
@geordie666 @forero I started playing around with a new target density fluctuations model using mixtures of Gaussian functions as part of #252.  If you get a chance could you take a look at [this notebook](https://github.com/desihub/desitarget/blob/new-fluctuations-model/doc/nb/target-fluctuations.ipynb) and send comments?


moustakas:
Once tests pass I'd like to self-merge this, since it is needed for #261 and the only substantive change is a new notebook (linked above).   Comments are still welcome but after a discussion with @geordie666 I've decided to move forward with a simple linear regression of density versus Galactic reddening.  I will reach out to the Clustering WG to see if they can develop a more realistic / sophisticated model for our use.

One question for @geordie666, though:  Do you know why the QSO and LRG target densities are gridded / binned in the `hp-info-dr5-0.17.1.fits` file (see, e.g., cells 28 and 30)?

geordie666:
@moustakas: Can you elaborate on what you mean by gridded/binned? `DENSITY_QSO` and `DENSITY_LRG` are single numbers for a cell, so it's hard for me to see any gridding.

moustakas:
Here is the QSO density vs EBV (directly from the "info" structure) from [this notebook](https://github.com/desihub/desitarget/blob/master/doc/nb/target-fluctuations.ipynb).  Any ideas on what's giving rise to this?  Shouldn't the density be a continuous variable (as it is for, e.g., ELGs and BGS)?

screen shot 2018-01-19 at 1 45 24 pm

geordie666:
I think this is a reasonable explanation:

At nside=256 the pixel area is 0.052456 sq. deg.

Therefore, if you have, 1, 2, 3, 4 QSOs in a pixel, etc., the density (and the `log10` of the density) is as follows:

```
np.arange(1,10)
    array([1, 2, 3, 4, 5, 6, 7, 8, 9])

np.arange(1,10)/0.052456
    array([  19.06359616,   38.12719231,   57.19078847,   76.25438463,
             95.31798078,  114.38157694,  133.4451731 ,  152.50876925,
            171.57236541])

np.log10(np.arange(1,10)/0.052456)
    array([ 1.28020483,  1.58123483,  1.75732608,  1.88226482,  1.97917483,
            2.05835608,  2.12530287,  2.18329482,  2.23444734])
```
If our quasar density is ~200 per sq. deg. then in a 0.052456 sq. deg. pixel your modal number of quasars should be ~10. Certainly, there will be pixels with 1, 2, 3 etc. quasars in them.

So, I think you're just seeing discreteness from this effect. We could merge adjacent pixels if they contain fewer than X sources?

moustakas:
Thank you, @geordie666, that explains it!  I don't think it's worth merging adjacent pixels -- I think this is just an artifact of an incomplete imaging survey.

PR #262, opened 2018-01-19
ADM mock QA

geordie666:
Numerous changes to the QA module to facilitate adding QA plots for mocks (targets and truth) to the standard QA targeting pages. Also includes some initial QA plots.

Issue #258, opened 2018-01-13
Add/organize mock target QA


sbailey:
We have various notebooks and emails about mock target QA — organize these into a standardized set of mock target QA numbers and plots similar to `desitarget/bin/run_target_qa` output.

moustakas:
Initial progress was made in #262 but let's keep this open for now.

geordie666:
We decided on today's telecon to close this and track it instead in the various "enhancement" issues that we span off:

https://github.com/desihub/desitarget/issues/266
https://github.com/desihub/desitarget/issues/267
https://github.com/desihub/desitarget/issues/268
https://github.com/desihub/desitarget/issues/269
https://github.com/desihub/desitarget/issues/270
https://github.com/desihub/desitarget/issues/271
https://github.com/desihub/desitarget/issues/272

Issue #253, opened 2017-12-22
should obsconditions be in mock targets catalog or not?


moustakas:
@forero In #207 (see especially https://github.com/desihub/desitarget/pull/207#issuecomment-321052091) after some discussion it was decided to *not* include `OBSCONDITIONS` in the mock targets catalog, since this is something that should be computed by the MTL on the fly.  

However, in https://github.com/desihub/desitarget/pull/211/commits/83627e28249e611f8319489bf567732240ef156a and https://github.com/desihub/desitarget/pull/211/commits/f43aa0e9f1d69a0f7cacca3b965cf1ccd4bc9a15 you added this back in.  (Mea culpa that I didn't ask this question sooner, since I was the reviewer on that PR!)

Perhaps @sbailey can weigh in as well, but should `OBSCONDITIONS` be in there or no?

moustakas:
Following up on this question.  @sbailey do you know, offhand?

sbailey:
Thanks for the reminder.

OBSCONDITIONS is a bitmask of allowable observing conditions for a given target, which should be completely derivable from the targeting bits themselves.  Having it in the targeting table itself may seem convenient, but it would be better to *not* have it in the table and provide utility functions instead, in case we change our mind about which bits are allowable under what conditions (then we just update targets.yaml instead of having to release a new targeting catalog just to update that column).  e.g.
```python
def get_obsconditions(desi_target):
    '''returns obsconditions mask for input array of desi_target targeting bits

    desi_target can be int or array-like; return value matches dimensionality
    '''
    ...

def ok_obsconditions(desi_target, conditions_mask):
    '''check whether desi_target can be observed under conditions_mask

    conditions_mask is int; desi_target can be int or array-like;
    return value matches dimensionality of desi_target
    '''
    ...
```
(or a better name than `ok_obsconditions`...)

moustakas:
It was decided in #264 to not include `obsconditions` in the output (mock) target catalogs, so I'm closing this.

Issue #251, opened 2017-12-21
MWS colors are oddly gridded in the dc17b truth catalogs


moustakas:
@segasai has identified odd / gridded colors in the MWS truth catalogs in the dc17b run, suggestive of an indexing bug somewhere.  Track this down and fix it.

![dc17b_cmd](https://user-images.githubusercontent.com/1431820/34260878-4daf15ca-e635-11e7-8afa-b703724e1fdc.png)

moustakas:
@segasai I looked into this issue in some detail as part of a refactor of the code which connects the Galaxia mocks to spectra, and I don't believe this is a bug -- it does indeed appear to be due to the gridding in the stellar templates, i.e., due to the fact that we're not interpolating in Teff and other stellar properties. 

For example, here's apparent (DECaLS) g-r color vs Teff for a mini dataset of a couple thousand MWS_MAIN targets:
![junk](https://user-images.githubusercontent.com/1431820/34507111-18e586a0-f001-11e7-9df7-bd68321da03f.png)

A high-priority ToDo item is to interpolate the stellar spectra so that this issue can be eliminated -- and @callendeprieto has code to do so -- but I have not had the cycles to incorporate it into the simulation infrastructure.

moustakas:
I'm fairly confident that this is an understood issue that should have improved somewhat with the higher density of stellar templates added in https://github.com/desihub/desisim-testdata/pull/9 and used in #264.  Meanwhile, we're also working on an improved (higher-density, irregularly gridded) set of stellar templates.

PR #264, opened 2018-01-23
major refactor of select_mock_targets

### LINKED ISSUE #88: Encode mock provenance into TARGETID.

### LINKED ISSUE #178: fix duplicate TARGETIDs in select_mock_spectra
The current method of having select_mock_target work in patches of sky, and then do bricks within that patch results in duplicate TARGETIDs since multiple sub-pieces get the same random seed.  Fix this, e.g. using the recipe in desisim.templates.

In the meantime, we are reassigning TARGETIDs while merging the select_mock_targets into the final bricks, using [two_percent_survey/join_bricks.py](https://github.com/desihub/two_percent_DESI/blob/master/join_bricks.py).

Note: final real data TARGETIDs will be derivable from imaging survey tractor outputs without requiring random numbers.  This problem is specific to generating TARGETIDs for simulated data in parallel.

### LINKED ISSUE #192: ensure that the QSO target densities aren't too low

### LINKED ISSUE #237: add units to target and truth files
From @weaverba137 in desihub/two_percent_DESI#6:

> The truth and targets files don't set units on any values, even obvious ones like RA, DEC.

This applies to both the output of `select_targets` and `(mpi_)select_mock_targets`.  For the mocks, units could be added during the merging by `join_mock_targets` but it would be better for them to be added in the original per-pixel files that are written.

### LINKED ISSUE #260: dc17b: discrete input high z qso
Carried over from https://github.com/desihub/desisim/issues/308 as reported by @ngbusca.

In dc17b, the input redshifts for high z quasar (above ~1.7) are discrete. Here's a plot of z_fit vs. z_input with a zoom at the transition redshift.

image

### LINKED ISSUE #261: refactor mock -> spectra connection
@moustakas is working on a major refactor of how input mocks are connected to template spectra to simulate.  I didn't find an issue that represents that work, so opening this one to facilitate tracking of when this is done.  If this is already covered by another issue, comment here and close this ticket.

This is probably related to a subset of the punchlist in #252, but that list is probably broader than just this refactor, which will simplify/enable many of those items.

moustakas:
This PR (which would close #261) includes a near-total rewrite of `select_mock_targets` (which really just calls `mock.build.targets_truth`) and introduces target-specific classes (in `mock.mockmaker`) which make it much easier, transparent, and extendable to connect mock catalogs to spectra.  There are still a variety of outstanding issues (see below), but I'd like to get the current (functional) code merged nowish and go from there.

I have tried to demonstrate some of the core functionality in [this notebook](https://github.com/desihub/desitarget/blob/fix-densities/doc/nb/connecting-mocks-and-spectra-examples.ipynb), which I'm hopeful will be used by various Working Groups to build smaller spectroscopic datasets.  Suggestions for additional tutorials or demonstrations of capabilities are welcome.

I tested this at NERSC with a single healpixel (using the `fix-densities` branch of `desitarget`):
```
salloc -N 1 -C haswell -t 00:60:00 --qos interactive -L SCRATCH,project
source /project/projectdirs/desi/software/desi_environment.sh master
package=desitarget
module unload $package
export PATH=/global/cscratch1/sd/ioannis/repos/$package/bin:${PATH}
export PYTHONPATH=/global/cscratch1/sd/ioannis/repos/$package/py:${PYTHONPATH}
```
then
```
cd /global/cscratch1/sd/ioannis/refactor
time select_mock_targets -c refactor.yaml --verbose --seed 123 \
  --healpixels 6507 --nside 32 --overwrite --nproc 32
join_mock_targets --mockdir . --force
run_target_qa ./targets.fits /project/projectdirs/cosmo/www/temp/ioannis/refactor \
  --nside 64 --mocks
```
which took about 7 minutes, with QA output here:
http://portal.nersc.gov/project/cosmo/temp/ioannis/refactor

Major improvements include:
* Incorporated a new (but still simplistic) target density fluctuations model (#254).
* Full utilization of healpixel organization; only reference to bricks is when generating the imaging-like `targets` catalog.
* Significantly better speed and memory usage.  I don't have numbers handy but this should be noticeably faster and scale much better to, e.g., the 2% footprint.
* Better documentation. streamlined functionality.

In addition:
* Simplified configuration (input) yaml file.
* Addresses #253 by removing `obsconditions`
* Fixes #260 - discrete QSO redshifts (maybe? I couldn't really reproduce this...)
* Fixes #88 (mock provenance in targetid -- should be tested, though)
* Fixes #178 (duplicate targetid)
* Addresses #201 (include r-band magnitude in truth catalog)
* Fixes #237 (add units to target and truth catalogs)
* Fixes #192 (too-low QSO and Lya target densities)
* QSO magnitudes are drawn from the BOSS/DR9 luminosity function (via `simqso`).
* Outputs are functional with nominal desitarget QA (see #258 and #262).

Known issues / problems / shortcomings:
* Still using the DR2-derived Gaussian mixture models to get shapes and (some) rest-frame colors (see #250).  @kaylanb has more recent fancy ones we might incorporate.
* Stellar spectra (including white dwarfs) are still not being interpolated based on the physical values of effective temperature, etc., in the Galaxia mocks (documented in #251 and being tracked in https://github.com/desihub/desisim/issues/315). 
* Contaminants exist but their densities have not yet been tuned.  This issue will require some thought and care.
* I totally and utterly broke the "mock catalogs without spectra" infrastructure used by `quicksurvey` (I'm sorry, @forero!) but I have a plan for putting it back together.
* Milky Way dust extinction is still not being handled correctly and consistently. In particular, the "truth" spectra are not being reddened, but they should be...
* Only the new-format (CoLoRe) Lya skewers are supported (#295) but maybe that's OK.  (Also note that the CoLoRe mocks are not yet full-resolution and they don't extend to z=1.6; see https://github.com/desihub/desisim/issues/295).
* Although the input `seed` provides reproducibility, it's non-trivial to regenerate the spectra from the output `truth` catalogs, particularly for the QSOs.  This is probably OK but medium-term we should probably implement something like what was proposed in https://github.com/desihub/desisim/issues/290.

Additional missing features:
* Does not yet support the buzzard mocks (but trivial to add) - #249.
* Top-level verbose option does not yet propagate to `mockmaker`.
* Still no bright-star mask for mocks (#103).
* No DLAs yet in the QSO spectra (but should be trivial to add).


forero:
@moustakas I will work on this review tomorrow.

moustakas:
Thanks @forero.  Please forgive me for the violence I did to the "no_spectra" part of the code--everything was temporarily moved to `desitarget.mock.quicksurvey` in a non-functioning state--because I wanted the refactored code clean, streamlined, and readable, but it was done with @sbailey's blessings!  I have every intention of restoring the "no_spectra" functionality as soon as possible.

geordie666:
@moustakas: As an initial check, I ran the command-line code on an interactive node. Your instructions worked perfectly for me in terms of executing the code.

However, I ran `select_mock_targets` 4 times (always creating HP 6507), and twice it failed with a memory error. This occurred at distinct parts of the code. Once was generating `skies`, once was generating `WDs`, e.g.:

```
INFO:build.py:109:_density_fluctuations: Dividing each nside=32 healpixel (3.36 deg2) into 16 nside=128 healpixel(s) (0.21 deg2).
INFO:build.py:364:targets_truth: Goal: generate spectra for 32 WD targets (9.53 / deg2).
WARNING:build.py:234:get_spectra_onepixel: Too few candidate targets (1) than desired (2).
WARNING:build.py:234:get_spectra_onepixel: Too few candidate targets (1) than desired (2).
WARNING:build.py:234:get_spectra_onepixel: Too few candidate targets (1) than desired (2).
DEBUG:build.py:269:get_spectra_onepixel: Selected 2 / 2 targets on chunk 1 / 2
salloc: Exceeded job memory limit
``

But, `select_mock_targets` ran to completion the other 2 times I tried it, and I recovered your exact QA pages:

http://portal.nersc.gov/project/desi/users/adamyers/garbage/

So, this could have been nothing (perhaps me futzing around too much on the interactive nodes), or it could be that memory can be exceeded when generating certain targets on certain nodes.

If I had to guess, the code seemed to run just fine every time I allocated a _new_ node, but not if I executed twice on the _same_ node (after deleting the output fits files and the `65` directory). So, perhaps the code is prone to memory leaks?

Anyway, as I couldn't easily recreate the memory fail when I obtained a _fresh_ node, this seems to work OK. I'll review mock provenance in how you create `TARGETID` later today, and then I'll leave the bulk of the review, particularly of aspects on the mocks side, to @forero.

**Comment by sbailey** [comment]
Notes / bug reports while trying it out in the two_percent_DESI minitest notebook:

* mockmaker.py line 2093: mXXL -> MXXL to match directories at NERSC
* logfiles included "Message: 'Galaxia top-level directory /global/project/projectdirs/desi/mocks/mws/galaxia/alpha/0.0.5_superfaint/healpix/None not found!'"
* logs of log message failures as reported in desihub/desiutil#98, but I think that is independent of this PR itself.

**Comment by moustakas** [comment]
Another issue that should be fixed before merging:  the Lya QSO photometry needs to be computed *after* the Lya forest is inserted (and before target selection).  Right now this is being done *before* (i.e., in `desisim.templates.SIMQSO`).

File: py/desitarget/mock/build.py [hunk 1]
```diff
@@ -1495,52 +446,98 @@ def finish_catalog(targets, truth, skytargets, skytruth, nside, healpix_id, seed
...
+        if nsky > 0:
+            skytargetid = encode_targetid(objid=skytargets['BRICK_OBJID'], brickid=skytargets['BRICKID'], mock=1, sky=1)
+            skytargets['TARGETID'][:] = skytargetid
+            skytargets['SUBPRIORITY'][:] = subpriority[nobj:]
+
+            skytargets['HPXPIXEL'][:] = radec2pix(nside, skytargets['RA'], skytargets['DEC'])
+                
+    else:
+        objid = np.arange(nobj + nsky)
+        targetid = encode_targetid(objid=objid, brickid=healpix, mock=1)
```
geordie666: Although your uses of `encode_targetid` are formally correct (thanks for making this switch!) this line is potentially mildly dangerous. `brickid` is set to a maximum of 20 bits, which will be exceeded at `nside = 512` (but is safe for `nside < 512`). I doubt that we'll ever use `nside=512` and `encode_targetid` is smart enough to flag an error if the bit-allowance of 20 bits for `brickid` is exceeded, but you might want to capture an error for the `nside > 256` case, here?

forero:
@moustakas I ran the commands on the command-line and everything worked as advertised. The QA plots look the same http://portal.nersc.gov/project/cosmo/temp/forero/refactor/. I tried it three times and didn't get any error messages.

py/desitarget/mock/build.py
```diff
@@ -4,1488 +4,439 @@
...
 
-        del source_data # memory clean-up
+    if rand is None:
+        rand = np.random.RandomState()
 
-        # Unpack the results removing any possible bricks without targets.
-        out = list(zip(*out))
+    # Fluctuations model coefficients from --
+    #   https://github.com/desihub/desitarget/blob/master/doc/nb/target-fluctuations.ipynb
+    model = dict()
```
forero: What happens for BGS?

moustakas: BGS has density fluctuations built into it, because of the sophistication of the MXXL mock.

py/desitarget/mock/build.py
```diff
@@ -4,1488 +4,439 @@
...
+
+    # Assign sources to healpix chunks.
+    healpix_chunk = radec2pix(nside_chunk, data['RA'], data['DEC'])
+
+    indxperchunk, ntargetperchunk = list(), list()
+    for pixchunk in set(healpix_chunk):
+        indx = np.where( np.in1d(healpix_chunk, pixchunk)*1 )[0]
+        indxperchunk.append(indx)
+
+        if coeff:
```
forero: How are Poisson + LSS fluctuations taken into account? Aren't those erased by setting each pixel to a fixed number density?

py/desitarget/mock/build.py
forero: Also due to imaging?

forero: I thought that the density fluctuations model was dependent on the DECaLS results. Are those also included in MXXL?

sbailey:
Memory problems that @geordie666 reported: this might be related to desihub/desisim#319, where NERSC changed how slurm monitors memory and completely broke it for python multiprocessing.  That was fixed as of last night, so please re-test and re-report if you continue to have memory problems with this branch.

geordie666:
I re-tested this 3 times on the same node and could *not* recreate the memory problem. So either this was fixed with the slurm update, or it wasn't a problem in the first place.

moustakas:
@sbailey Can you point me to the yaml configuration file you used when you ran the minitest notebook and reported [this comment](https://github.com/desihub/desitarget/pull/264#issuecomment-360360780)?  Regarding your second bullet I think the issue is that you did not define `nside_galaxia` in the parameter file.

An updated parameter file can be found in `/global/cscratch1/sd/ioannis/refactor/refactor.yaml`

py/desitarget/mock/build.py
moustakas: Thanks @geordie666 I've added a ValueError exception if nside > 256 at the start of `mock.build.targets_truth`.

py/desitarget/mock/build.py
moustakas: @forero Thank you for this and the previous question (on BGS) regarding how density fluctuations are being handled in this refactor. First, let me say that when I sat down to refactor the code from the brick-based model to a healpixel-based model (which utilized @geordie666's more recent "summary" statistics of the imaging data in #217) I didn't fully understand the fluctuation model that you, Adam, and Stephen had implemented.  However, after speaking with Stephen off-list, I now have a better idea of how you were down-sampling the various mocks to the desired target density while still preserving the imprint of large-scale structure.  

In this refactor I am definitely *not* doing the down-sampling correctly.  However, I'd like to propose -- and I believe that @sbailey agreed -- that we should get this merged in and I'll work on "fixing" the fluctuations model.

Regarding accounting for the imaging systematics:  developing a proper model is something the Clustering WG is (should be?) already doing.  I will reach out to them to see if someone from that group can contribute code so our simulated target densities are as realistic as possible.

sbailey:
For testing this, I was using the configuration file in the [two_percent_DESI minitest](https://github.com/desihub/two_percent_DESI/blob/master/mini/desisurvey-config.yaml), which indeed does not have `nside_galaxia` defined.  Please update `doc/mock_example/input.yaml` to be a working example mock configuration file. 

Agreed about merging despite the incorrect density fluctuation model (despite the name of the branch...), as long as fixing that model is high priority and fixable in the next 1-2 weeks.

moustakas:
@sbailey Your link pointed to the `desisurvey-config.yaml` file.  The [select-mock-targets.yaml](https://github.com/desihub/two_percent_DESI/blob/master/mini/select-mock-targets.yaml) configuration file for the minitest is *significantly* out of date.  

(Although what is the `/project/projectdirs/desi/users/sjbailey/mocks/sky/0.1/newsky.fits` file?  I wasn't familiar with the new sky mock.)

Either way, I'll update `doc/mock_example/input.yaml`.

py/desitarget/mock/build.py
forero: @moustakas I agree. This is good enough to merge now and implement a fix later.

sbailey:
/project/projectdirs/desi/users/sjbailey/mocks/sky/0.1/newsky.fits is from desitarget PR #219 to make skies that guarantee two targets per nside=2048 pixel, instead of a poisson average of 2 targets per pixel (and thus often having no skies available for a given fiber due to fluctuations).  This reflects the "more uniform than random" densities that we want for real sky targets so that every fiber always has a sky (or badsky) position as a backup option in case there are no science targets.

I moved a copy into /project/projectdirs/desi/mocks/uniformsky/0.1/uniformsky-2048-0.1.fits as a better final location to use instead of something under users/sjbailey/mocks.

moustakas:
This is nearly ready to merge, although I have not yet tested at NERSC.  On my laptop (with 4 cores) generating roughly 23k targets (of all types) in a single `nside=32` healpixel took `14m42s`, or roughly `65s/deg^2/cpu`.

moustakas:
OK, I think all conflicts have been fixed -- thanks for the off-list help, @geordie666!  

@sbailey once your tests have completed (and hopefully Travis tests pass and the documentation builds successfully!) please feel free to merge, otherwise post any crashes or bugs here or off-list.

Issue #274, opened 2018-02-13
Guide/Focus/Alignment targets


forero:
Carrying this issue over from `fiberassign` https://github.com/desihub/fiberassign/issues/94.

We need to dump the list of targets that are within reach of the GFA sensors for each tile.

forero:
Solved in #275

PR #275, opened 2018-02-13
select gfa targets and merge them into fiberassign results

forero:
Solves #274 

It can be tested using
```bash
select_gfa -o $SCRATCH/gfa_test/ -s /project/projectdirs/cosmo/data/legacysurvey/dr5/sweep/5.0/ -f /project/projectdirs/desi/datachallenge/dr5fiberassign/single_epoch/fiberassign/tileresults -t gfa_tiles_mini.txt
```
where gfa_tiles_mini.txt can be the following https://github.com/forero/SandboxDESI/blob/master/gfa/gfa_tiles_mini.txt

### DIFF
--- a/bin/select_gfa
+++ b/bin/select_gfa
@@ -8,8 +8,18 @@ import desimodel.io
 import desimodel.footprint
 import matplotlib.pyplot as plt
 import glob
+import time
 import os.path
 import argparse
+
+def near_tile(data, tilera, tiledec, window_ra=4.0, window_dec=4.0):
+    delta_RA = data['RA'] - tilera
+    delta_dec = data['DEC'] - tiledec
+    jj = np.fabs(delta_RA) < window_ra
+    jj = jj | ((delta_RA + 360.0) < window_ra)
+    jj = jj | ((360.0 - delta_RA) < window_ra)
+    jj = jj & (np.fabs(delta_dec)<window_dec)
+    return jj
 
 parser = argparse.ArgumentParser()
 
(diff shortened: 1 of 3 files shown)


geordie666:
Looks like @sbailey fixed 2 of my critical requests in his changes, so I'm happy to merge.

sbailey:
Thanks.  We'll need to define exactly which sweep columns get propagated into the GFA output table, but this is a good starting point.  I like that you included the ability to add a GFA HDU to the fiberassign tile*.fits files.  I made a few small changes:
* added `GFA_LOC` column to the output table so that you can tell which GFA each target is on.
* added EXTNAME=GFA in the gfa_targets*.fits files like you did in updated fiberassign files.
* made `--input_fiberassign` and `--input_tiles` optional

I also made a few attempts at making it faster and less memory hungry but without a lot of success.  We'll need to parallelize and/or optimize for this to be viable for running on all tiles for the full footprint.  @geordie666's idea of doing a mag cut while loading the sweeps is good (I was trying to do the geometry cut there, but that bogs down).  We should keep extended sources too (not just point sources).  Even if they aren't used for guiding and seeing measurements, we want to know that they are there on the GFAs.

While I was typing this @geordie666 submitted some comments, some of which are already addressed, so make sure to `git pull` before doing any further updates.

sbailey:
I went ahead and added the output_dir pre-check and made a mag<18 cut that makes things way faster.  We may still want to parallelize but this gets into the realm of viable to run on the full footprint.  I agree with @geordie666's other refactor comments, but OK to merge when Travis tests pass and save those for later.

forero:
@sbailey thanks for all the changes in the code. 
@geordie666 I will move the key functions into a gfa.py module and add the DESI logger.
Concerning speed, I spent a long time trying to speeding it up and the best result  I got was loading the whole thing up and then selecting small windows around the current tile.

bin/select_gfa
```diff
@@ -0,0 +1,146 @@
...
+sweep_data = []
+# n_sweep = 10
+for i in range(n_sweep):
+    sweep_file = sweep_files[i]
+    data = fitsio.read(sweep_file, columns=['RA', 'DEC', 'FLUX_R'])
+
+    #- Keep just mag>18
+    rfluxlim = 10**(0.4*(22.5-18))
+    ii = data['FLUX_R'] > rfluxlim
+    data = data[ii]
```
moustakas: This cut can be done on-the-fly using fitsio's `where` method, without having to read in the full sweep, which should save time and tons of memory.

bin/select_gfa
```diff
@@ -0,0 +1,146 @@
...
+print('There are {:.2f}M targets near tiles in the sweeps'.format(len(all_sweep)/1E6))
+
+#find IDs of targets on every individual tile
+
+for i in range(len(desi_tiles)):
+    tile_id = desi_tiles['TILEID'][i]
+    print('computing TILEID {:05d} on RA {:6.2f} DEC {:6.2f}'.format(tile_id, desi_tiles['RA'][i], desi_tiles['DEC'][i]))
+    
+    # select targets in a smaller window centered on tile
+    jj = near_tile(all_sweep, desi_tiles['RA'][i], desi_tiles['DEC'][i])
```
moustakas: I don't know if this comment belongs here or further up, but since we know the central coordinates and boundaries of all the sweeps (they're simple RA/Dec rectangles) can't we read the sweep file we need on-the-fly based on the tile we're considering?  I think this would scale much better than the current implementation.  Then the RA/Dec and magnitude cuts could be applied again using fitsio's `where` method, without reading in the whole file.

Another possibility (not for this PR, obviously) which we've talked about would be to repackage the sweeps into healpixels (for DESI purposes, not as part of the imaging survey DRs) and pre-compute the tiles that touch each sweep.  That would then make these kinds of target searches significantly faster.

bin/select_gfa
```diff
@@ -0,0 +1,146 @@
...
+import desimodel.io
+import desimodel.footprint
+import matplotlib.pyplot as plt
+import glob
+import time
+import os.path
+import argparse
+
+def near_tile(data, tilera, tiledec, window_ra=4.0, window_dec=4.0):
+    delta_RA = data['RA'] - tilera
```
moustakas: I would test this script against the `astropy.coordinates.SkyCoord` methods, which I think might be significantly faster (or maybe not!).

bin/select_gfa
```diff
@@ -0,0 +1,146 @@
...
+            a = fitsio.write(filename, gfa_targets, extname='GFA', clobber=True)
+            
+
+#rewrite a new tilefile with all the info in three HDUs
+gfa_files = glob.glob(os.path.join(args.output_dir, "gfa_targets_*.fits"))
+gfa_tile_id = {}
+for gfa_file in gfa_files:
+    f = gfa_file.split('/')[-1]
+    fileid = f.split("_")[-1]
+    fileid = fileid[0:5]
```
moustakas: These splits on `/` and `_` strike me as fragile.  Since you know the tile IDs you can just reconstruct the filenames.  Then you could add a check to make sure all the files you were expecting were actually written out.

forero: We know the tile IDs we want, but we don't know the tiles that were actually computed. That's why we have to go through the GFA files that were actually written to disk.

bin/select_gfa
forero: Memory is not a problem. We are using only 1GB of RAM. Time is not too bad. It takes 2 minutes to read all the sweeps.

bin/select_gfa
forero: Those are good ideas, but I think we are not struggling for memory or time. We only use 1GB of RAM and processing a single takes ~0.5 seconds.

bin/select_gfa
forero: I don't understand this. What outputs should be tested against what functions in `astropy.coordinates`? We are not doing coordinate transformations in the script.

forero:
Simple test. In orange the results from fiberassign, in blue the results from the gfa script.
gfatest

forero:
All test pass. I am merging now.

sbailey:
For the record, a few comments on the comments:

* using `fitsio.read` option `where` to trim while reading: as @forero noted memory isn't a big issue since each file is trimmed immediately after reading in to the total memory footprint remains small.  This feature of fitsio is still a neat trick to keep in mind, and worth trying at some point.
* reading in a subset of sweeps: that would speed things up when processing a small number of test tiles, but when doing the full footprint of tiles we have to read all sweep files anyway.  Several of my optimization attempts made the test case fast but the full case slower, so keep that in mind.  It would be nice to be able to do a small number of tiles without reading in every sweep file, but the current state is still useful.
* `near_tile` vs. `astropy.coordinates.SkyCoord`: I had pulled the logic into a separate function so that I could test against `desimodel.footprint.find_points_in_tiles`, which turned out to be much slower.  I didn't test against `astropy.coordinates.SkyCoord` because my previous experience is that code is pretty slow.  A key difference is that `near_tiles` is doing a very approximate rectangular cut (good enough for there) while the other codes are doing circular cuts that require more logic.

PR #278, opened 2018-02-23
open BGS mock read-only; fixes parallel crash

sbailey:
This PR fixes a bug discovered while testing a candidate for the desi software 18.2 release.  Previously the BGS mock file in hdf5 format was being opened in read-write mode instead of read-only mode.  Starting with h5py 2.7.1 this attempts to create a writelock and parallel processes from `mpi_select_mock_targets` were colliding with each other and crashing.  This PR updates the BGS mock reader to open the file read-only, which fixes the problem.  It also augments the logging to make it easier to debug any other crashes in the future.

This may create a merge conflict with open PR #264; sorry @moustakas.  You'll want to make sure to make the same `h5py.File(filename, mode='r')` change there too.

I intend to merge as soon as Travis tests pass so that we can move on with testing 18.2.

### DIFF
--- a/bin/mpi_select_mock_targets
+++ b/bin/mpi_select_mock_targets
@@ -155,6 +155,13 @@ if len(rankpix) > 0:
         t0 = time.time()
         try:
             with stdouterr_redirected(to=logfile):
+                log.info('Calling targets_truth() with the following options')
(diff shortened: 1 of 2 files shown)


moustakas:
Thanks, these I'll be sure to incorporate these changes.

PR #282, opened 2018-02-27
Attempt to fix the failing read the docs build

geordie666:
This seems to work as a way to circumvent unnecessary imports in __init.py__ and so hopefully fix the read the docs build problems. We may not want to go this route, however because this:

- could break other imports in code outside of desitarget
- has a somewhat hack-y workaround to mock up a function if sphinx is present

Really, I'm just thinking out loud, so feel free to close this PR and disregard this approach if it seems inadequate.

### DIFF
--- a/py/desitarget/QA.py
+++ b/py/desitarget/QA.py
@@ -28,7 +28,7 @@ from . import __version__ as desitarget_version
 from desiutil import brick
 from desiutil.log import get_logger, DEBUG
 from desiutil.plots import init_sky, plot_sky_binned
-from desitarget import desi_mask
+from desitarget.targetmask import desi_mask
 
 import warnings, itertools
 import matplotlib
--- a/py/desitarget/contammask.py
+++ b/py/desitarget/contammask.py
@@ -10,6 +10,10 @@ except ImportError:
     # This can happen during documentation builds.
     #
     def BitMask(*args): return None
+#ADM this is a way to force different behavior during sphinx builds
+import sys
+if 'sphinx' in sys.modules:
+    def BitMask(*args): return None
 
 _bitdefs = None
 def _load_bits():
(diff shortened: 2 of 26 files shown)


py/desitarget/QA.py
```diff
@@ -28,7 +28,7 @@
 from desiutil import brick
 from desiutil.log import get_logger, DEBUG
 from desiutil.plots import init_sky, plot_sky_binned
-from desitarget import desi_mask
+from desitarget.targetmask import desi_mask
```
weaverba137: Is there a particular reason why you're doing
```python
from desitarget.targetmask import desi_mask
```
instead of 
```python
from .targetmask import desi_mask
```

py/desitarget/contammask.py
```diff
@@ -10,6 +10,10 @@
     # This can happen during documentation builds.
     #
     def BitMask(*args): return None
+#ADM this is a way to force different behavior during sphinx builds
```
weaverba137: Why is `BitMask` defined in both contammask and targetmask?  The *code* should exist in one and only one place, and then the *configuration* of the masks can be in different places.

moustakas: I probably inadvertently duplicated this code when working on the mock target contaminants.  @geordie666 if you could fix this I'd be grateful.

weaverba137:
I'm fixing up the doc/conf.py and doc/api.rst files.  I've also moved `gitversion()` from `__init__.py` to `io.py`, because it is only ever used in `io.py`.

weaverba137:
Documentation should now build successfully.  There is still one warning because docs.h5py.org is down.

weaverba137:
I think this is ready to go now.

geordie666:
Thanks for the rigorous clean-up, @weaverba137. Everything works for me as usual, and sphinx seems to build correctly too, so I'm going to merge this in the next hour or so.

PR #288, opened 2018-03-01
include OBSCONDITIONS in std/sky output

sbailey:
This PR is a minimal hack to fix the end-to-end integration test.  We need a deeper refactor of how targets/standards/sky interact with MTL and fiberassign, but this is gets us back into a state where the select_mock_targets output can be used by fiberassign.

## Details ##

PR #264 removed OBSCONDITIONS as a column of the targets file.  Good - the target bits determine the conditions under which a target can be observed and that info gets added in along with the priorities by the merged target list (MTL) code prior to handing off to fiberassign.

The problem is that fiberassign requires the standards and sky targets to have the same format as the MTL file, and thus include OBSCONDITIONS, even though standards and sky don't go via MTL.  Thus this PR restores OBSCONDITIONS to the mock standards and sky output files, while still leaving it out of the science targets file.

I suggest that we merge this and tag so that we can get back into a state of a working end-to-end test, and then more carefully review what the long term right solution is, which will likely involve updates to both desitarget and fiberassign.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,10 +5,12 @@ desitarget Change Log
 0.19.1 (unreleased)
 -------------------
 
-* Fix bug whereby FLUX and WAVE weren't being written to truth.fits files [`PR
-  #287`_].
+* Fix bug whereby FLUX and WAVE weren't being written to truth.fits files
+  [`PR #287`_].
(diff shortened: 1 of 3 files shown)


moustakas:
I took this opportunity to completely remove references to `obsconditions` in `mock.mockmaker`, which had been commented out.

PR #290, opened 2018-03-06
Greatly speed up the module that assigns sky positions (while avoiding targets)

geordie666:
This is still not the expected final algorithm, which will instead piggyback on:

https://github.com/desihub/desitarget/blob/master/py/desitarget/skyfibers.py

but this should be good enough to begin experimenting with assigning fibers to sky positions downstream of `desitarget`. This update includes:
   * Use the formalism of https://github.com/desihub/desitarget/blob/master/py/desitarget/brightmask.py to greatly increase the speed with which sky positions are compared to masks.
   * Pass around a magnitude limit on masks from the sweeps (to better mask only objects that are genuinely detected in the sweeps).
   * Reduce the default margin to produce ~1700 sky positions per sq. deg...the previous version generated tens-of-thousands of positions per sq. deg., which was overkill.

The default code runs in under an hour on an interactive node, e.g.:

```
salloc -N 1 -C haswell -t 00:60:00 --qos interactive -L SCRATCH,project
select_targets /global/project/projectdirs/cosmo/data/legacysurvey/dr5/sweep/ $CSCRATCH/test-skies-margin-dr5.fits --skies
```

Here's an example of the results, where blue areas are masks (typically faint-ish galaxies from the sweeps):

skies

### DIFF
--- a/bin/select_targets
+++ b/bin/select_targets
@@ -27,7 +27,7 @@ ap.add_argument("src", help="Tractor/sweeps file or root directory with tractor/
 ap.add_argument("dest", help="Output target selection file")
 ap.add_argument('-s', "--skies", action='store_true',help="Generate blank sky positions in addition to targets")
 ap.add_argument("--skymaglim",
-                help='Magnitude limits for building avoidance zones from objects in the sweeps (e.g. 20,20,20)',
+                help='Faint (g,r,z) magnitude limits for masks (for building avoidance zones from objects in the sweeps; e.g. "20,20,20")',
                 default="20,20,20")
 ap.add_argument('-c', "--check", action='store_true',help="Process tractor/sweeps files to check for corruption, without running full target selection")
 ap.add_argument('-m', "--mask", help="If sent then mask the targets, the name of the mask file should be supplied")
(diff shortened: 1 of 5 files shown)


geordie666:
Ooops...sphinx isn't happy but I have to step away for a few hours. Nobody merge this until I fix the docs.

sbailey:
The example file `/project/projectdirs/desi/target/catalogs/test-skies-mag-g23-r22.5-z22-dr5.fits` has sky location TARGETID 598451974563697958 twice, which is crashing fiberassign downstream.  Please check TARGETID logic.

sbailey:
TARGETID 598452076808324993 is another example; that one is a BADSKY.  There are 518 duplicated TARGETIDs in the file.

geordie666:
Good catch. This was caused by the fact that bricks can span sweeps files (different sweeps files can contain objects with the same `BRICKID`. Consequently, when parallelizing across sweeps, the same `BRICKID` could be assigned the same `OBJID`.

I believe I've fixed this bug by insisting that `TARGETIDs` are unique *after* the results for individual sweeps have been gathered into one rec array...but I'll run a new version of the example file tonight and check that the `TARGETIDs` are, indeed unique before requesting that this PR is merged.

sbailey:
Other than the duplicate TARGETIDs, the example file worked fine for fiberassign, so go ahead and merge when ready after double checking the new code on the full dataset doesn't have duplicate TARGETIDs.  Further checks and improvements can come via other issues and PRs.

PR #294, opened 2018-03-23
Retune DR6 densities

geordie666:
This PR retunes DR6 densities by:

- tweaking the QSO random forest probability.
- adding a new ELG selection (see the first slide of https://desi.lbl.gov/trac/attachment/wiki/TargetSelectionWG/TSTeleconMinutes/DESITS180322/Strategy_ELG.pdf).
- incorporating slight color-and-flux shifts from https://desi.lbl.gov/DocDB/cgi-bin/private/RetrieveFile?docid=3390;filename=Raichoor_DESI_05Dec2017.pdf;version=1 .
- incorporating initial functionality to facilitate different selections in the "North" (MzLS/BASS) vs. the "South" (DECaLS).

Other updates and general clean-ups that were included:

- introducing a `NO_TARGET` bit that can be set downstream in the event that a fiber can't find a target.
- renaming the `BADSKY` bit `BAD_SKY` for consistency with other bits.
- including `FRACDEV` and `FRACDEV_IVAR` in the targeting outputs.
- slightly expanding the magnitude plot limits in QA.


geordie666:
BTW, @sbailey, I chose 49, in the end for the `NO_TARGET` bit, instead of 37, as I didn't want to lump it in with the calibration bits, and I can see the calibration bits expanding (perhaps `STD_WD_NORTH`, `STD_WD_SOUTH`, etc.).

sbailey:
A few comments for the record before merging:

This sets XYZ_NORTH bits even for targets with south-only photometry and vice-a-versa.  This is purposeful and I think good for systematic studies, but is something to keep in mind.  The joined set is resolved by the non-NORTH/SOUTH bits, e.g.
```
elg = (elg_north & photsys_north) | (elg_south & photsys_south)
```

However, more work will need to be done to handle data in the overlap regions so that we don't pass the same physical object twice to fiberassign.  See #295.

This `isXYZ_north` vs. `isXYZ_south` code appears fragile to maintain due to how similar they are, but each function is still different enough that there isn't an obvious solution that is more maintainable.  Let's merge this for now, but if we find ourselves making mistakes we could consider alternatives.

moustakas:
These changes look great. My only comment is that we be a little careful that we're using consistent color transformations. For example, I think there are at least 3-4 different transformations floating around, and we should be sure to adopt whichever ones make it into the imaging paper(s) (which I believe Arjun is currently drafting).

geordie666:
Thanks, @moustakas. Note that I added convenience functions (`isELG`) etc. that just call the "southern" (DECaLS-appropriate) cuts in an attempt to not break the code in the mocks.

PR #309, opened 2018-04-27
New qso rf strategy

CAClaveau:
Added two new QSO RF trained over DR5 and updated 'isQSO_randomforest_north/south' functions


geordie666:
Thanks @CAClaveau:

The code runs fine, although it is slowed down slightly by the new random forests. Thanks for keeping those combined RFs to less than 10MB (in total). They're still rather large for what is essentially code, but 10MB is just about manageable provided we only update the RFs once or twice more before freezing the `desitarget` code prior to the DESI survey.

Output QA plots (that are not weighted by the DR5 and DR6 imaging footprints, which I haven't constructed weights for, yet) are here:

http://portal.nersc.gov/project/desi/users/adamyers/desitargetQA-dr5.0-NEW_QSO_RF_STRATEGY/QSO.html
http://portal.nersc.gov/project/desi/users/adamyers/desitargetQA-dr6.0-NEW_QSO_RF_STRATEGY/QSO.html

If @Cyeche could review these QA plots and agree that they look OK, then I'm happy to merge this PR.

One note: The new code seems to include random forests for both a high-z sample of quasars and for a regular sample of quasars. This potentially makes modeling large-scale structure in the quasar sample even more difficult than for a single RF. Such a choice should probably be discussed by the target selection and clustering working groups.

Cyeche:
Hi,

I have a problem with the footprint which is not the same. It is particularly visible with DR6. The edge of the footprint are not the same if we compare the two QA plots

http://portal.nersc.gov/project/desi/users/adamyers/desitargetQA-dr6.0-NEW_QSO_RF_STRATEGY/QSO.html <http://portal.nersc.gov/project/desi/users/adamyers/desitargetQA-dr6.0-NEW_QSO_RF_STRATEGY/QSO.html>
http://portal.nersc.gov/project/desi/users/adamyers/desitargetQA-dr6-0.20.0/QSO.html

We should understand if it is due to the fact that we relax a cut such as (g>0 & r>0& ….) in the QSO selection or that the definition of the footprint has changed. 

Note that we have something similar with DR6 but less pronounced.

geordie666:
@Cyeche: As I mentioned, there's absolutely no weighting in these plots ("Output QA plots that are not weighted by the DR5 and DR6 imaging footprints, which I haven't constructed weights for, yet").

So, you shouldn't expect these "unofficial" QA plots to give you the correct densities, particularly at the edges of the footprint. You should just check that the overall distribution in color space, across the middle of the footprint, etc. is what you'd expect at the level that there are no obvious bugs in the new code.

Note that the plots at http://portal.nersc.gov/project/desi/users/adamyers/desitargetQA-dr6-0.20.0/QSO.html were cut to the edge of the expected DESI _spectroscopic_ footprint, so will have different densities at the edges when compared to the new plots.

Cyeche:
> 
> 
> Note that the plots at http://portal.nersc.gov/project/desi/users/adamyers/desitargetQA-dr6-0.20.0/QSO.html <http://portal.nersc.gov/project/desi/users/adamyers/desitargetQA-dr6-0.20.0/QSO.html> were cut to the edge of the expected DESI spectroscopic footprint, so will have different densities at the edges when compared to the new plots.

Adam your are right,  that’s the reason of the difference. Therefore it looks fine! That’s ok with me.

 Christophe.

Issue #310, opened 2018-04-27
run_target_qa crash


sbailey:
`run_target_qa` crashes if not provided a weightmapfile due to:
```
Traceback (most recent call last):
  File "/global/cscratch1/sd/sjbailey/desi/code/desitarget/bin/run_target_qa", line 33, in <module>
    make_qa_page(ns.src,qadir=ns.dest,mocks=ns.mocks,makeplots=makeplots,max_bin_area=max_bin_area,imaging_map_file=ns.weightmapfile)
  File "/global/cscratch1/sd/sjbailey/desi/code/desitarget/py/desitarget/QA.py", line 2569, in make_qa_page
    weight=weight,imaging_map_file= imaging_map_file)
  File "/global/cscratch1/sd/sjbailey/desi/code/desitarget/py/desitarget/QA.py", line 2330, in make_qa_plots
    totalpixweight = np.sum(pixweight[uniqpixset])
UnboundLocalError: local variable 'pixweight' referenced before assignment
```

`pixweight` is only defined if the weightmap file is loaded.  This is breaking the end-to-end integration test on mocks which don't have a weightmap file.

PR #311, opened 2018-04-27
Fixes a bug where QA code crashed if a pixel weights map was not passed

geordie666:
I've verified that after this PR the QA code runs to completion without a weights map.

This should fix https://github.com/desihub/desitarget/issues/310.

This probably highlights the need for unit tests for QA, which I'd previously resisted as QA is mostly a lot of plot creation. I'll need to find some time to add a few unit tests at a later date.

@sbailey: I'm out of commission for the next hour, but please review this and merge if it passes unit tests. If you think it's needed for a relatively minor update, then feel free to add a line to `changes.rst` before merging.


sbailey:
Tested and confirmed.  Thanks.

PR #322, opened 2018-06-01
Create monolithic file of GFAs via matching to Gaia

geordie666:
Adds a module `gaiamatch.py` that has utilities for making positional matches to the Legacy Surveys Gaia "chunk" files at NERSC. This module can:

- perform one-to-one matches between Gaia and the sweeps
- retain all Gaia objects in an RA/Dec box as well as objects that match between Gaia and the sweeps

Uses this module in `gfa.py`to loop through the sweeps and:

- assemble Gaia objects (to some magnitude limit in Gaia) 
- add matching fluxes from the sweeps, which should be useful for ETC/throughput calibration

The following runs in about 5 minutes on an interactive node on Cori (using 32 processors):

```
select_gfas /global/project/projectdirs/cosmo/data/legacysurvey/dr6/sweep/6.0/ $CSCRATCH/temp.fits
```

And creates ~18 million possible GFA targets. An example output file is here:

```
/global/cscratch1/sd/adamyers/gfas-dr6.fits
```


geordie666:
Will self-merge in ~12 hours so I can rebase some work-in-progress that also uses Gaia-matching. I'm sure @sbailey will have refinements when he's back online, but I can deal with those in a separate branch.

PR #324, opened 2018-06-11
Adds MWS selection to desitarget

geordie666:
This PR adds the capability to set the MWS bits from catalog-matching to Gaia.

Ultimately we will derive Gaia directly from the Legacy Survey sweeps files, which will require a minor refactor. But this puts all of the selection functions and mechanisms in place, which facilitates MWS selections prior to DR7 (and DR8 for BASS/MzLS) of the Legacy Surveys.


moustakas:
A brief comment:  this PR is a great development but it will almost certainly break the mock / data challenge code, as none of the stellar mocks include the necessary Gaia data...

geordie666:
Probably not...I was conscious of backwards-compatibility throughout. The existing function in `cuts.py` for MWS selection is still there, and any code that matches to Gaia has an option to turn off Gaia-matching. You may acquire some extra columns in `io`, but you don't have to use them.

I've no doubt missed _something_ that will be critical to the mocks, of course.

PR #325, opened 2018-06-12
allow environment variables in config file when running select_mock_targets

moustakas:
This PR adds some flexibility by allowing environment variables to be used in the `select_mock_targets` configuration file (a trick shamelessly stolen from `desisurvey.config` -- thanks @dkirkby!) 

So, for example, whereas the `config.yaml` file previously had hard-coded paths, e.g.,
```
dust_dir: /Users/ioannis/research/data/sfd_dustmaps/maps
sources:
    BGS: {
        target_name: BGS,
        mockfile: /Users/ioannis/work/desi/mocks/bgs/MXXL/desi_footprint/v0.0.4/BGS.hdf5,
        format: durham_mxxl_hdf5,
        magcut: 20.2,
    }
```
making it a pain to use the same configuration file on a laptop (for development) and NERSC (for production) without hand-editing, the code will now accept a config file like this:
```
dust_dir: '{DUST_DIR}/maps'
sources:
    BGS: {
        target_name: BGS,
        mockfile: '{DESI_ROOT}/mocks/bgs/MXXL/desi_footprint/v0.0.4/BGS.hdf5',
        format: durham_mxxl_hdf5,
        magcut: 20.2,
    }
```


sbailey:
+1 for adding environment variables to the config file parsing.  Some thoughts to consider, which could be done after merging or discarded:

* `DUST_DIR` is pretty generically named, but is tied to SFD dust maps specifically.  Consider a more specific name like `SFD_DUST_DIR` (albeit weighted against backwards compatibility issues considerations; people have been using "DUST_DIR" for this specific set of dust maps for many years...)
* consider basing the mock files from `$MOCK_ROOT` instead of `$DESI_ROOT`.  This would enable syncing mocks to scratch and setting just `$MOCK_ROOT` without having to switch all of `$DESI_ROOT` when using the default file.  The downside is that we have a new variable to set which is kinda yucky.
  * note: one could always do this for a separate config file so it isn't a big deal; it's just a matter of whether the default example config file should use `$DESI_ROOT` or `$MOCK_ROOT`.

I don't feel strongly about either of these, so consider them but its fine if you don't like them and want to proceed as-is.

moustakas:
These are all great suggestions but they add some complications to what I'm focusing on right now (the SV data challenge) and so I don't want to get side-tracked.  Let's revisit for when we address https://github.com/desihub/desispec/issues/513.

Also:  note that for the imaging data releases we've stopped copying everything to scratch before processing.  I don't know if it's because the I/O to /project has improved, or whether it wasn't worth the effort, but it may be something to keep in mind in terms of (not) copying the mock catalogs to $SCRATCH.

PR #327, opened 2018-06-14
Update to code that makes monolothic GFA target file

geordie666:
This PR improves the Gaia datamodel, particularly for GFA targets:

* Outputs columns names that are expected downstream.
* Aligns data model with what the Legacy Surveys will produce.


sbailey:
How do we distinguish object from GAIA that don't have a match from the Legacy Surveys or otherwise don't have observations in G/R/Z?  Is it this:
* FLUX_IVAR_G = FLUX_IVAR_R = FLUX_IVAR_Z = 0.0 means no match at all
* FLUX_IVAR_G/R/Z = 0.0 in some bands but not others means it was a match but missing some filters

?

Or should we also propagate NOBS_G/R/Z?

geordie666:
I think what you say about the `IVARS` is true for this particular PR, however, once the Legacy Surveys produce sources that will NOT be true for the `IVARS`. I agree that we should probably propagate `NOBS`, although it won't necessarily mean an observation has a high signal-to-noise of `NOBS > 0` so maybe you should just always apply criteria on `FLUX` and `IVAR` if the goal is to test the ETC?

sbailey:
Documenting call with @geordie666: will merge as-is for now, and use data quality cuts on IVAR downstream to decide if a FLUX should be used, regardless of whether it was an original Legacy Surveys detection or forced photometry.  @geordie666 will double check if/that IVAR=0.0 for cases where NOBS=0 in on band but it was detected in another band (e.g. at the edge of the footprint where we might have data in 2 out of 3 bands).

geordie666:
I checked, and there are corner cases in DR6 where sources have`NOBS = 0` but have measured fluxes. For example:

```
objs = fitsio.read("/global/project/projectdirs/cosmo/data/legacysurvey/dr6/sweep/6.0/sweep-160p030-170p035.fits")
w = np.where(  (objs["NOBS_G"] == 0) &  (objs["FLUX_IVAR_G"] != 0) )[0]
objs[w[0:5]]["NOBS_G"]
```
```
Out[69]: array([0, 0, 0, 0, 0], dtype=int16)
```
```
objs[w[0:5]]["FLUX_G"]
```
```
Out[70]: 
array([  3.31262970e+02,   2.76206270e+04,   1.72104311e+00,
         4.06871094e+02,   3.72734985e+02], dtype=float32)
```
```
objs[w[0:5]]["FLUX_IVAR_G"]
```
Out[71]: 
array([  3.75370798e-03,   5.12625775e-07,   1.35454731e+01,
         7.01126410e-03,   6.69344701e-03], dtype=float32)
```
An example in the viewer is here:

http://legacysurvey.org/viewer?ra=169.6806&dec=30.2631&zoom=16&layer=mzls+bass-dr6&sources-dr6

djschlegel:
NOBS is only evaluated at the central pixel coordinate, so I'd recommend against using it for target selection.

geordie666:
Good to know for sure, thanks @djschlegel

Issue #328, opened 2018-06-14
Add RA_IVAR and DEC_IVAR to GFA_TARGETS


sbailey:
Missed two columns on PR #327: we also want RA_IVAR and DEC_IVAR.

geordie666:
Will fix soon

sbailey:
fixed in PR #329

PR #329, opened 2018-06-14
Add IVARs to GFA output

geordie666:
Adds `RA_IVAR` and `DEC_IVAR` to the GFA Data Model.

New example output file at `/global/cscratch1/sd/adamyers/gfas-dr6.fits`

### DIFF
--- a/py/desitarget/gfa.py
+++ b/py/desitarget/gfa.py
@@ -29,7 +29,8 @@ start = time()
 #ADM the current data model for columns in the GFA files
 gfadatamodel = np.array([], dtype=[
     ('TARGETID', 'i8'),  ('BRICKID', 'i4'), ('BRICK_OBJID', 'i4'),  
-    ('RA', 'f8'), ('DEC', 'f8'), ('TYPE', 'S4'),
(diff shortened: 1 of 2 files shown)


geordie666:
Should close https://github.com/desihub/desitarget/issues/328 when merged.

py/desitarget/gfa.py
```diff
@@ -29,7 +29,8 @@
 #ADM the current data model for columns in the GFA files
 gfadatamodel = np.array([], dtype=[
     ('TARGETID', 'i8'),  ('BRICKID', 'i4'), ('BRICK_OBJID', 'i4'),  
-    ('RA', 'f8'), ('DEC', 'f8'), ('TYPE', 'S4'),
+    ('RA', 'f8'), ('DEC', 'f8'), ('RA_IVAR', 'f8'), ('DEC_IVAR', 'f8'),
```
sbailey: In the sweeps the RA/DEC IVARS are dtype f4 not f8.  Let's also keep them as f4 here.

Note: RA and DEC themselves are f8, which is needed.  This is only a comment on the IVARs.

Issue #336, opened 2018-07-30
SlaveException: column name 'GAIA_RA' not found (case insensitive)


moustakas:
@geordie666 I was trying to select a sample of ELGs from the COSMOS truth field and `select_targets` crashed.  Perhaps we shouldn't *require* the Gaia tags in the code?  

I also just noticed that there's an optional `gaiadir` input which assumes a NERSC infrastructure (i.e., not laptop-friendly).  If this directory isn't set does the code still proceed (again, under the theory that I don't---momentarily!---care about stars).

Here's a code snippet and the full traceback---

```python
from desitarget.cuts import select_targets
targ = select_targets('/global/projecta/projectdirs/desi/target/analysis/truth/dr7.0/matched/decals-dr7.0-cosmos-acs-match.fits')

---------------------------------------------------------------------------
Empty                                     Traceback (most recent call last)
~/repos/desihub/desitarget/py/desitarget/internal/sharedmem.py in get(self, Q)
    376             try:
--> 377                 return Q.get(timeout=1)
    378             except queue.Empty:

/usr/local/anaconda3/envs/desi/lib/python3.6/multiprocessing/queues.py in get(self, block, timeout)
    104                     if not self._poll(timeout):
--> 105                         raise Empty
    106                 elif not self._poll():

Empty: 

During handling of the above exception, another exception occurred:

StopProcessGroup                          Traceback (most recent call last)
~/repos/desihub/desitarget/py/desitarget/internal/sharedmem.py in map(self, func, sequence, reduce, star)
    673                 try:
--> 674                     capsule = pg.get(R)
    675                 except queue.Empty:

~/repos/desihub/desitarget/py/desitarget/internal/sharedmem.py in get(self, Q)
    379                 if not self.is_alive():
--> 380                     raise StopProcessGroup
    381                 else:

StopProcessGroup: StopProcessGroup

During handling of the above exception, another exception occurred:

SlaveException                            Traceback (most recent call last)
<ipython-input-9-f1b5451e3b7c> in <module>()
----> 1 targ = select_targets('decals-dr7.0-cosmos-acs-match.fits')

~/repos/desihub/desitarget/py/desitarget/cuts.py in select_targets(infiles, numproc, qso_selection, match_to_gaia, sandbox, FoMthresh, Method, gaiadir)
   2241                 targets = pool.map(_select_sandbox_targets_file, infiles, reduce=_update_status)
   2242             else:
[...]

PR #338, opened 2018-08-03
Code updates for DR7 targeting

geordie666:
This PR updates `desitarget` to produce current sets of target bits for DR7:

- Update the LRG, QSO, STD and MWS algorithms to align with https://desi.lbl.gov/trac/wiki/TargetSelectionWG/TargetSelection.
- In particular, major updates to the STD and MWS
- Don't match to Gaia by default, only if requested.
- Maintain the capability to match to Gaia if needed for earlier Data Releases.
- Run individual target classes by passing, e.g., --tcnames QSO,STD to the executable.
- Update unit test files to not rely on Gaia.
- Bring Data Model into agreement with Legacy Surveys sweeps files.
- Rename `FSTD` to be `STD` throughout.
- QA fails gracefully if the weight maps required to produce systematics aren't passed.


geordie666:
@moustakas: If you have some time in the next hour or two, it would be great if you could check that this PR solves https://github.com/desihub/desitarget/issues/336. Even more awesome would be if you could scan my changes to the `MWS` and `STD` selections and check that they agree with the logic in https://desi.lbl.gov/trac/wiki/TargetSelectionWG/TargetSelection.

geordie666:
I'm going to go ahead and merge so I can keep working on related branches.

PR #340, opened 2018-08-06
Write Gaia columns into sweeps

geordie666:
The data model for the DR6 sweeps currently doesn't include the Gaia columns necessary for target selection. This PR creates a simple function and executable that can write the sweeps with only the columns necessary for target selection (including the Gaia columns). I also addressed a few typos and minor bugs.

Example outputs are in `/global/cscratch1/sd/adamyers/gaiamatch/dr6/sweep/6.0`.


geordie666:
This should be ready to merge, now (if Travis passes). This handful of updates shouldn't be controversial so I'll self-merge by tomorrow morning if nobody else looks at it before then.

PR #341, opened 2018-08-09
Fix stdstar bit names

sbailey:
Changes in this PR:
* Rename STD -> STD_FAINT for symmetry with STD_BRIGHT.
  * Very recently this was renamed from STD_FSTAR -> STD, but now we're renaming it again.
* Fix mock targets usage of STD_FSTAR vs. STD vs. STD_FAINT and related function calls.
* Fix mock targets STD/MWS selections which now require Gaia.
* Reserve bits for STD_FAINT_BEST and STD_BRIGHT_BEST, but don't use them yet (they are still commented).

@geordie666 made the original changes which I vetted, and then I hijacked this to do the STD -> STD_FAINT renaming too, so I'll punt it back to him for final approval.

Once this is merged I'll make a new desitarget tag and I have a desisim PR ready to go which will re-fix desisim for the bit renaming.  I'm not submitting that PR yet because Travis tests will fail anyway until this PR is merged.

Due to great foresight I believe fiberassign is already ready for the new bit names, but I will also verify that after desitarget + desisim have been updated.

### DIFF
--- a/py/desitarget/QA.py
+++ b/py/desitarget/QA.py
@@ -1417,7 +1417,7 @@ def _load_targdens(bitnames=None):
     targdens['LRG'] = targdict['ntarget_lrg']
     targdens['QSO'] = targdict['ntarget_qso'] + targdict['ntarget_badqso']
     targdens['BGS_ANY'] = targdict['ntarget_bgs_bright'] + targdict['ntarget_bgs_faint']
-    targdens['STD'] = 0
+    targdens['STD_FAINT'] = 0
     targdens['STD_BRIGHT'] = 0
     targdens['MWS_ANY'] = targdict['ntarget_mws']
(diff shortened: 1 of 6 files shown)


geordie666:
Looks great. I've confirmed that `select_targets` still runs on various data releases. Feel free to merge after you update the `changes.rst` file (or update that file when you tag).

Thanks for doing the extra work to incorporate the new `STD_FAINT` nomenclature throughout.

moustakas:
For the record, the standard-star bit names were changed in this PR but not updated on the wiki --
https://desi.lbl.gov/trac/wiki/TargetSelectionWG/TargetSelection#SpectrophotometricStandardStarsSTD

geordie666:
I updated this on the wiki.

Issue #319, opened 2018-05-08
Additional columns for GFA targets


sbailey:
Additional columns to add to `select_gfa` output:

  * [x] FLUX_G, FLUX_Z (float32, nanomaggies) [FLUX_R is already there]
  * [x] TYPE (S4, morphological type)
  * [x] BRICKID (int32)
  * [x] BRICK_OBJID (int32) [OBJID in sweep file]

Placeholders for proper motions from Gaia, though these aren't yet in the sweeps:

  * [x] PMRA, PMDEC (float32) default 0.0 (proper motions from Gaia)
  * [x] PMRA_IVAR, PMDEC_IVAR (float32) default 1.0 (inverse variances of PM*)
  * [x] PARALLAX (float32) default 0.0
  * [x] PARALLAX_IVAR (float32) default 1.0
  * [x] Header keyword (not column) REFEPOCH = 2015.5 // Gaia Proper Motion Reference Epoch

Placeholder boolean flags, for now set True if TYPE='PSF' though we may add additional cuts later for isolation, color, etc.

  * [ ] GUIDE_OK  // ok to use for guiding
  * [ ] SEEING_OK  // ok to use for seeing measurements
  * [ ] FOCUS_OK  // ok to use for focusing

geordie666:
@sbailey: At some point I'm going to update the GFAs code so that it can either match to Gaia, or just directly take columns from the sweeps. Do we still need the `PARALLAX` columns? Should I add the `GUIDE_OK`, `SEEING_OK`, `FOCUS_OK` columns, or does that happend downstream?

sbailey:
Implemented by PR #342

Issue #232, opened 2017-11-06
Generate GFA target catalog for each tile


sbailey:
In addition to fiber assigned targets, we need to provide ICS with a list of every object that might be visible on the GFAs.  These will be used for field acquisition, guiding, and focus.

`desimodel.focalplane.get_gfa_targets` and lower-level `.on_gfa`, and `.on_tile_gfa` provide routines for doing this.  These need to be combined with desitarget and/or fiberassign into output files to pass forward.

To do:
* Define format.  e.g. options include
  * an extra HDU of the fiberassignment files (easy and natural for interface with ICS, but mixes C++ and python code from separate repos...)
  * separate files so that GFA target identification can be developed and run separately from fiberassignment
* Write wrapper code to actually do that.

This isn't particularly hard since `desimodel.focalplane.get_gfa_targets` already exists, but finalizing the details is absolutely a blocking factor for first light.

sbailey:
Update: the format output by `select_gfas` in PR #342 is good, e.g. /global/cscratch1/sd/adamyers/gfas-dr6.0-PR342.fits (note: on $SCRATCH, will not be around forever).  We still need a standalone wrapper or instructions for using that to generate the subset that actually overlaps real GFAs.  fiberassign does that subselection but doesn't propagate all columns.  Adding this to 18.9 project and leaving it open until we have that wrapper.

sbailey:
This is included in fiberassign output now.  Closing ticket.

PR #342, opened 2018-08-10
Update code that produces GFA targets

geordie666:
Better processing of GFA targets:

- Handles reading Gaia columns directly from the sweeps in addition to matching to Gaia files. 
- Defaults to *not* matching to Gaia files.
- Makes Gaia matching radius stricter to return only the best Gaia objects.
- Retains Gaia RA/Dec when matching, instead of RA/Dec from sweeps.
- Fixes a bug where potential Gaia matches in some HEALPixels weren't being read.
- Adds a default Gaia epoch to the GFA file header (the actual epoch still needs passed through from the sweeps).

Should fix:

https://github.com/desihub/desitarget/issues/319
https://github.com/desihub/desitarget/issues/232

### DIFF
--- a/bin/select_gfas
+++ b/bin/select_gfas
@@ -44,7 +44,8 @@ if len(infiles) == 0:
 
 log.info('running on {} processors...t={:.1f}mins'.format(ns.numproc, (time()-time0)/60.))
 
-gfas = select_gfas(infiles, maglim=ns.maglim, numproc=ns.numproc, gaiadir=ns.gaiadir)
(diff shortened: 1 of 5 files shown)


geordie666:
@sbailey: It would be useful  if you review this PR to check it fixes

https://github.com/desihub/desitarget/issues/319
https://github.com/desihub/desitarget/issues/232 

or if we still need something else for the GFAs. An example output file is at `/global/cscratch1/sd/adamyers/gfas-PR342.fits`

Issue #333, opened 2018-07-20
Update LRG NUMOBS to use only targeting bits


sbailey:
`desitarget.targets.calc_numobs` still uses magnitudes + LRG targeting bits to determine the requested number of observations.  This should be updated to use just the targeting bits LRG_1PASS and LRG_2PASS without a separate magnitude cut in `calc_numobs`.  We don't currently have LRG_3PASS defined as a bit, but perhaps we should add that as a placeholder possibility.

PR #345, opened 2018-08-14
Set NUMOBS for LRGs in MTL using targeting bits instead of magnitude cuts

geordie666:
This should fix https://github.com/desihub/desitarget/issues/333


moustakas:
This looks good to me.

Issue #354, opened 2018-08-29
cuts.isSTD(usegaia=False) still depends on Gaia inputs


moustakas:
@geordie666 I know that Gaia catalog entries are now required for selecting standard stars.  However, calling `cuts.isSTD` with `usegaia=False` (and without passing the army of Gaia arguments) crashes because the brightness cut is made on the Gaia G-band magnitude.  

Please feel free to --
1. Ignore and close this issue.
2. Wrap the applied brightness cut in the `usegaia` logic (thereby not requiring

moustakas:
I had missed the documentation note, but this actually came up in the context of applying `isSTD(usegaia=False)` to the `targets-dr7.1-0.23.0.fits` catalog which, to my surprise, crashed, i.e., not in the context of the mocks.  I did this because I was specifically trying to isolate the effect of the Gaia+color cuts from the color-cuts by themselves.

I don't think it's crazy to have the code not crash if `usegaia=False` without having to read the code to understand why.  But in any case, take a look at the fix I proposed in #355.

geordie666:
That fix looks fine. But, please change the documentation to reflect the fix! You and I understand what's going on, but I'd hope that someone looking at the code for the first time would determine what to pass from the doc string.

geordie666:
Oh, and we should change your fix to r-band, not g-band. I believe r-band better approximates Gaia G-band.

PR #355, opened 2018-08-29
adapt isSTD_colors to deal with differences in north/south color-cuts

moustakas:
Adds a `south` argument (defaulting to `True`) to `cuts.isSTD_colors` to allow for the possibility that we apply different color-cuts to standard stars with north vs south photometry (separate from the footprint-agnostic Gaia cuts).  

Currently the *grz* color-cuts are identical, as documented [here](https://desi.lbl.gov/trac/wiki/TargetSelectionWG/TargetSelection#SpectrophotometricStandardStarsSTD).

This PR also addresses #354.

### DIFF
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -806,9 +806,10 @@ def isSTD(gflux=None, rflux=None, zflux=None, primary=None,
         std &= gaiagmag >= gbright
         std &= gaiagmag < gfaint
     else:
-        gmag = 22.5 - 2.5 * np.log10( gflux.clip(1e-16) )
-        std &= gmag >= gbright
-        std &= gmag < gfaint
+        # Use LS r-band as a Gaia G-band proxy.
+        gaiamag_proxy = 22.5 - 2.5 * np.log10( rflux.clip(1e-16) )
+        std &= gaiamag_ >= gbright
+        std &= gaiamag_ < gfaint
 
     return std
 
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -746,16 +746,17 @@ def isSTD(gflux=None, rflux=None, zflux=None, primary=None,
            if ``True`` apply magnitude cuts for "bright" conditions; otherwise, 
            choose "normal" brightness standards. Cut is performed on `gaiagmag`.
         usegaia: boolean, defaults to ``True``
-           if ``True`` then  call :func:`~desitarget.cuts.isSTD_gaia` to set the 
+           if ``True`` then call :func:`~desitarget.cuts.isSTD_gaia` to set the
            logic cuts. If Gaia is not available (perhaps if you're using mocks)
-           then send ``False`` and pass `gaiagmag` as 22.5-2.5*np.log10(`robs`) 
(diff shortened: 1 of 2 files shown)


moustakas:
Belatedly, I also added a `south` argument to the `isMWS_STAR` function, which is *only* used in desisim-only simulations (e.g., by `desisim.obs.new_exposure`) (i.e., not by mocks and not for target selection).

py/desitarget/cuts.py
```diff
@@ -788,8 +802,13 @@ def isSTD(gflux=None, rflux=None, zflux=None, primary=None,
         gbright = 16.
         gfaint = 19.
 
-    std &= gaiagmag >= gbright
-    std &= gaiagmag < gfaint
+    if usegaia:
+        std &= gaiagmag >= gbright
+        std &= gaiagmag < gfaint
+    else:
+        gmag = 22.5 - 2.5 * np.log10( gflux.clip(1e-16) )
```
geordie666: I believe that Gaia G-band is better approximated by r than g. So, consider changing this to r fluxes.

py/desitarget/cuts.py
```diff
@@ -740,6 +750,9 @@ def isSTD(gflux=None, rflux=None, zflux=None, primary=None,
            logic cuts. If Gaia is not available (perhaps if you're using mocks)
            then send ``False`` and pass `gaiagmag` as 22.5-2.5*np.log10(`robs`) 
            where `robs` is `rflux` without a correction.for Galactic extinction.
+        south: boolean, defaults to True
```
geordie666: Please modify the documentation to indicate that it is not necessary to pass an r flux if `usegaia=False`

py/desitarget/cuts.py
```diff
@@ -749,13 +762,14 @@ def isSTD(gflux=None, rflux=None, zflux=None, primary=None,
...
             https://desi.lbl.gov/trac/wiki/TargetSelectionWG/TargetSelection?version=121#STD
+
     """
     if primary is None:
         primary = np.ones_like(gflux, dtype='?')
     std = primary.copy()
 
     #ADM apply the Legacy Surveys (optical) magnitude and color cuts.
-    std &= isSTD_colors(primary=primary, zflux=zflux, rflux=rflux, gflux=gflux)
+    std &= isSTD_colors(primary=primary, zflux=zflux, rflux=rflux, gflux=gflux, south=south)
```
geordie666: Ultimately, we'll need bits like `STD_FAINT_NORTH`, `STD_BRIGHT_SOUTH` etc. to be set. For now, let's leave this. I'm not sure that the standard stars will need different optical selections as they're point sources and they're quite far from the flux limits of the imaging. If the MWS team decides that separate optical cuts *are* necessary, we'll update the bits then.

moustakas:
Thanks @geordie666.  For the G-band magnitude proxy I cheated and used the extinction-corrected r-band magnitude, since it was already an argument.  This is such a narrow application of this function that I don't think it really matters.  

If you're satisfied with these changes please merge this PR.  Thanks.

geordie666:
Looks good. I'll merge when tests pass.

PR #358, opened 2018-08-30
add isELG_colors, isELG_colors_north, and isELG_colors_south functions

moustakas:
This PR adds new `isELG_colors`, `isELG_colors_north`, and `isELG_colors_south` functions in `desitarget.cuts` in order to separately apply color-cuts for data from the south vs the north (defaulting to south, i.e., DECaLS).

The corresponding functions for LRGs, STDs, and QSOs already exist.


geordie666:
Looks great. Thanks.

Issue #351, opened 2018-08-24
clean up QA module


moustakas:
A very low-priority suggestion:

There are a bunch of legacy functions in `desitarget.QA` (e.g., brick fluctuations code) that may be useful to clean up, just so it's easier to navigate the core pieces of code required to do target selection QA.

geordie666:
Closed in https://github.com/desihub/desitarget/pull/359.

PR #359, opened 2018-08-31
First fully operational commissioning target selection; remove old code in QA

geordie666:
This PR is focused on making commissioning target selection fully functional. Includes:

- All of the cmx target selection algorithms currently listed on the wiki at https://desi.lbl.gov/trac/wiki/TargetSelectionWG/CommissioningTargets.
- First unit tests for cmx (> 90% coverage).
- Adding SV and CMX as columns in the output files instead of as a bit.

I also followed one of @moustakas suggestions and removed most of the old code in QA (this code mostly predates our use of HEALPixels, and the functionality to make HEALPix weight maps is now in `desitarget.randoms.pixmap`)


geordie666:
@sbailey: You may want to review that the unit tests are satisfactory.
@moustakas: You may want to check that I've removed the legacy code from QA, as you suggested in https://github.com/desihub/desitarget/issues/351.

geordie666:
Thanks @moustakas: This only deals with updates to commissioning not SV or the main survey. 

I have, indeed, run `select_cmx_targets` on both DR6 and DR7 files. I've only implemented what is currently on the wiki at https://desi.lbl.gov/trac/wiki/TargetSelectionWG/CommissioningTargets. Other cmx target classes will certainly be added.

`OBSCONDITIONS['APOCALYPSE']` has been an option for the main survey since I took over the `desitarget` code. Whether it's still used downstream is a question for @sbailey, I guess!

PR #361, opened 2018-09-01
grab Gaia columns in a simple wrapper

moustakas:
Adds a convenience function for pulling out the requisite Gaia columns (both existing and derived) from an input catalog before target selection cuts are applied.

Having this wrapper makes it easier for the mocks to use the same columns, including whatever future updates occur.

### DIFF
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -1971,6 +1971,76 @@ def _is_row(table):
     else:
         return False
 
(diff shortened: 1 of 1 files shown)


geordie666:
Looks good. We should consider a similar refactor to `_prepare_gaia` for the optical columns, to clean the code up...and so we can just call those `_prepare` functions in SV and commissioning.

Feel free to self-merge when tests pass.

PR #362, opened 2018-09-02
top-level selection of QSOs using just optical color-cuts

moustakas:
This small PR propagates the option of applying just optical color-cuts to the selection of QSOs to the top-level `cuts.apply_cuts` wrapper.  Optical color-cuts are all we can do when generating targets using `desisim.templates.QSO`, which do not extend red enough to allow us to synthesize WISE photometry.

Although we have `desisim.templates.SIMQSO` the Lya WG still would like to be able to generated QSO continuum spectra using the PCA-based technique implemented in `desisim.templates.QSO`.


geordie666:
Looks good. If you want to comment out the Gaia/parallax warning you noted offline in this PR, too, then go ahead and self-merge. Or, feel free to self-merge without suppressing that warning if you're handling it in a different branch.

PR #364, opened 2018-09-04
Add STD_WD target class to main survey cuts

geordie666:
This PR adds the `STD_WD` class to `desitarget.cuts` and sets that bit in output files. The `STD_WD` class is currently identical to the `MWS_WD` class.

Also adds a `_prepare_optical_wise` function to clean up some of the initial setting of columns of data, and utilizes that function and the `_prepare_gaia` function in `cmx_cuts`.

PR #366, opened 2018-09-05
use new desiutil.dust.SFDMap module

moustakas:
* Removes `desitarget.mock.sfdmap` module;
* Uses new `desiutil.dust.SFDMap` module (see https://github.com/desihub/desiutil/pull/116)


geordie666:
Should we remove this completely, as you've done, or keep the module but add a deprecation warning? I *think* it's probably only you (@moustakas) and I that we're calling `mock.sfdmp`, so I *think* we're safe to remove it completely rather than deprecate it. Just a thought.

geordie666:
Feel free to merge when tests pass if you think it's only the two of us that are using this.

moustakas:
Tests are failing because of the same conda bug discussed in #364, so I'm going to go ahead and merge.

@geordie666 Since getting rid of the dust module here only impacts me and you, I decided to just pull the bandaid off and skip the deprecation warning.

PR #370, opened 2018-09-07
Define additional commissioning classes and priorities.

geordie666:
This PR is focused on updating the commissioning code to include:

- New functions to define several more commissioning classes.
- A $CMX_DIR directory that contains files of sources to which to match.
- The functionality to reset initial priorities for commissioning targets.
- Downloading fitsio using pip/astropy to fix Travis.


geordie666:
Example output file (run over all of DR7.1) is at `/global/cscratch1/sd/adamyers/cmx-test-dr7.1.fits`.

geordie666:
And an example `$CMX_DIR` is `/project/projectdirs/desi/target/cmx_files`

moustakas:
@geordie666 A couple comments:

* I didn't have any problems reading the file, but if you could fix the permissions for `/global/projecta/projectdirs/desi/target/cmx_files` that'd be great.
* The binary table HDU should have `EXTNAME` set (to something).
* The columns of the table also have mixed case, i.e.,
```
<Table length=88>
 name  dtype 
----- -------
 Name bytes12
   RA float64
  Dec float64
 pmra float64
pmdec float64
```
I didn't dig to see what code makes this file but it'd be good if the columns were all uppercase (our adopted convention, if I recal), and I also wonder if that table could have the Gaia `SOURCE_ID` added?
* The `cmx-test-dr7.1.fits` catalog has 55 million(!) targets and the file is 20GB small.  I think this is OK---these are stars after all---but I wonder if the over-abundance of targets is a good thing or whether it may cause unforeseen issues (e.g., pushing us to select brighter targets).

geordie666:
I've changed the permissions on that file, and also asked @crockosi, who produced the file, to remake it to conform better to Data Systems standards.

I don't *think* we need the code that produced this file. I'm treating the file as an "external list of RAs and Decs."

sbailey:
Code looks fine, though agreed with John's comments about not liking mixed case columns.  Please document exactly which columns in $CMX_DIR files are required vs. optional.  It appears that setting the cmx bits will require a specific set of files in $CMX_DIR; /project/projectdirs/desi/target/cmx_files/ is a good start at curating that.  Should we move that into versioned subdirectories, or do you expect updates to only be additions of new files, not changes to pre-existing files?  Alternately we could keep a single curated directory but require versions in the filenames.

For both this and external catalog matching in general, we should strive to keep the entry barrier as low as possible for contributing a new file.  e.g. we could be flexible in *recommending* EXTNAME and UPPERCASE columns, but not *requiring* it, defaulting to HDU 1 and using `fitsio.read(..., upper=True)` so that code can standardize on RA, DEC even if the input file has lower or mixed-case.

geordie666:
My thoughts:

I agree with the low barrier to external files. I'll write in defaults of `upper=True` and no `EXTNAME`.

As regards versioning: I'll think carefully about that for secondary targets *and* if we add any other files to `CMX_DIR` for commissioning. I quite like the idea of adding a version to the file name and only reading in the highest version number for any file. For secondary targets, we'll need to also add a sub-directory structure. A single directory would probably fill up quickly.

Issue #369, opened 2018-09-07
Add columns to target output files.


geordie666:
This is transcribed from https://github.com/desihub/desitarget/issues/209, which was fixed in the mocks but not yet in the real data.

A few items (for the real target catalog):
* `BRICKID`: let's use the real ra,dec -> brickid mapping rather than the healpix number, especially since we also have a HPXPIXEL column.  Create a `b = desiutil.brick.Bricks(bricksize=0.25)` object, then call `b.brickid(ra,dec)`.
* `TYPE` -> `MORPHTYPE`.  Although we generally don't want to rename columns arbitrarily, there are too many other "types" downstream (spectype, postionertype, objtype, ...) so by the time this gets to the ICS it needs to be "MORPHTYPE".  Let's do this change here at target selection (I would prefer it to even happen upstream in the sweeps or tractor, but that may be a can of worms).
* Let's drop the unused W3 and W4 fields (FLUX_, FLUX_IVAR_, MW_TRANSMISSION_) unless there is a strong science case for keeping them (they aren't used by targeting or the spectro pipeline).

moustakas:
> Let's drop the unused W3 and W4 fields (FLUX_, FLUX_IVAR_, MW_TRANSMISSION_) unless there is a strong science case for keeping them (they aren't used by targeting or the spectro pipeline).

I actually think we should get the input of the science working groups on this one.  Of course one could always go from redshift catalogs back to the sweeps to get these fluxes, but that's pretty inconvenient given how the sweeps are organized.  (Maybe we really should post-process them into healpixels...)  Anyway, I can send that email if you want @geordie666, or we could just move on and drop W3,W4.

geordie666:
@moustakas: If you could shoot a quick email to the mailing lists just to check W3/W4 aren't critical, that'd be great.

I say we drop them unless people really need them. Potentially, anyone can ask for any columns in the sweeps, or the Tractor files. I think it would make more sense to have code to match back on `TARGETID` and pick up extra information, rather than carry around all of the columns.

On the post-processing into HEALPixels: One option would be to just write an indexing file that noted which sweeps files needed to be touched to retrieve a HEALPixel. It's easy to traverse the nested tree to different HEALPixel resolutions.

forero:
I am working on adding some columns in `fiberassign` outputs. It seems that the following should come from files written by `select_targets`. 

* [ ] LAMBDA_REF 
* [ ] SECONDARY_TARGET	
[...]

PR #372, opened 2018-09-10
Update Data Model for target files

geordie666:
This PR updates the Data Model for target files to fix issue https://github.com/desihub/desitarget/issues/369. It also sets up code in `io` to read `RA` and `DEC` from loosely formatted files with a view to coordinate-matching external sources for commissioning and secondary targets.


geordie666:
I'm going to leave this until tomorrow before contemplating merging it, to see if anyone has a critical use of the W3 and W4 fluxes for targeting.

geordie666:
Note, though, that the schema for secondary targets will be for people to run their own algorithms outside of the Data Systems infrastructure and then `desitarget` will match back to the sweeps on `RA`/`DEC` or on `BRICKID`/`OBJID`. So, no additional columns are necessary in the sweeps for secondary target selection. People proposing secondary targets will curate their own algorithms, and don't have to select anything from the sweeps if they don't want to (they could use an X-ray or radio selection, for instance).

geordie666:
I think this is the consensus state for the data model with `TYPE`-> `MORPHTYPE`, `EBV` included and `W3` and `W4` columns still included. I'll merge this in a few hours unless anyone complains.

py/desitarget/cmx/cmx_cuts.py
```diff
@@ -446,12 +446,12 @@ def isSTD_calspec(ra=None, dec=None, cmxdir=None, matchrad=1.,
...
     # ADM get the CALSPEC objects.
     cmxfile =  os.path.join(cmxdir,'calspec.fits')
-    cals = fitsio.read(cmxfile)
+    cals = io.read_external_file(cmxfile)
 
     # ADM match the calspec and sweeps objects.
     calmatch = np.zeros_like(primary, dtype='?')
     cobjs = SkyCoord(ra*u.degree, dec*u.degree)
-    ccals = SkyCoord(cals['RA']*u.degree, cals["Dec"]*u.degree)
+    ccals = SkyCoord(cals['RA']*u.degree, cals["DEC"]*u.degree)
```
moustakas: I didn't explicitly test this, but I think this will fail if the external RA, Dec columns already have the `unit` attribute (which may not be true for the *current* `calspec.fits` file, but may be true in the future and/or for other external catalogs). In `cuts._gal_coords` I did something like:
```
    import astropy.units as u
    from astropy.coordinates import SkyCoord

    if hasattr(ra, 'unit') and hasattr(dec, 'unit') and ra.unit is not None and dec.unit is not None:
        c = SkyCoord(ra.to(u.deg), dec.to(u.deg))
    else:
        c = SkyCoord(ra*u.deg, dec*u.deg)
```

geordie666:
@moustakas: I'm on a machine with an old browser and it seems that I can't respond to your review directly, but I will when I get back to my laptop. In any case, I think that forcing `unit=degree` in `SkyCoord` is more elegant, and should work, as in:

```
import astropy.units as u
from astropy.coordinates import SkyCoord
a = Table.read("/project/projectdirs/cosmo/data/legacysurvey/dr7/tractor/004/tractor-0046m007.fits")
print("Unit of ra is {}, unit of dec is {}".format(a["ra"].unit,  a["dec"].unit))
```
```
blat = SkyCoord(a['ra'], a["dec"], unit='degree')
```
Apologies that I keep forgetting to enforce support for Tables with units!

I'll update the code accordingly and push, but let me know if that still fails for your use cases.

moustakas:
@geordie666 do you have an example output file from this code so I can compare the latest data model with the output of `select_mock_targets`?   Thanks.

geordie666:
@moustakas: 

`/project/projectdirs/desi/target/catalogs/dr7.1/targets-dr7.1-PR372.fits`

PR #373, opened 2018-09-12
Remove 2.7 checks from Travis

geordie666:
We deprecated support for Python 2.7 some time ago, but we still run the Travis tests for 2.7. This PR removes 2.7 checks from our Travis build, and adds an allowed-to-fail PEP 8 compliance check, which I personally find useful for cleaning up code style.


geordie666:
This doesn't actually change any code, just what is triggered in `.travis.yml`. So, I'll merge fairly soon after tests pass.

Issue #377, opened 2018-09-14
standard-star target density in the mocks should be tunable


moustakas:
In the `select_mock_targets` configuration file we really need a parameter which sets the (desired) target density of standard stars.  Currently the downstream code requires roughly 30 standards/deg2, so we should make sure this gets achieved.

We've been bitten several times by this issue (i.e., ending up with too few standards).

sbailey:
+1.  The current (as of PR #378) density of 1000/deg2 is way too high to be realistic.

moustakas:
I looked into implementing a new parameter in the configuration file to tune the number of standard stars, but I am confident that the excess was due to the bug which was fixed in #383, so I'm closing this issue.

Issue #352, opened 2018-08-27
add Gaia-style columns to Galaxia/MWS mocks?


moustakas:
@apcooper I don't know how much effort would be involved, or if this activity is already in progress, but are there any plans to add Gaia-style columns to the Galaxia mocks?  As of #338, our standard star selection now *requires* Gaia, so this is a necessary update/upgrade to make our mocks and simulated datasets as accurate as possible.

Thanks--

moustakas:
Apologies for the premature post, but it looks like there are already some Gaia catalogs in `/global/projecta/projectdirs/desi/mocks/mws/galaxia/alpha/v0.0.5` with columns:
```
['V_LUPTON_NODUST',
 'I_LUPTON_NODUST',
 'V_LUPTON',
 'I_LUPTON',
 'G_GAIA',
 'G_GAIA_RVS',
 'PM_RA_GAIA_ERROR',
 'PM_DEC_GAIA_ERROR',
 'PM_RA_STAR_GAIA',
 'PM_DEC_GAIA',
 'PARALLAX_GAIA_ERROR',
 'PARALLAX_GAIA',
 'V_HELIO_GAIA_ERROR_G0V',
 'V_HELIO_GAIA_G0V',
 'V_HELIO_GAIA_ERROR_F0V',
 'V_HELIO_GAIA_F0V',
 'V_HELIO_GAIA_ERROR_K1III',
 'V_HELIO_GAIA_K1III',
 'V_HELIO_GAIA_ERROR_K1IIIMP',
 'V_HELIO_GAIA_K1IIIMP']
```
However, we would need all the Gaia columns listed in the sweep files--
  http://www.legacysurvey.org/dr7/files/#sweep-catalogs 
in order to select standards --
https://desi.lbl.gov/trac/wiki/TargetSelectionWG/TargetSelection#SpectrophotometricStandardStarsSTD

Looking a bit more closely, I can ignore / mock up the `GAIA_ASTROMETRIC_EXCESS_NOISE` and `GAIA_DUPLICATED_SOURCE` columns, so all I need is `GAIA_PHOT_BP_MEAN_MAG`, `GAIA_PHOT_BP_MEAN_FLUX_OVER_ERROR`, `GAIA_PHOT_RP_MEAN_MAG`, and `GAIA_PHOT_RP_MEAN_FLUX_OVER_ERROR`, assuming the G-band magnitudes, proper motions, and parallaxes in the Galaxia mocks are in the same units as the sweep files.

[...]

PR #378, opened 2018-09-14
bug fix: correctly select standards from MWS_MAIN mocks

moustakas:
This PR fixes a bug which resulted in only WDs being selected as standards by `select_mock_targets`, i.e., too few for the down-stream pipeline.

Briefly: standards are required to have `FRACMASK_[G,R,Z]>0` but in the mock target catalog `FRACMASK` had (previously) been set to zero in all bands.  (This question is being discussed on @desi-targets---follow it there if interested.)

Another bug was that only `MWS` targets were being selected from the MWS_MAIN/Galaxia mocks, instead of `MWS` *and* `STD` now.

The PR also adds some hooks to restore support for the FAINTSTAR mocks (once they have Gaia; see #352), but those are still not being used (nor do they need to be).

Testing over one healpixel yields approximately 1000 standards/deg2 (which is now overkill, but see #377).
```
select_mock_targets -c select-targets.yaml --seed 1 --nside 16 --healpixels 352 --nproc 4
```
with the following configuration file:
```
targets:
    MWS_MAIN: {
        target_type: STAR,
        mockfile: '{DESI_ROOT}/mocks/mws/galaxia/alpha/v0.0.5/healpix',
        nside_galaxia: 8,
        format: galaxia,
    }

```

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,6 +5,8 @@ desitarget Change Log
 0.24.0 (unreleased)
 -------------------
 
+* Fix a bug which resulted in far too few standard stars being selected in the
+  mocks [`PR #378`_].  
 * Fix a bug in how the `objtruth` tables are written out to by
   `select_mock_targets` [`PR #374`_].
 * Remove Python 2.7 from Travis, add an allowed-to-fail PEP 8 check [`PR #373`_].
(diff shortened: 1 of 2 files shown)


sbailey:
Thanks for the quick fix.  It looks like the current FRACMASK cut is just wrong, but this provides a workaround that both passes the current cuts and likely will pass the corrected cuts too.  OK to merge as soon as Travis tests pass.

PR #380, opened 2018-09-17
First fully operational SV code

geordie666:
This sets up the final infrastructure for SV code, with the philosophy of using as much code from the main survey "cuts" code as possible, while retaining independent cuts for each iteration of SV. Includes

- Refactor such that (almost) the only evolving part of the code for SV will be the cuts.
- Unit tests for SV that should be easy to maintain even with additional iterations of SV.
- Bit and column setting for SV that should be easy to maintain.
- SV0 (commissioning) MWS cuts.
- Updated STD cuts (in both main survey and SV) to fix a `fracmasked` typo.
- Minor alterations to Travis coverage to exclude some external code.

One note: To pass unit tests, I had to alter the `travis.yml` file to install `healpy` from pip instead of conda.


geordie666:
I'll merge this in a few hours to push on with other development unless I receive any feedback.

moustakas:
Thanks for the heads up @geordie666 but I think these changes are only affecting me, and it would be great to get the bug fixes in for the stuff I'm working on.

Issue #375, opened 2018-09-13
add FIBERFLUX columns to mock targets catalog


moustakas:
Related to #369, the `targets.fits` files generated by `select_mock_targets` needs to add the `FIBERFLUX_[GRZ]` and `FIBERTOTFLUX_[GRZ]` columns needed by fiberassignment.  

@dkirkby is't there code (e.g., in `specsim`) to estimate these quantities from a given surface brightness profile, ellipticity, and size?  If so, can you point me to the code here?

For reference `FIBERASSIGN_X` is defined as the "band flux within a fiber from this object in 1 arcsec Gaussian seeing" (see [here](http://legacysurvey.org/dr7/files).

sbailey:
This is now a blocking factor because fiberassign master requires these columns (albeit for propagation only) but select_mock_targets isn't generating them.


@moustakas the relevant code is in `specsim.fiberloss.calculate_fiber_acceptance_fraction`, using the "fastsim" method so that we don't require galsim.  Trace downwards from `desisim.simexp.simulate_spectra` for an example that uses "fastsim", but there should be a way to directly call `calculate_fiber_acceptance_fraction` without going through the full spectral simulation.  @julienguy provided the "fastsim" method and may have a simpler example of how to use it.  See desihub/specsim#78 .

Related: This notebook fiberloss calculations for the "galsim" method:

https://github.com/desihub/specsim/blob/master/docs/nb/FiberlossExamples.ipynb

it calls https://github.com/desihub/specsim/blob/master/specsim/quickfiberloss.py .  It looks like the calling sequence for "fastsim" is different enough than for "galsim" that I'm not sure how useful those will be.

Even typical/median values as a fixed ratio from `FLUX_[GRZ]` for each target type would be helpful for moving forwards in the meantime.

julienguy:
```
FastFiberAcceptance.value(self, source, sigmas, offsets, hlradii=None)

returns the fiber acceptance for the given source,sigmas,offsets

Args:
    source (string) : POINT, DISK or BULGE for point source, exponential profile or De Vaucouleurs profile
    sigmas (np.array) : arbitrary shape, values of sigmas in um for the PSF due to atmosphere and telescope blur
    offsets (np.array) : same shape as sigmas, values of offsets on focal surface between fiber and source, in um

Optional:
    hlradii (np.array) : same shape as sigmas, half light radius in arcsec for source

Returns np.array with same shape as input
```

Example

```
import numpy as np
import os
[...]

PR #381, opened 2018-09-18
estimate and return FIBERFLUX_[G,R,Z] for mock targets

moustakas:
Addresses #375 by estimating `FIBERFLUX_[G,R,Z]` (which we set equal to `FIBERTOTFLUX_[G,R,Z]`; see the [DR7 documentation](http://legacysurvey.org/dr7/files/#sweep-7-0-sweep-brickmin-brickmax-fits)) for each mock target using a DR7-calibrated mixture of morphological types and sizes (see [gmm-dr7.ipynb](https://github.com/desihub/desitarget/blob/master/doc/nb/gmm-dr7.ipynb)) and the `specsim.fastfiberacceptance.FastFiberAcceptance` methods.

I assume nominal 1 arcsecond FWHM seeing at 5500 Angstrom, no systematic fiber offset, and a lambda^(-1/5) variation of the seeing disk with wavelength (using the appropriate filter effective wavelength for sources with `PHOTSYS=N` vs `PHOTSYS=S`.

It would be great to have the QA check to check these and other values, but small tests show that the values look pretty sensible.


moustakas:
Testing with a single `nside=64` healpixel (5636) over all target classes here's the distribution of FIBERFLUX/FLUX (in the r-band), which prompted me to push a fix such that FIBERFLUX cannot exceed FLUX.  

There are some oddities (e.g., QSO should be point-like!) but unfortunately I'll be offline until tomorrow evening.

hist

sbailey:
Thanks for this rapid PR while on your way to Chile.  Something is odd with the stars -- they shouldn't be piled up around 0.  In the DR7 data they are around 0.78.  Similarly ELGs shouldn't be able to go out all the way to 1.0, since even a point-source at seeing=1.0 has FIBERFLUX < FLUX (apparently ~0.78 if using the same assumptions as Tractor).  And yes, QSOs would be better modeled as point-like, having a nearly identical distribution as stars.

moustakas:
@sbailey  @julienguy 

Apologies for the noise of the previously attached figure, which was made in haste from the airport. 

I'm fixing the calculation of fiberflux in the mock catalogs, but I need some clarity / confirmation regarding the typical fiber acceptance fraction for point sources.  Using `FastFiberAcceptance` I find a typical value (in one arcsec seeing) of ~0.66, not 0.78 (see below).  Including a simple wavelength-dependent seeing model (defining 1 arcsec at 5500 A) I find fiber acceptance fractions for point sources of 0.64, 0.68, and 0.74 in g, r, and z, respectively.  

I'll move forward with these values unless you see something obviously wrong with what I'm doing.

BTW: as I recall, fiberflux in DR7 was something of a hack, I think based on simple aperture photometry (@mlandriau can perhaps confirm), so that could be the origin of the discrepancy.

```
import os
import numpy as np
from specsim.fastfiberacceptance import FastFiberAcceptance

FFA = FastFiberAcceptance(filename=os.path.join(os.getenv('DESIMODEL'), 'data', 'throughput', 'galsim-fiber-acceptance.fits'))
FFA.value('POINT', np.atleast_1d(1.0/2.355 * 107.0 / 1.52), np.atleast_1d(0.0), hlradii=np.atleast_1d(0))
```
`Out[5]: array([0.65950679])`

sbailey:
From the [legacysurvey data model](http://legacysurvey.org/dr7/files/#sweep-7-0-sweep-brickmin-brickmax-fits), those FIBERFLUX values are based on *Gaussian* seeing, while FastFiberAcceptance is tuned for a *Moffat* profile with larger wings, thus we would expect a lower (and more accurate) FIBERFLUX.  Breaking out the ancient desimodel/fiberloss.py script, the difference you are finding wrt legacysurveys is consistent with the difference between Gaussian vs. Moffat profile seeing.

i.e. Let's proceed with this calculation, and hopefully update legacysurvey FIBERFLUX values for DR8.

sbailey:
That distribution of `FIBERFLUX_R / FLUX_R` looks much more reasonable.  For comparison here is DR7.1 data (using approximately the same color scheme, though I didn't figure out how you got those colors in that order in the legend...):
fiberflux

For the large radius objects: for now let's just put in a floor of 5% and move on.  We'll consider a better extrapolation for large radii in FastFiberAcceptance itself, but let's not make that a blocking factor for this.

moustakas:
OK, I added a floor of 5% (although note that a small fraction of large--but not too large--EXPs can have fiber fractions of <5%, which is fine), fixed a couple bugs, and did some additional sanity checks of the outputs, so I think this is ready to be merged.  

@sbailey or @geordie666 please do so if you agree as I'll be offline a chunk of the day tomorrow.

PR #382, opened 2018-09-19
Update QA to work automatically with SV and commissioning target files

geordie666:
This PR updates the desitarget QA code to work with all flavors of target files. It automatically detects whether a target file is a commissioning or SV file based on column names, and then flips the correct switches to generate QA pages for that flavor of targeting. 

This PR also includes the first unit tests of desitarget QA. 

Note that most of the noise in the commits below is finding a version of matplotlib and basemap that would mimic the versions on NERSC to pass Travis.

Example outputs are here for commissioning:
http://portal.nersc.gov/project/desi/users/adamyers/PR382CMX/

and here for (a small subset of) SV:
http://portal.nersc.gov/project/desi/users/adamyers/PR382SV/


sbailey:
Nice!  OK to merge as soon as Travis tests pass.

As you have painfully discovered, basemap is incompatible with matplotlib 2.2.2, but downgrading to matplotlib 2.1.2 works.  Ugh.

moustakas:
> As you have painfully discovered, basemap is incompatible with matplotlib 2.2.2, but downgrading to matplotlib 2.1.2 works. Ugh.

This pain was also documented in https://github.com/desihub/desiconda/issues/23....

PR #383, opened 2018-09-20
bug fixes in selection of MWS_NEARBY targets in data and MWS_WD,MWS_NEARBY,STD_WD in mocks

moustakas:
This PR fixes a small bug in how we were selecting [MWS_NEARBY targets](https://desi.lbl.gov/trac/wiki/TargetSelectionWG/TargetSelection#a100pcvolume-limitedsurveyMWS-NEARBY) in the data: we were applying a `parallax>10` cut rather than a `(parallax+parallax_err)>10` cut (which I confirmed off-list with Boris Gaensicke and the MWS WG co-chairs).

More importantly this fixes a bug in `select_mock_targets`, which was using the MWS_MAIN/Galaxia mocks to select MWS_NEARBY, MWS_WD, and (consequently) STD_WD targets, which is wrong and which led to the 1000 standards/deg2 ridiculousness (see #377).  With these changes I'm getting a much more sensible faint-star standard-star density of 90-100/deg2.

I also took this opportunity to deprecate the BOSS (SDSS color-color) selection of standard stars, which is what we were previously using as a temporary hack.

@sbailey merging this should be a high priority of the 18.9 software release, since the results are so much more sensible.  And then I'll stop.  I promise.  Well, not really.

@geordie666 please make sure I didn't mess anything up in `desitarget.cuts` if you get a chance.


moustakas:
I'm sorry, I was too hasty and merged before allowing Travis tests to complete.  I'll fix the failures in a separate PR.

PR #385, opened 2018-09-20
fix failing tests due to #383

moustakas:
Self-explanatory PR to fix the problems introducted in #383.

I got tripped up because `sv1.sv1_cuts` has an identical copy of `cuts.isMWS_nearby` that came out of sync, even though this is not a documented target class in
https://desi.lbl.gov/trac/wiki/TargetSelectionWG/SurveyValidation

I'm worried that this is going to happen more and more...

### DIFF
--- a/py/desitarget/cmx/cmx_cuts.py
+++ b/py/desitarget/cmx/cmx_cuts.py
@@ -535,7 +535,7 @@ def apply_cuts(objects, cmxdir=None):
         raise IOError
 
     # Process the Gaia inputs for target selection.
-    gaia, pmra, pmdec, parallax, parallaxovererror, gaiagmag, gaiabmag,   \
+    gaia, pmra, pmdec, parallax, parallaxovererror, parallaxerr, gaiagmag, gaiabmag,   \
       gaiarmag, gaiaaen, gaiadupsource, gaiaparamssolved, gaiabprpfactor, \
       gaiasigma5dmax, galb = _prepare_gaia(objects, colnames=colnames)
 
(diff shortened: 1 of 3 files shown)


moustakas:
@geordie666 

> Amused by your "ugh" comment: sv1_cuts has to have it own version of ismws_nearby (in some form). As does sv2_cuts (should we get that far). Otherwise there's no way to trace which version of ismws_nearby was used for SV1/SV2 etc. Populating the main cuts module with a bunch of SV cuts would get messy.

If `cuts.isMWS_nearby` and `sv1.sv1_cuts.isMWS_nearby` were truly independent then I don't understand how/why the SV unit test failed (although admittedly I didn't dig).  Before the fix in this PR, the version of this function in `cuts` took `parallaxerr` as an input, while the one in `sv1.sv1_cuts` did not, which should not have been a problem (but it was).

I'm not arguing that `cuts` needs to be polluted with a bunch of SV[n]- and/or CMX-related functions.  All I'm saying is that in so far as the selection of MWS_NEARBY targets in SV1 is *identical* to the selection in the "main" survey then we should just import those functions from `cuts` into, e.g., `sv1.sv1_cuts` to the greatest extent possible, rather than copying and pasting identical code (specifically to minimize the kind of problem I introduced, where I updated one version but not the other).  The provenance of the cuts are still maintained by git commits and tags, are they not?

I was also thrown off by the fact that `MWS_NEARBY` targets are not [documented](https://desi.lbl.gov/trac/wiki/TargetSelectionWG/SurveyValidation) SV targets, and so I didn't think to look in the `sv1` directory.

For clarity, we could consider---although I'm sure you did and decided against it---to explicitly add an `_sv1` suffix to these functions, but only when the cuts are different than those in the main survey (and even though the functions live in the `sv1.sv1_cuts` module).

Or maybe I'm fundamentally not understanding your comment:
> there's no way to trace which version of ismws_nearby was used for SV1/SV2

All that said, I happily defer to the structure you've implemented, but also happy to talk about this offlist/offline if that would be helpful or useful.

PR #386, opened 2018-09-21
rescale parameters when generating and querying kd-trees

moustakas:
This PR is a long-overdue and unadvertised improvement in how we use kd-tree to assign spectral templates to mock targets.  In essence, the kd-trees are constructed and queried using normalized and logarithmic (where appropriate) spectral properties rather than the properties themselves (which is how I should have written it in the first place).

The result is a much improved mapping between mock targets and their corresponding templates.

A well-hidden bug was also discovered and fixed when assigning templates to cool main-sequence stars.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,6 +5,8 @@ desitarget Change Log
 0.24.0 (unreleased)
 -------------------
 
+* Rescale spectral parameters when generating and querying kd-trees in
+  `select_mock_targets` [`PR #386`_].
 * Bug fixes: [`PR #383`_].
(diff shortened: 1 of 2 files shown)


moustakas:
Self-merging because I'd like to run `mpi_select_mock_targets` over a chunk of sky to help identify any remaining issues.

moustakas:
Sorry, this version of `select_mock_targets` failed at NERSC because of an unanticipated bug. I don't think @sbailey is waiting around for me but just in case this is a heads up that I'm fixing the issue now and will test at NERSC before the PR.

PR #387, opened 2018-09-22
Fix bug in code that makes unit test data

geordie666:
The code that produces the data for the unit tests (`test/make_testdata.py`) had a bug. 

This PR fixes that bug and remakes the unit test data files. I also conducted some general clean-up while I was waiting for the test data to regenerate.


geordie666:
These are minor and inconsequential changes, so I'll merge when tests pass.

PR #368, opened 2018-09-06
(another) major enhancement of select_mock_targets

moustakas:
This PR is a(nother) major enhancement of the engine underlying `select_mock_targets` (i.e., `desitarget.mock.mockmaker`) for *all* the target classes; it's not quite ready to merge but we should strive to include it in the 18.9 software release.  Also note that the changes herein depend upon https://github.com/desihub/desisim/pull/409.

Some **highlights**, in no particular order:
* Updated default Lya (london/v2.0), Gaussian random field (v0.0.8_2LPT), and Galaxia (v0.0.6) mocks.
* New `zmax_qso`, `zmin_lya` parameters in the configuration file to ensure no double-counting of QSOs (see #308).
* New `balprob` parameter in the configuration file to probabalistically insert BALs into Lya QSO spectra (DLAs and metals are still ToDo; see below).  The default is zero.
* Support for using either `desisim.templates.QSO` *or* `desisim.templates.SIMQSO` in mock.mockmaker.QSOMaker (via a new `use_simqso` parameter in the configuration file, which defaults to `True`).  However, see the notes/warnings below.
* Self-consistent treatment of "north" (BASS/MzLS) vs "south" photometry (DECaLS) in terms of synthesized (noiseless and noisy photometry), targeting bits, `PHOTSYS` value, etc., using a constant-declination cut (see #353).  
* Enhanced metadata tables for individual target classes are now written out in dedicated HDUs in the `truth.fits` file.  I'll add some notes below, but see https://github.com/desihub/desisim/issues/290 and [this comment](https://github.com/desihub/desisim/commit/569e44206aaf05edb17aa1372f997972ba36699e#commitcomment-30360004).
* Significantly more complete (but larger!) output `target.fits` catalog, which conforms almost exactly to the DR7 photometric data model, including new Gaia columns, `PRIORITY`, `NUMOBS`, etc.  `RELEASE` is set to a constant 9999 for lack of a better value.
* Application of the *same* algorithms (in `desitarget.cuts.apply_cuts`) used to select "real" targets and assign targeting bits, including updated selection of standard stars based on Gaia and Legacy Survey color cuts (except for `MWS_NEARBY` and `STD_WD`; see below).
[...]

PR #393, opened 2018-10-01
include TRUTH_ELG, TRUTH_BGS, etc HDUs

sbailey:
PR #368 moves the per-type truth entries into separate HDUs, e.g. `TRUTH_ELG`, 'TRUTH_BGS`, `TRUTH_STAR`, etc.  This PR propagates those into the final merged truth file when using `join_targets_truth`.

I don't know if this case can happen, but this merging will support the case where some input truth files are missing a particular HDU (e.g. if some but not all inputs have TRUTH_WD).

If *all* input truth files are missing a particular flavor of truth, this is treated as a warning but not an error: you might be working with truth files for a particular class but don't care about the others.


moustakas:
One Travis test failed (and I can't seem to find the "restart build" button) otherwise I'd merge this as-is.

Issue #390, opened 2018-09-26
Update mocks to make new footprints easier


sbailey:
Update our suite of mocks to make it easier to re-use them for different footprints, e.g. the updated dithering in https://desi.lbl.gov/trac/browser/code/desimodel/branches/newtiles/data/footprint/ or the upcoming update to match the final imaging footprint.

Suggestion from @schlafly: generate mocks for a superset of what we might possibly ever include, e.g. dec>-20 and |b|>10.  After that, I see two basic options:

1. provide simple code to trim the bigger mocks into mocks matched to a particular tiling.  This is similar to our current mocks whose footprint is tuned to a particular tiling, and is convenient for stuff like `efficiency = num_targets_in_zcatalog / num_targets_in_mock`.
2. always use the superset, and leave it to the final analyses to make edge corrections for the particular tiling used.

I prefer (2) because then we don't have to maintain unique versions of mocks for unique tilings, and because you'll need to do something like that when using a mock to evaluate DESI part way through the survey anyway.

Mocks currently in use by the end-to-end tests, starting from `/project/projectdirs/desi/mocks/`:
* [ ] bgs/MXXL/desi_footprint/v0.0.4/BGS.hdf5
* [ ] GaussianRandomField/v0.0.8_2LPT/ELG.fits
* [ ] GaussianRandomField/v0.0.8_2LPT/LRG.fits
* [ ] GaussianRandomField/v0.0.8_2LPT/QSO.fits
* [ ] lya_forest/london/v2.0/master.fits
* [ ] mws/galaxia/alpha/v0.0.5/healpix
* [ ] mws/100pc/v0.0.3/mock_100pc.fits
* [ ] mws/wd/v0.0.2/mock_wd.fits
* [ ] uniformsky/0.1/uniformsky-2048-0.1.fits

@moustakas and/or @forero can you shepherd this through?

Useful code:
  * `desimodel.footprint.tiles2pix` (tiles table -> healpixels covering those tiles)
  * `desimodel.footprint.is_point_in_desi` (tiles table + ra,dec arrays -> yes/no)

sbailey:
Mostly history.  I grabbed the list that are the current defaults in desitarget/doc/mock_example/input.yaml .  I don't know whether the current mock making code makes use of the `E1` and `E2` columns in the default mocks but which aren't present in the Dark Sky mocks (probably not).

[...]

PR #398, opened 2018-10-11
generate mock sky position over a larger footprint

moustakas:
Addresses the need in #390 to have mocks over a larger footprint.

The output file generated with the code snippet below has been copied to `/global/projecta/projectdirs/desi/mocks/uniformsky/0.2/uniformsky-2048-0.2.fits` and the README has been updated:
```
from desitarget.mock.sky import random_sky
ra, dec, pix = random_sky(allsky=True, outfile='uniformsky-2048-0.2.fits')
```
The catalog contains 37,748,739 entries (vs 18,335,706 previously) and extends over the full sky northward of -30 deg.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,10 +5,15 @@ desitarget Change Log
 0.24.1 (unreleased)
 -------------------
 
+* Updated mock sky catalog with positions over a larger footprint [`PR #398`_].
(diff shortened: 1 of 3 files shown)


geordie666:
Un-indenting `used` on this line should fix sphinx:

https://github.com/desihub/desitarget/blob/allsky-uniformsky/py/desitarget/mock/sky.py#L18

PR #405, opened 2018-10-16
solves crashing problem in `mtl` when called from `quicksurvey`

forero:
One liner that solves many problems downstream in `quicksurvey`.

The problem was that the new `target` file included a `NUMOBS` column, that column clashed with the `NUMOBS` column in `zcat`. Solution: delete the input `NUMOBS` in `target`.

### DIFF
--- a/py/desitarget/mtl.py
+++ b/py/desitarget/mtl.py
@@ -47,8 +47,11 @@ def make_mtl(targets, zcat=None, trim=False):
 
     n       = len(targets)
     targets = Table(targets)
+    if 'NUMOBS' in targets.keys():
+        del targets['NUMOBS'] # the relevant information coms from zcat['NUMOBS']
+    if 'PRIORITY' in targets.keys():
+        del targets['PRIORITY'] # the relevant information coms from zcat['NUMOBS']
 
-    del targets['NUMOBS'] # the relevant information coms from zcat['NUMOBS']
 
     # Create redshift catalog
     if zcat is not None:
--- a/py/desitarget/mtl.py
+++ b/py/desitarget/mtl.py
@@ -48,6 +48,8 @@ def make_mtl(targets, zcat=None, trim=False):
     n       = len(targets)
     targets = Table(targets)
 
+    del targets['NUMOBS'] # the relevant information coms from zcat['NUMOBS']
+
     # Create redshift catalog
     if zcat is not None:
 


forero:
I have the opposite opinion. Target selection cannot freeze those columns because they change in time, i.e. they don't have useful information after the first redshift catalog is computed. `NUMOBS` in zcat is the useful quantity, not `NUMOBS` in targets.

geordie666:
I agree with @moustakas. I think @sbailey and I brainstormed something in his office at one point where we might rename the initial `NUMOBS` and `PRIORITY` assigned by targeting to be `NUMOBS_INIT`, etc. Hopefully @sbailey  remembers the conversation better than I do!

forero:
OK. In that case we need to redefine the meaning and names of all columns. Usually what is in `targets` is true in absence of redshift information, `mtl` has updated data including redshift information.

forero:
If you want to calle it `NUMOBS_INIT` I am fine with that, we won't use that column anyway.

geordie666:
I think @kdawson1000 was advocating for a process where you could track `NUMOBS` at every run of mtl from the zeroth run (i.e. the targets with `NUMOBS_INIT`) to the final run (the final zcat). @kdawson1000 was also advocating writing out a mask of bits that indicated why `NUMOBS` changed with every run of mtl.

This is a targets->fiberassign coordination issue, though, so it's really in @sbailey's purview how to handle the data model at different stages.

forero:
We already have that process. Quicksurvey saves every mtl and zcat file from the zeroth run. That's what we call epochs in quicksurvey.

geordie666:
@forero: You're certainly the expert on quicksurvey, not me, and I don't want to speak for @kdawson1000. I'm just relating why I remember @kdawson1000 requesting that `NUMOBS` and `PRIORITY` be added to the targets file. It's entirely possible that everything that @kdawson1000 wanted is already in place and that including `NUMOBS` in the target file is obsolete.

forero:
@geordie666  I am sure there are other reasons to include that information in the `targets` file, I would simply prefer a different name (that would also solve the problem we had here).

geordie666:
OK, so, thinking about it some more, I think @kdawson1000's request was to set `NUMOBS` by hand (i.e. with human intervention) in the `targetmask.yaml` file:

https://github.com/desihub/desitarget/pull/405#issuecomment-430309237

rather than to calculate it by a set of rules in `calc_numobs(ztargets)`. The idea (again, I think) would be to have fine control over how many observations each targeting bit got during SV.

I think this would be a relatively simple fix, though, I hope. This line:

https://github.com/desihub/desitarget/pull/405#issuecomment-430309237
`ztargets['NUMOBS_MORE'] = np.maximum(0, calc_numobs(ztargets) - ztargets['NUMOBS'])`

would instead read:
`ztargets['NUMOBS_MORE'] = np.maximum(0, ztargets['NUMOBS_INIT'] - ztargets['NUMOBS'])`

Again, though, I'd appreciate data model feedback from @sbailey on whether `NUMOBS_INIT` and `PRIORITY_INIT` are the way to go (and this shouldn't hold up this PR).

forero:
Test pass. I am merging this, waiting for @sbailey to open another issue to discuss the issue on `NUMOBS_INIT` and `PRIORITY_INIT`.

sbailey:
Thanks, guys.  @kdawson1000 and I will talk and try to resolve this during the collaboration meeting this week.  Thanks for proceeding with this PR so that we aren't a blocking factor in the meantime.

Issue #348, opened 2018-08-22
assigning SKY and BAD_SKY targeting bits


moustakas:
* In the current version of `desitarget.skyfibers`, sky targets which have imaging data missing in *any* band are assigned the bit `BAD_SKY`, although [this notebook](https://github.com/desihub/desitarget/blob/master/doc/nb/skies-dr7.1.ipynb) shows that most of these are probably fine.  Consequently, consider instead requiring `APFLUX_IVAR_[G,R,Z] == 0` in *all* bands for "bad" skies.  (May be worth also checking to make sure the cut on ivar also captures targets whose *flux* is identically zero -- it should!)

* Related: in the current version of the code, sky targets which have zero flux *and* zero inverse variance still have the `SKY` or `BAD_SKY` targeting bits set (about 4% of the dr7.1 skies).  These should really be excluded from the sky catalog, since presumably we're missing the information we need to assess these as viable sky targets (whether `SKY` or `BAD_SKY`).

geordie666:
A couple of notes:

- Talking with Schlegel offline, a better criterion might be `APFLUX_IVAR_[G,R] == 0` in *both* of `g` and `r` (i.e. place less trust in `z`). Agreed about explicitly checking *flux* as well as *ivar*.

- My understanding of `BAD_SKY` is that it is meant to indicate a location where we can put a fiber, but which the spectroscopic pipeline should not (necessarily) trust for sky subtraction. I think the 4% of the dr7.1 skies that you're talking about should therefore *not* be excluded from the sky catalog and *should* be labeled `BAD_SKY`. I think these are the very definition of locations that we need but that shouldn't (necessarily) be trusted for sky subtraction. If we discard these sorts of locations from the sky catalog, there will be places where we can't place a fiber. I will match these to Gaia to guard against having substantial flux at these "gap" locations.

geordie666:
Addressed in https://github.com/desihub/desitarget/pull/408.

Issue #399, opened 2018-10-11
Give WDs higher priority than BGS galaxies 


apcooper:
At the Tucson meeting, BGS and MWS agreed WDs in the science sample (not just WD standards) should be prioritised above BGS (e.g. at priority 3000). This doesn't seem to be implemented yet:

https://github.com/desihub/desitarget/blob/9f7a2571e073ae9e1b7b8eaeae02294247419a7f/py/desitarget/data/targetmask.yaml#L210-L213

geordie666:
Good catch...I'll add this in my next iteration of the code (early next week).

geordie666:
Fixed in https://github.com/desihub/desitarget/pull/408.

PR #407, opened 2018-10-18
Refactor BGS selection to separate masking from color cuts

geordie666:
This PR addresses issue https://github.com/desihub/desitarget/issues/406 by splitting the masking part of the BGS selections from the color cuts.

While I was refactoring the BGS algorithms, I worked through the BGS functions to remove any redundant code. Hopefully this will make the code easier to maintain.

PR #408, opened 2018-10-22
Updates to the ELG and MWS target classes and the selection of sky locations

geordie666:
Modify several target classes, mostly to reflect changes on the main survey targeting wiki. Includes:

- Refactor of the `ELG` and `MWS_MAIN` selection algorithms to better split color cuts/masking and north/south.
- Update of the `ELG` and `MWS_MAIN` selection cuts. In particular, the ELG masking logic now requires `BRIGHTSTARINBLOB`.
- Change `MWS_WD` targets to have a priority higher than all of the `BGS` target classes.
- Set sky locations to `BAD` only if *both* of the g-band and the r-band are missing.

Should fix issues https://github.com/desihub/desitarget/issues/348 and https://github.com/desihub/desitarget/issues/399.


moustakas:
Can we get clarification from the Targeting WG about the use of cuts on `allmask_[g,r,z]` for "north" ELGs but not for "south" ELGs?  Actually, it looks like the code applies the `allmask` to both north and south ELG targets, which doesn't quite match the wiki.

Also, for the sake of clarity I wonder if we can combine the g- and r-band magnitude cuts into *one* magnitude cut and a color cut.  E.g., in the north we have `21.0 > g < 23.7` and `r < 23.3`.  For many applications it's helpful to think about the sample as either being g- or r-band limited (in the FDR it was a g-band limit).  The results are of course the same but the added clarity is useful.

sbailey:
Thanks @geordie666 for the updates and @moustakas thanks for the review.  @moustakas can you also confirm that this won't break select_mock_targets?  We have a historical fragility of changes to the "real data" cuts interfaces breaking mocks in ways that aren't caught by the simple unit tests.  Thanks.

geordie666:
@moustakas: I'm done with this branch. If you get a chance, can you check that it doesn't break `select_mock_targets` and if it doesn't, merge it? 

Hopefully I won't have to ask that every time I update the "real data" cuts. My next effort is working on SV and secondary targets, so you should get a break from me pestering you for a while.

moustakas:
Actually, it looks like #407 broke `select_mock_targets` in that the BGS selection now requires `FRACIN_[G,R,Z]`, which we'll need to add to the targets data model---
https://desidatamodel.readthedocs.io/en/latest/DESI_TARGET/targets.html

PR #409, opened 2018-10-23
Changing prioritization scheme for Lya-QSOs

fjaviersanchez:
@andreufont suggested during last week's collaboration meeting to change the prioritization scheme for the QSOs. Before, the possible Ly-a QSOs were all pushed to priority=3500. With this code, the priority is set as a function of their "cosmological value" as predicted by the Fisher studies included in a table provided in this PR. @sbailey suggested to change `mtl.py` but I thought it was cleaner to change `targets.py` and include an external script for the Ly-alpha case. 

Please, let me know if this is ok or you would rather prefer to change the code in a different way.


sbailey:
@fjaviersanchez I think we misunderstood each other last week.  I agree with @geordie666 that MTL should not be changing priorities based upon target properties like FLUX_R.  We used to do that for LRGs and eventually decided that was a bad idea and gave them separate bits for 1-pass vs. 2-pass LRGs.  This is a similar case.

I still support the idea of allowing an external catalog to override classification and redshift (e.g. from QuasarNET or visual inspections).  I *think* I still support the option for that external catalog to also override priorities, which could be used for the case you are implementing here.  But I don't think we should be changing the *automatic* prioritization within MTL.  The fundamental philosophy is that if targets are to be treated differently, they should receive different bits, such that MTL changes priorities and NUMOBS_MORE only based upon bits, observing history, and optionally Z and ZWARN, but not other target properties like FLUX.

fjaviersanchez:
@geordie666 and @sbailey thanks a lot! Yes, this code is intended to pre-select the most valuable QSOs for the second pass (where we downstream from ~250 per exposure to ~50 if I understood correctly?). I could get rid of the dependence in `FLUX` (integrating the value over a range of magnitudes) and leave the dependence in redshift only. I am happy to close the PR without merging if there's no agreement and/or to migrate this code elsewhere. Please, let me know what is preferable. Thanks!

andreufont:
Hi @sbailey @geordie666, could someone from data systems call tomorrow to the Lyman alpha telecon? 9am Pacific.

Even if only for 5-10 minutes to clarify this issue would be useful!

sbailey:
@andreufont sure I'll join the LyA call tomorrow to discuss and clarify.

andreufont:
Hi @fjaviersanchez @geordie666 @sbailey 

Let me try to summarize here what Stephen and I discussed during the Lyman alpha telecon on Wednesday:
- At the end of each observing run, redrock will be run automaticaly. Unless the Lya WG does anything, this classification will be used to select the targets that will reobserved in the next pass. The result is written in a file called the "merged target list".
- However, the plan is that the Lya WG will run their own code(s) and generate a file (with a well documented, standarized format) that will be read by the pipeline to improve the merged target list. 
- Ideally, all code used in this step would sit in desihub. Each "quasar classifiers" should have its own repository, and we will have another meta-repository that will combine the information from the classifiers and talk to the main pipeline to update the merged target list.
- We should then delete this pull request, and move the new code from Javi to the new meta-repository.

Who should start the new meta-repo? How should we call it? For now, I would just put there placeholders for SQUEEzE / QuasarNET, and write the code to combine them and to add quasar value.

geordie666:
@andreufont: In partial answer to some of your questions:

- Only @sbailey, @weaverba137 and select others (who I can't remember) can add a repo to desihub, so one of them will have to set up a new repo for you at some point.
- I think the goal should be to have a FITS file that includes the `TARGETID` of a target (which is a unique value for each target) and the new priority of that target. I believe only those two columns will be necessary.
- I will then need to include code in MTL in the `desitarget` repo that matches to the existing target on `TARGETID` and overwrites the `PRIORITY`. This last item is just a reminder to me that this hook will be necessary.

moustakas:
I can make the repo if `qso-finder` is the name you've settled upon.

andreufont:
I'm not good at naming, so I'll let others with stronger opinions to decide :-)

It would be great to avoid confusion with target selection, that's why I didn't use something like lya-qso-selection... 

lya-qso-reobserve?

andreufont:
Hi @sbailey @moustakas @fjaviersanchez - Should we start this repository?

moustakas:
As requested, I've created [qso-finder](https://github.com/desihub/qso-finder) and initialized its contents with the [desitemplate](https://github.com/desihub/desitemplate) structure.  However, the Lya WG should feel free to customize the repo to fit their needs.

I believe I've set the permissions correctly but if you have any issues please, well, open an issue!

geordie666:
@fjaviersanchez, @andreufont: Can we delete this PR and the associated branch now that the work is occurring in https://github.com/desihub/qso-finder?

fjaviersanchez:
I'm also deleting the branch

PR #417, opened 2018-11-06
SJB travis

sbailey:
Don't merge this PR.  I'm testing Travis options for healpy and conda channels.  Apologies for the spam that github will generate.


sbailey:
Installing healpy from conda-forge worked, but as @tskisner noted in #416 this has the side effect of getting nearly all packages from conda-forge.

Now I'm trying switching back to the astropy channel and `pip install --no-binary :all: healpy` as a separate install command instead of lumping it in with fitsio and coverage, for which the binary packages may be ok.

PR #421, opened 2018-11-13
fix install for cmx and sv1

sbailey:
This PR fixes `python setup.py install` for desitarget cmx and sv1 directories:

* `desitarget.sv1` wasn't installed at all because it was missing an `__init__.py` file (ah, python, you used to be so simple...)
* The `data/` directories for both cmx and sv1 were not installed because they weren't included in the `setup.py` file (it auto-discovered python directories, but not data directories).

i.e. desitarget was working fine for "bare" installations of just adding directories to $PATH and $PYTHONPATH, but not for full installations.  I fixed this by hand for the 18.11 software release (desitarget/0.25.0).

Asking @weaverba137 for review about whether I implemented these fixes correctly.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,6 +5,7 @@ desitarget Change Log
 0.26.0 (unreleased)
 -------------------
 
+* Fix `python setup.py install` for cmx and sv1 directories [`PR #421`_].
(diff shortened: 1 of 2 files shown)


weaverba137:
This looks fine to me.  In particular, the declaration of the data directories is fine.

Issue #419, opened 2018-11-12
MWS-BROAD selection needs a colour cut


apcooper:
@geordie666, a minor correction to the changes in respect of #410. Not urgent, can be fixed in the next round of updates.

The MWS-BROAD targets should also be limited to g-r > 0.7, like the MAIN-RED targets. We want BROAD to be the complement of the astrometric cuts that only apply to RED. We don't apply any of the astrometry cuts to BLUE in the first place, so we don't want all the blue targets that fail those cuts to be counted as BROAD as well as BLUE.

However, it looks like the code added in #418 doesn't include that colour cut for BROAD:

https://github.com/desihub/desitarget/blob/eb1c2d20c912eb38053994aeff68c89d6c57af2d/py/desitarget/cuts.py#L686-L693

Rather than`broad = mws.copy()` I think this should have `broad = red.copy()` in the same place (i.e. before the astrometry cuts are applied to `red`).

This is what we wrote on the TS page, but sorry I didn't catch it in the PR.

Issue #420, opened 2018-11-12
PSF type requirement missing in MWS-MAIN cuts


apcooper:
Sorry @geordie666 , here's another one I only just noticed. The MWS-MAIN-* and MWS-BROAD cuts on the TS page specify `TYPE==PSF`, but I can't see that implemented in `desitarget` (the MWS selection routines don't take `objtype`?).

geordie666:
Fixed in https://github.com/desihub/desitarget/pull/422.

PR #422, opened 2018-11-15
Fix bugs in MWS main selections.

geordie666:
This PR updates the main MWS selection code to fix https://github.com/desihub/desitarget/issues/419 and https://github.com/desihub/desitarget/issues/420.


apcooper:
All looks good to me, thanks for these changes @geordie666.

moustakas:
@geordie666 Do you mind merging this (or I can) so I can get the changes into a new PR of my own?

moustakas:
Assuming I'm not misreading the code, the issue is that I don't think the north/south splits are being handled correctly.  I noticed this specifically for MWS because the `MWS_RED_S` and `MWS_RED_N` targeting bits (in a set of mocks) were being set *for the same targets*.

For example, `isMWS_main_colors` takes a `south=True` argument [here](https://github.com/desihub/desitarget/blob/ADMfixMWS/py/desitarget/cuts.py#L649-L652) but the code doesn't actually do anything with that information.

In addition, when these targeting bits get set [here](https://github.com/desihub/desitarget/blob/ADMfixMWS/py/desitarget/cuts.py#L1958-L1972) in a `for south in [False, True]:` loop, the code passes *all* the same set of photometry (i.e., the `gflux`, `rflux`, `zflux`, etc.), instead of indexing the north / south photometry separately.  

And in fact it looks like all the target classes are affected by the same problem (except maybe the random-forest QSOs).

Finally note that `select_mock_targets` contends with the same issue, which I'll now double-check to make sure the various "Maker" classes in `mock/mockmaker` are handling it correctly.

Finally finally, feel free to fix this in a separate PR; there's no urgency on my end.

PR #423, opened 2018-11-21
Added two new QSO RF trained over DR7 and updated 'isQSO_randomforest…

CAClaveau:
…_north/south' functions


sbailey:
Assigned @geordie666 for review.  Note that this week is the Thanksgiving holiday in the US so he may not get to it until next week.

geordie666:
Thanks @CAClaveau. This PR looks fine, and I'm going to merge it soon.

@Cyeche: We're once again increasing the size of the repo by adding the new Forests but I'm not going to let this be a blocking factor (particularly as we now have over 20M in notebooks in the `doc/nb` directory, alone!). But, let me know if/when I can remove the DR3 and DR5 versions of the Forests to free up some space.

Issue #401, opened 2018-10-13
add contaminants to select_mock_targets


moustakas:
Including contaminants in `bin/select_mock_targets` is firmly on the critical path, not only to help train Lya QSO identification algorithms but also to help with various fiber-assignment investigations (e.g., wide vs deep).

[Note that I'm creating this new issue in favor of #285, which I intend to close, to more clearly state the problem and the path forward.]

We expect each target class to be "contaminated" with other types of objects that live in that part of color-magnitude space.  The spatial distribution of extragalactic contaminants should be basically random, whereas the level of stellar contamination should depend on Galactic latitude.

Based on various estimates (which may not be the most accurate of up to date!), I roughly expect:

**BGS** - Here we expect just stellar contamination, of order a few percent, which will vary systematically with Galactic latitude.

**ELGs** - Stellar contamination should be of order 5% (see, e.g., [elg-properties-dr7.1.ipynb](https://github.com/desihub/desitarget/blob/master/doc/nb/elg-properties-dr7.1.ipynb)), while contamination from lower-redshift interlopers should be of order 10%-15%.

**LRGs** - Principle contaminants should be M stars (conservative maximum of 5%), while contamination from extragalactic targets should be negligible.

**QSOs** - Contamination from intermediate-redshift galaxies should (very roughly) be about 10%, while stellar contamination should be approximately 25%.  (Please see [SimulateQuasarTargets](https://desi.lbl.gov/trac/wiki/LymanAlphaWG/SimulateQuasarTargets) for a detailed discussion of QSO contaminants.)

To do this right we really want the apparent magnitude distribution, redshift distribution, and space-density (and morphologies!) of the extragalactic contaminants, as well as dedicated templates.  However, the goal is to implement something simple and then improve from there.

For the extragalactic contaminants I intend to use the Buzzard mock (see #249), which, like the BGS/MXXL mock, is flux-limited, but independent of the current default BGS mock.

Please post any additional suggestions and thoughts here.

akrolewski:
I notice in the [targeting QA](http://portal.nersc.gov/project/desi/users/adamyers/desitargetQA-dr6.0-0.22.0/) that there there are ELG/QSO overlaps, suggesting that some of the QSO contaminants are ELGs (and vice versa).

Are these overlaps reflective of reality, or are they merely an artifact of the current targeting scheme and are expected to go away by the final release of imaging/targeting?

[...]

Issue #284, opened 2018-02-27
include Milky Way extinction in mock spectra


sbailey:
Include Milky Way extinction in generated mock spectra.

moustakas:
The choice of Galactic extinction curve (and R_V), whatever it is, should be consistent with the pipeline -- see https://github.com/desihub/desispec/issues/513.

PR #425, opened 2018-11-29
use pre-computed template photometry for stars and WDs

moustakas:
Work-in-progress PR to use the pre-computed template photometry added in https://github.com/desihub/desisim/pull/453 when calling `select_mock_targets` with the `--no-spectra` option.  The speed-up is about a factor of 10(!) (although this is a one-time gain when the code is first called because the photometry is cached).

The code changes here use the as-yet unavailable `v3.1` basis templates, but the changes *are* backwards-compatible with the current `v3.0` templates.

Next step (in a separate PR) is to use these pre-computed colors to incorporate contaminants (#401).

### DIFF
--- a/py/desitarget/mock/mockmaker.py
+++ b/py/desitarget/mock/mockmaker.py
@@ -3908,38 +3908,61 @@ class STARMaker(SelectTargets):
         self.meta = self.template_maker.basemeta
 
         # Pre-compute normalized synthetic photometry for the full set of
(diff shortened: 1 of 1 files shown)


moustakas:
I took advantage of https://github.com/desihub/desiutil/pull/128 to add MW dust extinction to the spectra written out by `select_mock_targets`, addressing #284.

All other changes here are backwards-compatible, so I intend to merge once tests have passed.

PR #426, opened 2018-12-03
bug fix in N/S targeting bits

moustakas:
@geordie666 As a follow-up to my comment [here](https://github.com/desihub/desitarget/pull/422#issuecomment-442106426), I'm fairly certain that given how `photsys_n` and `photsys_s` are being defined, the changes proposed in this PR are required in order to get the various `_SOUTH` and `_NORTH` targeting bits set correctly.  

Thanks in advance for taking a look.

### DIFF
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -2023,14 +2023,14 @@ def set_target_bits(photsys_north, photsys_south, obs_rflux,
     mws_red = (mws_red_n & photsys_north) | (mws_red_s & photsys_south)
 
     # Construct the targetflag bits for DECaLS (i.e. South).
-    desi_target = lrg_south * desi_mask.LRG_SOUTH
-    desi_target |= elg_south * desi_mask.ELG_SOUTH
-    desi_target |= qso_south * desi_mask.QSO_SOUTH
+    desi_target = (lrg_south & photsys_south) * desi_mask.LRG_SOUTH
+    desi_target |= (elg_south & photsys_south) * desi_mask.ELG_SOUTH
+    desi_target |= (qso_south & photsys_south) * desi_mask.QSO_SOUTH
 
     # Construct the targetflag bits for MzLS and BASS (i.e. North).
-    desi_target |= lrg_north * desi_mask.LRG_NORTH
-    desi_target |= elg_north * desi_mask.ELG_NORTH
-    desi_target |= qso_north * desi_mask.QSO_NORTH
+    desi_target |= (lrg_north & photsys_north) * desi_mask.LRG_NORTH
+    desi_target |= (elg_north & photsys_north) * desi_mask.ELG_NORTH
+    desi_target |= (qso_north & photsys_north) * desi_mask.QSO_NORTH
 
     # Construct the targetflag bits combining north and south.
     desi_target |= lrg * desi_mask.LRG
@@ -2038,11 +2038,11 @@ def set_target_bits(photsys_north, photsys_south, obs_rflux,
     desi_target |= qso * desi_mask.QSO
 
     # ADM add the per-pass information in the south...
-    desi_target |= lrg1pass_south * desi_mask.LRG_1PASS_SOUTH
-    desi_target |= lrg2pass_south * desi_mask.LRG_2PASS_SOUTH
+    desi_target |= (lrg1pass_south & photsys_south) * desi_mask.LRG_1PASS_SOUTH
+    desi_target |= (lrg2pass_south & photsys_south) * desi_mask.LRG_2PASS_SOUTH
     # ADM ...the north...
-    desi_target |= lrg1pass_north * desi_mask.LRG_1PASS_NORTH
-    desi_target |= lrg2pass_north * desi_mask.LRG_2PASS_NORTH
+    desi_target |= (lrg1pass_north & photsys_north) * desi_mask.LRG_1PASS_NORTH
+    desi_target |= (lrg2pass_north & photsys_north) * desi_mask.LRG_2PASS_NORTH
     # ADM ...and combined.
     desi_target |= lrg1pass * desi_mask.LRG_1PASS
     desi_target |= lrg2pass * desi_mask.LRG_2PASS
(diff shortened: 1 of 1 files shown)


geordie666:
@moustakes: I see your point, now.

The background for this is that the N/S bits are not meant to be used to assign fibers. They're meant to be informational. All of the other targeting bits are set as you suggest...for instance the setting of the `1PASS` and `2PASS` LRG bits has exactly the structure you used in this PR:

https://github.com/desihub/desitarget/blob/master/py/desitarget/cuts.py#L1862-L1863

The original intent was that, in overlap regions, targeting folks could check which objects from imaging in the South region passed their North selection and vice versa. In other words the "SOUTH" bit wasn't intended to mean "is a target in the south" but rather "is a target that passes the southern set of imaging cuts".

On reflection, I agree that this is potentially dangerous, though. It leads to there being targets in the targeting file on which we should never place a fiber. I'm going to patch in @sbailey and ask for the behavior that he wants downstream. The options:

(1) We retain the current structure. There are potentially targets in the targeting files that have the N/S bits set but that do not have the main (`LRG`, `QSO`, `ELG` etc.) bits set. We should ensure that those end up with `NUMOBS_INIT=0` so that they can easily be trimmed in MTL. The benefit is that targeting folks can see which objects in the N/S would have been selected in their S/N set of cuts.

(2) We switch over to what @moustakas is suggesting here. The benefit is that it's cleaner and that we (eventually) have smaller target files. This currently isn't affecting the size of the target files for the real data as we have yet to pass a sweeps file through targeting that had mixed N/S imaging and different N/S cuts defined (I think).

A second benefit of (2) is that we can also speed the code up by a factor of 2 by not passing N/S imaging to the S/N cuts in the first place.

moustakas:
Thanks for your input, @geordie666.  It would be very helpful for the updates I'm making to `select_mock_targets` if we could resolve this issue one way or another before `18.12` ships.

For the record, I "discovered" this behavior because I had a small number of stars-as-QSO-contaminants that had the `QSO_NORTH` bit set (`2**10=1024`) but *not* `QSO` bit (`2**2=4`).  So when making the QAplots these objects were being dropped.  

And, more importantly, these were being counted (correctly) as contaminants by `select_mock_targets` (which stops generating spectra when the desired target density has been reached) but anyone computing the target density form the output `targets.fits` file would have dropped these objects (since they were `QSO_NORTH` but not `QSO` targets), and therefore achieved a lower-than-advertised target density.

Now, of course, I'm dealing with the mock catalogs here, so maybe we just need to cope, but I agree that your point (2) is much clearer from an end-user standpoint.

geordie666:
@moustakas: I should point out that I'm not opposed to working towards option (2) at all. And we can always hack in a flag to retain all of the targeting bits if option (1) becomes necessary. I think that @sbailey professed a preference towards option (1) at some point in the past, though, which is why I'd like to hear from him.

If we decide to retain the option (1) framework, then I hope it would fix the mocks just to have a trim step before calculating the target density where bits that contain only the words NORTH/SOUTH are dropped from the calculation.

sbailey:
Adding thoughts but not a decision (yet), using `LRG` as an example but concepts apply to all classes with different N/S cuts and bits:

Priorities, numobs, and fiber assignment should only be based upon `LRG`, not `LRG_NORTH` and `LRG_SOUTH`.  So for downstream *pipeline* code it doesn't matter which definition is used for `LRG_NORTH/SOUTH`.  The choice can be made by someone who would actually use those bits, e.g. @djschlegel @kdawson1000 @Cyeche et al.  Reaching out to them off github might be more effective.

Repeating @geordie666 options 1 & 2:

> (1) We retain the current structure. There are potentially targets in the targeting files that have the N/S bits set but that do not have the main (LRG, QSO, ELG etc.) bits set. We should ensure that those end up with NUMOBS_INIT=0 so that they can easily be trimmed in MTL. The benefit is that targeting folks can see which objects in the N/S would have been selected in their S/N set of cuts.

I agree with @geordie666 that this is what we had originally discussed and decided upon, thinking it would facilitate systematics comparisons of north vs. south imaging and cuts, especially in the overlap region, and if you don't care about that (e.g. fiberassign) you just use the normal `LRG` bit.  If we drop this option, we need to have a description of what someone should do to make this study.

Note: a full overlap study would need all 4 combinations of (N,S cuts) applied to (N,S photometry) for targets in the overlap region.  I believe that the current DR8 plan is for the sweeps to have a hard dec cut for N/S photometry such that the overlap objects aren't double listed, and thus these targeting bits only address half the question anyway.

> (2) We switch over to what @moustakas is suggesting here. The benefit is that it's cleaner and that we (eventually) have smaller target files. This currently isn't affecting the size of the target files for the real data as we have yet to pass a sweeps file through targeting that had mixed N/S imaging and different N/S cuts defined (I think).

I agree that this is simpler, but but it is also so simple that it begs the question of whether the LRG_NORTH and LRG_SOUTH bits are useful anymore, since they become redundant with `LRG_NORTH = LRG and PHOTSYS='N'` and `LRG_SOUTH = LRG and PHOTSYS='S'`.  Do we even need extra NORTH/SOUTH bits in this case, or does it just become
```
LRG = (PHOTSYS='N' and {northern cuts}) or (PHOTSYS='S' and {southern cuts})
```
and if you want to distinguish them you just cut on PHOTSYS and proceed (which is e.g. what the flux calibration is doing anyway to pick which filter to integrate the stdstar models in).

moustakas:
OK, thank you for the more detailed discussion and details.  I'm happy to withdraw this PR (after a few days, in case others want to chime in) and make the appropriate changes to `select_mock_targets`, although it's clear that we'll have some work to do on the real target catalogs in the overlap regions, post-DR8.

geordie666:
I agree that we should probably close this and revisit it later, if needed.

I believe that if the imaging surveys provide "special" catalogs in the overlap region that contain both northern and southern objects with the `RELEASE` number reflecting which are northern objects and which are southern objects, then the current version of desitarget should be able to produce all combinations of N/S targets and N/S imaging based on setting `PHOTSYS="N/S"` for the imaging and the `NORTH/SOUTH` bits for the cuts.

But, we'll definitely need to check that plays out if we ever have files that contain multiple sources from overlapping imaging.

moustakas:
OK, sticking with the current definition of the various `NORTH/SOUTH` bits and since there's no concept of an "overlap" region in the mocks, I implemented the following "fix" in `select_mock_targets` (in PR #427) which just subtracts out / removes any/all of the `NORTH/SOUTH` targeting bits.  (Otherwise, the output targeting catalog will have the wrong target density, since there will be targets with, e.g., `QSO_NORTH` set but *not* `QSO`).

Comments are welcome.

https://github.com/desihub/desitarget/blob/d8d3662a6f45981247b2544c0d106e82bbd12e3b/py/desitarget/mock/mockmaker.py#L1184-L1211

sbailey:
> Otherwise, the output targeting catalog will have the wrong target density, since there will be targets with, e.g., QSO_NORTH set but not QSO

@moustakas I'm not sure I understand your point.  Target density calculations should be based upon ELG/LRG/QSO bits, not on QSO_SOUTH, QSO_NORTH, etc.  i.e. it shouldn't matter if there are targets with QSO_SOUTH but not QSO set for the purposes of target density QA.  It might be a *different* QA to check the density of ((QSO_SOUTH or QSO_NORTH) and not QSO), etc. but the presence of that class of (non-)targets shouldn't impact fiber assignment and thus shouldn't be included in standard density calculations.  i.e. maybe the target QA needs to be fixed, not select_mock_targets.  Or am I missing something?

I think we've converged on not merging this PR, but I'll leave it open for the moment until the conversation has converged about what the other code should do in the meantime.

moustakas:
> @moustakas I'm not sure I understand your point. Target density calculations should be based upon ELG/LRG/QSO bits, not on QSO_SOUTH, QSO_NORTH, etc. i.e. it shouldn't matter if there are targets with QSO_SOUTH but not QSO set for the purposes of target density QA. It might be a different QA to check the density of ((QSO_SOUTH or QSO_NORTH) and not QSO), etc. but the presence of that class of (non-)targets shouldn't impact fiber assignment and thus shouldn't be included in standard density calculations. i.e. maybe the target QA needs to be fixed, not select_mock_targets. Or am I missing something?

Apologies for the noise; this statement was just about the simplicity of my code.  Previously, after generating candidate targets and applying the targeting cuts I was keeping any object that had `DESI_TARGET != 0`.  I can still do this in `select_mock_targets` but only after I've subtracted out the various `NORTH/SOUTH` bits as in my [previous comment](https://github.com/desihub/desitarget/pull/426#issuecomment-444834919).

Let's close away!

geordie666:
@moustakas: I think this

PR #427, opened 2018-12-03
include (stellar) contaminants in select_mock_targets

moustakas:
Work in progress PR to include contaminants in `select_mock_targets` target and truth catalogs (addressing #401).  Currently only stellar contaminants are included, while extragalactic contaminants may be spun off into a different PR.

More details of the assumptions and output to follow but I wanted to open this to be sure it gets into [18.12](https://github.com/orgs/desihub/projects/9).

No need for a review yet.


doc/mock_example/input.yaml
```diff
@@ -1,10 +1,41 @@
...
+            stars: 0.05,
+            galaxies: 0.1,
+        }
+        LRG: {
+            stars: 0.04,
+        }
+        QSO: {
+            stars: 0.25,
+            galaxies: 0.1,
+        }
```
sbailey: Please document with a comment in the yaml file what these numbers are.

Are these fractions of targets to treat as contaminants of the particular types?  If so, should that concept be owned somewhere in desimodel/data/targets/ instead along with densities and n(z)?  I recognize that updates to desimodel are more of a hassle, so it could be reasonable to keep contaminant fractions here until our understanding has stabilized, and then move to desimodel.  For consideration.

moustakas: These are indeed fractions.  I agree with the comment in principle, but until we have a set of mocks that are specifically for contaminants anything I do is going to be a hack.  I propose we move forward with what I'm implementing---which we'll be able to evaluate much more easily in the past because I'm adding more QA to this PR---and then go from there.

moustakas:
Note: In https://github.com/desihub/desitarget/pull/427/commits/141d8aa583b8c2a4aac5092990e26cc8eee97c90 I implemented a simple polynomial model to relate the apparent magnitude of ELGs to the [OII] flux (figure below), which is used when running `select_mock_targets` with the `--no-spectra` option.  This change removed the final dependence of this code on the [gmm-quicksurvey.ipynb](https://nbviewer.jupyter.org/github/desihub/desitarget/blob/master/doc/nb/gmm-quicksurvey.ipynb) Gaussian mixture models.

All target classes now use the more complete set of [gmm-dr7.ipynb](https://nbviewer.jupyter.org/github/desihub/desitarget/blob/master/doc/nb/gmm-dr7.ipynb) mixture models, which include not only magnitudes and colors, but also morphologies *and* uncertainties on those quantities.

rmag-vs-oiiflux

moustakas:
@sbailey I need to make sure I haven't broken the official QA code with my changes, but `select_mock_targets` itself *should* be ready for integration testing.  I haven't signed off on this PR yet until my tests complete, but if you identify any problems please ping me on slack.

moustakas:
I've confirmed that I have I haven't broken QA for either commissioning or SV via the incantations (thanks to @geordie666):
`run_target_qa /global/cscratch1/sd/adamyers/cmx-test-dr7.1.fits . --nosystematics`
or
`run_target_qa /global/cscratch1/sd/adamyers/sv1-test-dr7.1.fits . --tcnames QSO --nosystematics`

sbailey:
Running mpi_select_mock_targets some pixels failed with
```
ERROR:mpi_select_mock_targets:176:<module>: Pixels [5283] failed after 3.0 minutes
ERROR:mpi_select_mock_targets:179:<module>: Traceback (most recent call last):
  File "/global/cscratch1/sd/sjbailey/desi/code/desitarget/bin/mpi_select_mock_targets", line 170, in <module>
    healpixels=rankpix[i:i+n], no_spectra=args.no_spectra)
  File "/global/cscratch1/sd/sjbailey/desi/code/desitarget/py/desitarget/mock/build.py", line 872, in targets_truth
    MakeMock=AllMakeMock[ii])
  File "/global/cscratch1/sd/sjbailey/desi/code/desitarget/py/desitarget/mock/build.py", line 171, in read_mock
    nside_galaxia=nside_galaxia, mock_density=mock_density)
  File "/global/cscratch1/sd/sjbailey/desi/code/desitarget/py/desitarget/mock/mockmaker.py", line 3753, in read
    mock_density=mock_density, seed=self.seed)
  File "/global/cscratch1/sd/sjbailey/desi/code/desitarget/py/desitarget/mock/mockmaker.py", line 1407, in readmock
    oiiflux[ii] += rand.normal(loc=0, scale=sigma)
  File "mtrand.pyx", line 1646, in mtrand.RandomState.normal
ValueError: scale < 0
```

In `/global/cscratch1/sd/sjbailey/desi/dev/end2end/targets`:
```
srun -A desi -N 15 -n 30 -c 16 -C haswell -t 01:00:00 --qos interactive mpi_select_mock_targets --output_dir $(pwd) --config ./select-mock-targets.yaml --seed 1 --nproc 16 --nside 64 --tiles ../survey/test-tiles.fits
```
(though it would be more efficient to just run pixel 5283 with seed 1 to test this...)

moustakas:
> Running mpi_select_mock_targets some pixels failed with

Yes, sorry, this was fixed in https://github.com/desihub/desitarget/pull/427/commits/7161b2e5e8729534f2c54b27c7a5723e4beba311.  If you could rerun all the pixels that'd be helpful, as I also identified a bug in the white dwarf fluxes.  

I'm running ~50 deg2 on my laptop now and haven't seen any additional problems.

moustakas:
Some benchmarks:  Testing this version of `select_mock_targets` with `--no-spectra` on my laptop over roughly 55 deg2 (~10 tiles or 16 `nside=32` healpixels) with 4 cores and all target classes took approximately 111 min *including stellar and extragalactic contaminants*.  Naively, then, that's approximately 30 seconds per square degree per core, which doesn't seem too bad.

Once this is merged we should test `quicksurvey` to quantify exactly how well the current code scales.

moustakas:
I've generated some additional QA with this code which reveals many of the improvements and remaining shortcomings (but which should not hold up this PR):

* One `nside=64` healpixel with spectra: http://portal.nersc.gov/project/desi/users/ioannis/mock-targets/PR427-spectra-onepixel/
* `16` healpixels (`~50 deg2`) without spectra: http://portal.nersc.gov/project/desi/users/ioannis/mock-targets/PR427-nospectra-50deg2/
* All of `DR7.1` (to make sure I didn't break anything critical): http://portal.nersc.gov/project/desi/users/ioannis/mock-targets/PR427-dr7.1/

sbailey:
Thanks.  I forgot to git pull before running my tests this morning.  I confirm that it works for the minitest, albeit ~60% slower than before (comparing the same pixels to the 18.11 reference run).

I disagree with your speed calculation units though: I think you want (111*4)/55 = 8 core-minutes per deg2 instead of 111/55/4 = 30 sec/deg2/core.  e.g. consider the case where you had twice as many cores but it still took the same amount of time.

I'm getting *much* worse performance at NERSC than you report when using 15 nodes = 480 cores in parallel on nside=64 pixels so I'll poke at that a bit more, but it doesn't have to hold up merging.

Issue #415, opened 2018-11-03
MTL: use less memory


sbailey:
Refactor `desitarget.mtl.make_mtl` to use less memory.  The current code doesn't appear to be viable for running a full-footprint targets -> MTL.

For example, `/project/projectdirs/desi/target/catalogs/dr7.1/PR372/targets-dr7.1-PR372.fits` is 29GB and only covers part of the footprint, but running `make_mtl` on that spikes at 211 GB of memory (!), which means that we can't run it on a compute node and we can only run it on a login node if we are lucky and don't get caught.

For starters, using fitsio instead of astropy Tables would probably help a lot, though be careful when adding columns because it could trigger a memory copy.  Or we may need to do it in batches by healpix (or similar) and then re-assemble at the end.  We may also need to revisit the plan of propagating all input target columns into the output MTL.

moustakas:
I also think we should take another hard look at what we really need in the targets catalog--
https://desidatamodel.readthedocs.io/en/latest/DESI_TARGET/targets.html

See also https://github.com/desihub/desitarget/issues/197.

sbailey:
Fixed in PR #429.  @moustakas question about what we really need in the output catalog is covered by issue #197 so closing this one.

PR #429, opened 2018-12-04
Refactor MTL to use less memory.

geordie666:
This PR refactors MTL to use less memory and has other MTL-related enhancements. It should address https://github.com/desihub/desitarget/issues/415. Includes:

- Using `NUMOBS_INIT` and `PRIORITY_INIT` in the targets files to set initial priorities and observation numbers instead of repeatedly calling the expensive `calc_priority` and `calc_numobs` functions. This was referenced at the bottom of https://github.com/desihub/desitarget/pull/405 but never resolved.
- Removing table copies from `desitarget.mtl` to use less memory.
- Setting informational (`NORTH/SOUTH`) bits to 0 for `PRIORITY` and `NUMOBS` so that they never influence the main bits.

I tested this with some fake redshift catalogs and found that, generally:

- The new code is 2x slower than the original MTL code, mostly because I can't beat the speed of the outer join in `astropy.table`.
- The new code uses 2-5x less memory than the current MTL code, depending on the size of the redshift catalog. The savings in memory increases for increasingly larger input catalogs.

@forero: I've checked that my new code always returns the same results as the old MTL code, but I'm requesting a review in the hope that you can confirm that equivalence in a real-world scenario. Note that because this new code, itself, changes how `NUMOBS_INIT` and `PRIORITY_INIT` are assigned, to test the new code on a file of targets you'll first have to recreate those new column values, e.g.:

```
import fitsio
from desitarget.targets import initial_priority_numobs                                         
targets = fitsio.read("/project/projectdirs/desi/target/catalogs/dr7.1/PR372/targets-dr7.1-PR372.fits")
for name in ['NUMOBS', 'PRIORITY']:
    targets.dtype.names = [name+'_INIT' if col == name else col for col in targets.dtype.names]
targets["PRIORITY_INIT"], targets["NUMOBS_INIT"] = initial_priority_numobs(targets, survey='main')
```

and, as the outer join in `astropy.table` sorts on the matching key, you'll need to re-sort the output from the new code, e.g.:

```
import numpy as np
from desitarget import mtl
blat = mtl.make_mtl(targets,zcat=zcat)
foo = np.argsort(blat["TARGETID"])
blat = blat[foo]
```

before comparing to the output from the old code.

@sbailey: I'm requesting a review because I'm curious if the output of the MTL code has to be a Table? I can save even more memory if I can just pass out a numpy rec array. Also, perhaps you know a memory-neutral trick to match the zcat to the targets catalog and return the index of the match more quickly than my dictionary-based solution at:
https://github.com/desihub/desitarget/blob/MTLmemADM/py/desitarget/mtl.py#L54-L56.


apcooper:
@geordie666, could you explain why the following change was needed, please?

https://github.com/desihub/desitarget/blob/7cb31d66d2d38a4c41f5220f5190ba6850ed5585/py/desitarget/mtl.py#L79-L82

geordie666:
@apcooper: This isn't really a change. It's just a shift of where in the code this step occurs. 

In master, resetting `NUMOBS_MORE` for every target class in the BGS sample so that it is always 1 is currently done in `calc_numobs`, here:

https://github.com/desihub/desitarget/blob/master/py/desitarget/targets.py#L571-L580

This is the only "dynamic" part of the current MTL code in terms of recalculating `NUMOBS_MORE` outside of the body of the `mtl.make_mtl()` function itself. 

I thought about retaining this "dynamic" part of the code in a new `calc_numobs_no_table` function that would mirror the `BGS` lines in the current version of `calc_numobs`...but as it's equivalent to only a couple of lines of code, I decided to just leave it in `mtl.make_mtl()`.

apcooper:
Thanks. Now I understand why the code is there but I'm still confused why it's needed.

I was about to launch into more questions, but, to my considerable chagrin, git blame tells me I wrote that bit of code and comments that I struggled to understand now. Oh dear.

Now I've had a moment to think about it, I reckon it's harmless enough (BGS targets do drop down to low priority when they get a redshift, even if they are observed forever) but I'm struggling to see why this needs a special hack. Why doesn't the standard framework take care of this requirement? Maybe this same logic should apply to other classes?

Anyway, perhaps this isn't the right place to discuss this, so I'll look into it more for myself and raise an issue if needed.

geordie666:
Thanks for the note @apcooper: I tried to hew close to your original code to obtain the same outputs without as many tables. It could well be that we don't want some of the original behavior, though!

My "special hack" arises from the comment (which I assume is yours) that we need to "force the calculation of nmore in targets.calc_priority (and ztargets['NOBS_MORE'] in mtl.make_mtl) to give nmore = 1" so it was always a hack/force. There could well be a better way of coding this, but it's essentially caused by the fact that for all of the other target classes `NUMOBS_MORE` is decremented by 1 with every observation, but the BGS targets are (apparently) meant to hold steady at 1.

sbailey:
> @sbailey: I'm requesting a review because I'm curious if the output of the MTL code has to be a Table? I can save even more memory if I can just pass out a numpy rec array. 

I don't think is has to be a Table.  If any code using it really does need a table, simply casting it to a Table is fast and doesn't copy memory.  At the same time, it isn't clear to me that passing out a numpy rec array is more memory efficient, since adding new columns to a recarray temporarily makes a copy of the entire input array, while adding a column to a Table does not.

> Also, perhaps you know a memory-neutral trick to match the zcat to the targets catalog and return the index of the match more quickly than my dictionary-based solution at:
https://github.com/desihub/desitarget/blob/MTLmemADM/py/desitarget/mtl.py#L54-L56.

When I have a fresher brain I think I can modify another recipe to do that.  But even in the meantime, I see that once you create `zmatcher`, you use it multiple times like `targets[zmatcher]` which is resorting and copying the array every time.  Better to create `targets_zmatcher = targets[zmatcher]` and use that.

py/desitarget/targets.py
```diff
@@ -336,64 +336,77 @@ def initial_priority_numobs(targets, survey='main'):
...
 
-    #ADM set up the output arrays
-    outpriority = np.zeros(len(targets), dtype='int')-1
-    outnumobs = np.zeros(len(targets), dtype='int')-1
-
-    for colname, mask in zip(colnames,masks):
-        #ADM first determine which bits actually have priorities
+    # ADM much of the code is set up to expect NUMOBS==2 for "just" an LRG
+    # ADM but we need NUMOBS==2 for a 1-PASS LRG. The easiest way to deal
+    # ADM with this is just to unset the LRG bit if the LRG sub-bits are set.
```
sbailey: I think you can avoid this special case handling by setting `numobs['desi_mask']['LRG'] = 0` in targetmask.yaml so that the `LRG_1PASS` and `LRG_2PASS` bits will determine numobs instead of the `LRG` bit.

If that doesn't work, then we have a different logic error, since numobs should be the max of the numobs for any of the individual bits.

geordie666: I initially set the code up exactly as you're suggesting (with `NUMOBS` being set to 0 for the `LRG` bit). I decided I was concerned about the possibility that an LRG slipped through the cracks somehow (perhaps due to future code changes) without `LRG_1PASS` or `LRG_2PASS` being set. On reflection, though, I think I'll do it your way, but add an assert in `cuts.py` that tests for that case.

sbailey:
Regarding the thread with @apcooper about BGS NUMOBS special case: let's not bog down this PR with this, but I'm concerned about that.  If we tile all bright layers at the same time, then it doesn't matter so much.  But if we tile each layer at different times (like the dark program) I think this code would have an unintended consequence of re-observing a high SUBPRIORITY BGS target a second time instead of giving a new fiber to a different BGS target.  Let's followup with BGS about exactly what they need and see if there is a cleaner way to handle this (which might still require changes to desitarget or fiberassign)

geordie666:
@sbailey: I updated this PR to hopefully cover all of your requests, both for this PR and for https://github.com/desihub/desitarget/issues/431 and https://github.com/desihub/desitarget/issues/432. Some highlights:

- Everything now works consistently with the old version of MTL, but by setting `NUMOBS` for the `LRG` bit to 0 and calculating priorities using `LRG_1PASS` and `LRG_2PASS` instead. 

- To achieve this, I had to deprecate the use of `LRG` completely in `calc_priority`. To guard against this catching us out in the future (at least for real targets), I added an assert in `targets.finalize` that insists that `LRG` isn't set without a `PASS` bit also being set.

- I updated the bits in the manner you described in https://github.com/desihub/desitarget/issues/431 and https://github.com/desihub/desitarget/issues/432.

- I was more judicious in my use of `zmatcher` for my zcat-to-targets join. This seems to have both sped the code up and (further) reduced memory use. Thanks!

I'm happy to merge this without a review from @forero if you are. It is quite a major refactor of MTL, though, so I'm also happy to hold off a while.

py/desitarget/targets.py
geordie666: Fixed in the latest push.

sbailey:
Changes look good, but please debug the following which fails on the 18.11 reference run:
```python
from desitarget.mtl import make_mtl

basedir = '/project/projectdirs/desi/datachallenge/reference_runs/18.11'
targets = Table.read(basedir+'/targets/targets.fits')
zcat = Table.read(basedir+'/spectro/redux/mini/zcatalog-mini.fits')

mtl = make_mtl(targets, zcat=zcat)
```

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-1-fa746666f70e> in <module>()
      6 zcat = Table.read(basedir+'/spectro/redux/mini/zcatalog-mini.fits')
      7 
----> 8 mtl = make_mtl(targets, zcat=zcat)

/global/cscratch1/sd/sjbailey/desi/code/desitarget/py/desitarget/mtl.py in make_mtl(targets, zcat, trim)
     54         d = dict(tuple(zip(targets["TARGETID"], np.arange(n))))
     55         # ADM loop through the zcat and look-up the index in the dictionary.
---> 56         zmatcher = np.array([d[tid] for tid in zcat["TARGETID"]])
     57         ztargets = zcat
     58         if ztargets.masked:

/global/cscratch1/sd/sjbailey/desi/code/desitarget/py/desitarget/mtl.py in <listcomp>(.0)
     54         d = dict(tuple(zip(targets["TARGETID"], np.arange(n))))
     55         # ADM loop through the zcat and look-up the index in the dictionary.
---> 56         zmatcher = np.array([d[tid] for tid in zcat["TARGETID"]])
     57         ztargets = zcat
     58         if ztargets.masked:

KeyError: 288230398217951040
```

Likely cause: zcat containing sky fibers with TARGETIDs that aren't in the original target list.  I think it is safe to use `np.in1d` to identify and filter out any zcat entries that aren't in the original target list and then proceed.

Full disclosure: this test also fails on master albeit with a different error, so it might be a problem with the input data.  Maybe.

geordie666:
@sbailey: I needed to make one minor addition to ensure that `make_mtl()` would work with targets that were cast as Tables.

Now, The reason the reference test fails (both for the code in this branch and the old code in master) is that some of the zcat objects are not targets, as you say. I can excise these because they don't have `DESI_TARGET` set, as follows:

```python
from desitarget.mtl import make_mtl

basedir = '/project/projectdirs/desi/datachallenge/reference_runs/18.11'
targets = Table.read(basedir+'/targets/targets.fits')
zcat = Table.read(basedir+'/spectro/redux/mini/zcatalog-mini.fits')

truetarg = zcat["DESI_TARGET"] != 0

mtl = make_mtl(targets, zcat=zcat[truetarg])
```
Question: should I embed this formalism to excise non-targets directly in `make_mtl` or is this an issue with how the input data is made for the reference test?

sbailey:
I believe that `zcat['DESI_TARGET'] == 0` is an upstream error.  Everything should have some bit set, e.g. even broken fibers that weren't placed on anything should have `NO_TARGET` set.  But I suggest that `make_mtl` should gracefully handle unexpected extra zcat entries and move on.  I added that to this branch and also added a card to the [18.12 project](https://github.com/orgs/desihub/projects/9) to identify why those occur the first place.

Trimming out the DESI_TARGET=0 zcat entries, I compared the outputs of this branch to master:
* MWS unobs priority 1400 became 1500.  ok.
* ELG_NORTH/ELG_SOUTH entries that weren't also ELG became priority=0 numobs=0.  Good, fixes bug in previous version.
* Some but not all standards went from priority=0 numobs=0 to priority=-1 numobs=-1.  This might be a bug; please investigate and/or explain.

geordie666:
Regarding the final point in your list...this is caused by the fact that this branch itself changes how `NUMOBS_INIT` and `PRIORITY_INIT` are set. Consider these two snippets of code:

```python
from desitarget.mtl import make_mtl
from desitarget.targets import initial_priority_numobs 
from desitarget.targetmask import desi_mask, bgs_mask, mws_mask
basedir = '/project/projectdirs/desi/datachallenge/reference_runs/18.11'
targets = Table.read(basedir+'/targets/targets.fits')
zcat = Table.read(basedir+'/spectro/redux/mini/zcatalog-mini.fits')
```
```python
mtl = make_mtl(targets, zcat=zcat)

stdb = mtl["DESI_TARGET"] & desi_mask["STD_BRIGHT"] != 0
stdf = mtl["DESI_TARGET"] & desi_mask["STD_FAINT"] != 0
notmws = mtl["DESI_TARGET"] & desi_mask["MWS_ANY"] == 0
std = (stdf | stdb) & notmws
print("PRIORITY:",set(mtl[std]["PRIORITY"]))
print("NUMOBS_MORE:",set(mtl[std]["NUMOBS_MORE"]))
 
PRIORITY: {0, -1}
NUMOBS_MORE: {0, -1}
```

```python
mtl = make_mtl(targets, zcat=zcat)

stdb = mtl["DESI_TARGET"] & desi_mask["STD_BRIGHT"] != 0
stdf = mtl["DESI_TARGET"] & desi_mask["STD_FAINT"] != 0
notmws = mtl["DESI_TARGET"] & desi_mask["MWS_ANY"] == 0
std = (stdf | stdb) & notmws
print("PRIORITY:",set(mtl[std]["PRIORITY"]))
print("NUMOBS_MORE:",set(mtl[std]["NUMOBS_MORE"]))
 
PRIORITY: {0, -1}
NUMOBS_MORE: {0, -1}
```
```python
for name in ['NUMOBS', 'PRIORITY']:
    targets.rename_column(name, name+'_INIT')
targets["PRIORITY_INIT"], targets["NUMOBS_INIT"] = initial_priority_numobs(targets, survey='main')

mtl = make_mtl(targets, zcat=zcat)
 
stdb = mtl["DESI_TARGET"] & desi_mask["STD_BRIGHT"] != 0
stdf = mtl["DESI_TARGET"] & desi_mask["STD_FAINT"] != 0
notmws = mtl["DESI_TARGET"] & desi_mask["MWS_ANY"] == 0
std = (stdf | stdb) & notmws
print("PRIORITY:",set(mtl[std]["PRIORITY"]))
print("NUMOBS_MORE:",set(mtl[std]["NUMOBS_MORE"]))

PRIORITY: {0}
NUMOBS_MORE: {0}
```

i.e., you have to run the reference target through the new version of `initial_priority_numobs()` in this branch to correctly set `PRIORITY_INIT` and `NUMOBS_INIT`

Issue #428, opened 2018-12-03
apply bright cut when selecting all target classes?


moustakas:
Analogous to #296 (for ELGs), are there any arguments against applying a (soft) bright cut when selecting all target classes, especially QSOs and LRGs?

I'm asking in the context of #427, as right now stellar contaminants for QSOs as bright as `r=17` are being selected, which is not what the real target catalogs look like.

@geordie666 @Cyeche

geordie666:
There is already such a cut enforced on the LRGs at about r > 18, e.g.:

https://github.com/desihub/desitarget/blob/master/py/desitarget/sv1/sv1_cuts.py#L109

So, as a bright-end cut already exists for the LRGs and ELGs, the question is really the QSOs.

There are over 300 bright quasars in BOSS with a redshift of greater than 2.5 and r < 18, and I know those are particularly valuable for forest work. There are even a couple of dozen with a redshift of greater than 2.5 and r < 17, and a couple with r < 16!

I think we'd be completely safe with a soft limit of r~15.5. If we want to go to r~17, then that would be a cut for the quasar target selection to explicitly apply to best balance completeness and purity. @Cyeche?

Cyeche:
We are currently applying a cut at 17.5 for the RF selection which is the baseline selection now.
https://github.com/desihub/desitarget/blob/master/py/desitarget/cuts.py#L1277

Cheers, Christophe.

geordie666:
@moustakas: Then I think it might meet your requirements to just add an r > 17.5 cut to the `isQSO_cuts_north` and `isQSO_cuts_south` functions?

moustakas:
Yes, that would be great.  Here's one of the QSO QA plots so you can see that `r>17.5` would toss out the worse offending stars.  And I had missed the bright cut on LRGs, so that works, too.  Thanks.

<img width="350" alt="screen shot 2018-12-03 at 4 47 27 pm" src="https://user-images.githubusercontent.com/1431820/49403747-8c033c00-f71b-11e8-98e4-b40fc9e9d22b.png">

geordie666:
Added an r > 17.5 limit for the quasar color cuts in #433.

PR #433, opened 2018-12-08
Add explicit r > 17.5 limit to QSO color cuts selection.

geordie666:
This PR adds an r > 17.5 "bright-end" limit to the QSO color cuts selection for the FDR (see https://github.com/desihub/desitarget/issues/428).

As long as I was adding this limit, I refactored the QSO color cuts selection slightly to make it a little easier to maintain.

### DIFF
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -986,168 +986,36 @@ def isBGS_colors(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None,
     return bgs
 
 
-def isQSO_colors(gflux, rflux, zflux, w1flux, w2flux, optical=False, south=True):
-    """Convenience function for backwards-compatability prior to north/south split.
+def isQSO_cuts(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None, 
+               w1snr=None, w2snr=None, deltaChi2=None, brightstarinblob=None, 
+               release=None, objtype=None, primary=None, optical=False, south=True):
+    """Definition of QSO target classes from color cuts. Returns a boolean array.
 
-    Args:
-        gflux, rflux, zflux, w1flux, w2flux: array_like
-            The flux in nano-maggies of g, r, z, W1, and W2 bands.
-        optical: boolean, defaults to False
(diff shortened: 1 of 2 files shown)


geordie666:
This is a simple PR, and I'll self-merge tomorrow morning if nobody has comments.

py/desitarget/cuts.py
```diff
@@ -986,89 +986,71 @@ def isBGS_colors(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None,
...
     """
-    if south is False:
-        return isQSO_colors_north(gflux, rflux, zflux, w1flux, w2flux,
-                                  optical=optical)
-    else:
-        return isQSO_colors_south(gflux, rflux, zflux, w1flux, w2flux,
-                                  optical=optical)
 
+    if not south:
+        gflux, rflux, zflux = shift_photo_north(gflux, rflux, zflux)
```
moustakas: I thought we had trained distinct color-cuts in the north vs the south rather than transforming the colors like this. I'm sure I'm just misremembering.

Also, I have a pass-by-reference vs pass-by-value question: since you're using the same variable names here won't the *input* grz fluxes be modified after this script is called and after they've been shifted to a different photometric system?  Or do the changes remain within the function (I honestly can't remember what happens in this case in python).

geordie666: The collaboration as a whole certainly expressed a preference for training color cuts rather than transforming the colors, and we've removed transformations from all of the other cuts except for the quasars. This shift was added by @CAClaveau [here](https://github.com/desihub/desitarget/commit/3345acd9ea799f0f7b6c384b98b6df00ecf52e8a), presumably as a placeholder. Although, we could ask @CAClaveau and @Cyeche to ensure that this shift is reverted to distinct cuts in the future.

The pass-by-reference vs pass-by-value question is a good one. I'm pretty sure that the changes remain within the function...but just to check I made a version of the code that copied `rflux`, `gflux` and `zflux` before calling `isQSO_cuts` and then compared the output `rflux`, `gflux` and `zflux` to the copies. They were identical.

Issue #435, opened 2018-12-11
move input.yaml to an installable place


moustakas:
We should move `doc/mock_example/input.yaml` to an installable location like `py/desitarget/mock/data` so it can be used by `desitest` during integration testing.

moustakas:
Done in #436.

PR #436, opened 2018-12-14
move select-mock-targets.yaml file to an installable location

moustakas:
Simple PR to address #435 (and to also take the opportunity to remove some obsolete files).

In `setup.py` we do
```
setup_keywords['package_data'] = {'desitarget': ['data/*',],
                                  'desitarget.cmx': ['data/*',],
                                  'desitarget.sv1': ['data/*',],
                                  'desitarget.test': ['t/*',],
                                  'desitarget.mock': [os.path.relpath(_,'py/desitarget/mock') for _ in [os.path.join(_[0],'*') for _ in os.walk('py/desitarget/mock/data')]],
                                 }
```
so to my inexperienced eye it looks like `py/desitarget/mock/data/select-mock-targets.yaml` should be installed correctly.


sbailey:
I verified that yes, it does install it (yay for conda environments that we can throw away afterward).  Thanks.  I will update the minitest to use this instead of carrying around it's own copy.

Issue #437, opened 2018-12-14
GFA targets MORPHTYPE


sbailey:
Small, but putting this into an issue so that we don't forget:

For consistency with the target files and to match what ICS needs downstream, the GFA target list should also rename `TYPE` -> `MORPHTYPE`.

geordie666:
Fixed in #438.

PR #438, opened 2018-12-14
Remove dependencies on the Legacy Surveys Gaia astrometry files

geordie666:
This PR removes our reliance on Legacy Surveys chunked "astrom" files when handling Gaia data. Changes include:

- Using a `$GAIA_DIR` environment variable instead of passing a Gaia directory around.
- New functions to retrieve and organize Gaia files in the `$GAIA_DIR`. Including:
    * A function to wget Gaia DR2 CSV files from the ESA archive.
    * A function to convert the Gaia DR2 CSV files to FITS.
    * A function to reorganize the Gaia FITS files into (NESTED) HEALPixels.
- Using the NESTED HEALPix scheme for Gaia files throughout `desitarget` (the Legacy Surveys "astrom" files used the RING scheme).
- Renaming the output column ``TYPE`` to ``MORPHTYPE`` for GFA targets.

### DIFF
--- a/bin/make_imaging_weight_map
+++ b/bin/make_imaging_weight_map
@@ -31,8 +31,7 @@ ap.add_argument("--nside", type=int,
                 help='The resolution (HEALPixel nside number) at which to build the map (defaults to 256)',
                 default="256")
 ap.add_argument("--gaialoc",
-                help='Directory of Gaia "chunks" files (to derive stellar density). A FITS file that already contains the Gaia stellar densities at nside can be passed as a speed-up',
-                default='/project/projectdirs/cosmo/work/gaia/chunks-gaia-dr2-astrom')
+                help='Directory of Gaia "chunks" files (to derive stellar density). A FITS file that already contains the Gaia stellar densities at nside can be passed as a speed-up')
 
 ns = ap.parse_args()
 
--- a/bin/select_gfas
+++ b/bin/select_gfas
@@ -21,9 +21,6 @@ ap.add_argument("surveydir",
                 help="Base directory for a Legacy Surveys Data Release (e.g. '/global/project/projectdirs/cosmo/data/legacysurvey/dr6/' at NERSC)")
 ap.add_argument("dest", 
                 help="Output GFA targets file (e.g. '/project/projectdirs/desi/target/catalogs/gfas-dr6-0.20.0.fits' at NERSC)")
-ap.add_argument('-g', '--gaiadir',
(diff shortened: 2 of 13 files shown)


geordie666:
For the record, example output files are at NERSC in:
```
/project/projectdirs/desi/target/gaia_dr2/csv
/project/projectdirs/desi/target/gaia_dr2/fits
/project/projectdirs/desi/target/gaia_dr2/healpix
```
Which correspond to setting the $GAIA_DIR environment variable to `/project/projectdirs/desi/target/gaia_dr2.`

geordie666:
Also, this should fix #437.

moustakas:
This is fabulous, @geordie666.  May I make a request for a data file that directly pertains to this PR and which would be very helpful for a higher level of realism in `select_mock_targets`?  The (FITS) file I would like is a healpix map of the (Gaia-derived) stellar density over the whole sky, which I could then use to modulate the stellar contamination rate with Galactic coordinates.  Having a simple function (like what you wrote for the pixel weights) to downsample / upsample from the nominal `nside` (e.g., 128 or 256) to another `nside` would also be very useful. 

If this makes sense and is something you’d be willing to write and include here that’d be awesome.  If not, no worries—it’s definitely something I can tackle separately.

sbailey:
@moustakas please put your request into a new issue.  Since we're about to loose @geordie666 for the holidays, I'm eager to get this merged and tagged so that we can get a new TS run to go along with 18.12.

geordie666:
@moustakas: I think what you requested already exists in the `pixweight-` files. The ``STARDENS`` columns in those files is agnostic to the Data Release (it's determined for the whole sky). For instance:

```
a = fitsio.read("/project/projectdirs/desi/target/catalogs/dr7.1/0.22.0/pixweight-dr7.1-0.22.0.fits")
b = fitsio.read("/project/projectdirs/desi/target/catalogs/dr6/0.22.0/pixweight-dr6.0-0.22.0.fits")
set(a["STARDENS"]-b["STARDENS"])
{0.0}
```
That's derived from the Gaia density in the `randoms.py` code with some restrictions to derive only "stellar" sources:

https://github.com/desihub/desitarget/blob/master/py/desitarget/randoms.py#L531-L577

As a bonus, I think the code to upsample/downsample those pixweight files is already in `desitarget.io.load_pixweight_recarray`:

https://github.com/desihub/desitarget/blob/master/py/desitarget/io.py#L907

Let me know if you need something else, though, and open a post-holidays issue for me.

PR #444, opened 2018-12-17
use zmin_lya when estimating target density of mock

moustakas:
Minor fix to use `zmin_lya` (usually read from the configuration file, which defaults to `zmin_lya=1.8`) when estimating the average target density of the Lya mock.  The impact of the bug is minimal because the minimum redshift of the current default Lya mock is `1.792` (including RSD).

Merging this can happen after `18.12`--there's no rush.


moustakas:
With some very helpful input from the Lya working group, I'd like to use this PR to fix a related normalization issue which affects the QSOs (Lya+tracers) in `select_mock_targets`.   I'll implement this in this PR the next couple days.

PR #449, opened 2019-01-15
Update QSO selection for SV, minor improvements to random catalog generation

geordie666:
This PR:

- updates the QSO random forest for SV to facilitate different selections in the north and south.
- adds seeds to the random catalog generation for better reproducibility.
- defaults to using the `DUST_DIR` environment variable when generating randoms if a map directory is not passed.


geordie666:
@CAClaveau: After our conversation over email, I've made your requested correcting to the SV QSO selection code. Can you please check to see if this now looks correct?

Issue #450, opened 2019-01-23
calc_priority vs. calc_priority_no_table


sbailey:
`desitarget.targets.calc_priority` and `desitarget.targets.calc_priority_no_table` have lots of replicated code and are just begging for us to update one and forget about making the same change to the other.

Is `calc_priority` still needed?  A quick grep only shows calls to `calc_priority_no_table`.  If both are genuinely needed, let's refactor into
```python
def calc_priority(..., copy=False):
    ...
    if copy:
        targets = Table(targets).copy()  #- or perhaps without the Table part...
    ...
```

geordie666:
I left `calc_priority` for backwards-compatibility in case any code outside of `desitarget` was calling it. I'm happy to break that, though and just rename `calc_priority_no_table` to be `calc_priority`, though. I personally know of no other code that is calling `calc_priority` than MTL itself.

sbailey:
I grepped all our repos and `desitarget.mtl.make_mtl` and tests are the only places that call `calc_priority`.  Let's drop current `calc_priority` and rename `calc_priority_no_table` -> `calc_priority` and proceed.

If `calc_priority(_no_table)` alters the input `targets` table (which I presume was the reason for the original copy), please make that very clear in the `calc_priority` and `make_mtl` docstrings.

sbailey:
done in PR #452.

PR #452, opened 2019-01-28
Deprecate table-based version of calc_priority

geordie666:
This PR fully deprecates the version of `calc_priority` that included a memory-intensive table copy and replaces it with the new `calc_priority_no_table` (which has been renamed to `calc_priority`). Other than renaming the `calc_priority_no_table` function, the main changes are in `test_priorities` to reflect the fact that `calc_priority` now doesn't update `NUMOBS_MORE` (calculating priorities and numbers of observations were already being handled separately in `mtl`).

This change is not backwards-compatible. Any code that is calling `calc_priority` will break because `calc_priority` now takes two smaller tables as input (the target information and the redshift/status information) rather than allowing a monolithic table to be passed.

Should solve https://github.com/desihub/desitarget/issues/450.


sbailey:
I have grepped all of our core repositories and can't find any that call `calc_priority` outside of desitarget itself.  Let's merge.  Thanks for the cleanup.

PR #460, opened 2019-02-27
fix desitarget.mock.build.targets_truth

sbailey:
PR #456 updated `desitarget.targets.initial_priority_numobs()` to no longer need a `survey` parameter, but this broke `desitarget.mock.build.targets_truth()` which was still passing in a `survey` parameter.  This PR fixes that.

Non-mock desitarget updates often break `desitarget.mock.build.targets_truth()`, so I've finally added a unittest for that.  Since this is an expensive test and requires big mock file inputs, it only runs if `$DESITARGET_RUN_MOCK_UNITTEST` is set.  The intension is that we would *not* run this on Travis, but I would include it in the nightly NERSC tests so that we would at least catch any problems within 24 hours.  I don't think it is worth maintaining stripped down versions of the input mocks for testing, since the input mocks themselves frequently change and we really want to test that those files still work too.

@geordie666 or @moustakas please take a look (though if you can't look with a day or so, I may go ahead and self-merge since this is a blocking factor for end-to-end integration tests)

### DIFF
--- a/py/desitarget/mock/build.py
+++ b/py/desitarget/mock/build.py
@@ -935,7 +935,7 @@ def targets_truth(params, healpixels=None, nside=None, output_dir='.',
                             seed=healseed)
         
 def finish_catalog(targets, truth, objtruth, skytargets, skytruth, healpix,
-                   nside, log, seed=None, survey='main'):
+                   nside, log, seed=None):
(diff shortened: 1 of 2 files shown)


geordie666:
This looks fine to me (and the Travis failures look like glitches). So, feel free to merge.

PR #461, opened 2019-03-02
Select GFAs for commissioning across wider sky areas.

geordie666:
This PR updates the GFAs code to select targets from Gaia across large sky areas (rather than just the DESI tiling footprint). 

The main functionality is through the executable `select_cmx_gfas`, which allows cuts on Gaia G-band, minimum Dec and Galactic latitude. Information from DESI imaging is still returned in areas that overlap the Legacy Surveys.


sbailey:
Looks good; thanks.  Merging now.

PR #462, opened 2019-03-06
Resolve duplicate targets across 2 imaging releases in preparation for DR8

geordie666:
This PR adds a new `desitarget.targets.resolve` function to resolve duplicate N/S targets, which will be useful for DR8 imaging (and beyond). It includes:

- Resolve N/S duplicates by combining `RELEASE` (was the photometric system N/S?) with defined areal cuts (is the footprint officially N/S?).
- Add the `RELEASE` column to the GFA data model as its needed to resolve duplicates.
- Apply the new resolve code to GFAs.
- Deprecate Gaia-matching code for GFAs, as we no longer need it.

Although the resolve code is only applied to GFAs in this PR, it should only require one line of code (one function call) to apply it to other target classes in future updates.

### DIFF
--- a/bin/select_cmx_gfas
+++ b/bin/select_cmx_gfas
@@ -25,8 +25,6 @@ ap.add_argument("dest",
 ap.add_argument('-m', '--maglim', type=float,
                 help='Magnitude limit on GFA targets in Gaia G-band (defaults to [18])',
                 default=18)
-ap.add_argument("--gaiamatch", action='store_true',
-                help="DO match to Gaia DR2 chunks files in order to populate Gaia columns for the GFA locations")
 ap.add_argument('-n', "--numproc", type=int,
                 help='number of concurrent processes to use (defaults to [{}])'.format(nproc),
                 default=nproc)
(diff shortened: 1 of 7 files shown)


djschlegel:
The two possible dec split choices in the NGC are 32.125 deg or 32.375 deg (chosen to be at brick boundaries).  I think we could use the higher declination cut, with the risk that there might be some small places with not complete fill near RA=250 and RA=262 deg.

In the SGC, we should only be targeting from DECam, extending to even higher declinations wherever the imaging data happens to go a bit further north.

geordie666:
The code is, indeed, set up to only target from DECam in the SGC, and explicitly makes a cut on Galactic latitude as part of the resolve.

Currently, I set the NGC split at 32o. But, it's a one-line change [in the new desitarget_resolve_dec() function](https://github.com/desihub/desitarget/blob/ADM-resolve/py/desitarget/io.py#L90-L93) when we decide whether 32.125o or 32.375o is the correct choice.

geordie666:
I'll quickly update to a default of Dec = 32.375o for now, though.

PR #466, opened 2019-03-15
Initial updates for DR8

geordie666:
This PR handles changes for moving to DR8 of the imaging surveys. It's possible that additional tweaks will be needed for the final DR8 imaging release, but this should be the vast majority of the updates and I don't want this branch to go stale waiting for the final version of DR8. Includes:

- Update the Data Model (e.g. boolean `BRIGHTSTARBLOB` -> bitmask `BRIGHTBLOB`)
- Resolve targets and randoms for duplicate bricks in regions where the imaging surveys overlap.
- Handle the "dual" file structure where BASS/MzLS and DECaLS exist in the same directory.

Still to do:

- Resolve skies so that we don't have additional sky locations in the overlap regions.

### DIFF
--- a/bin/gather_targets
+++ b/bin/gather_targets
@@ -21,7 +21,7 @@ ap.add_argument("infiles",
                 help="SEMI-COLON separated list of input files, which may have to be enclosed by quotes (e.g. 'file1;file2;file3;file4')")
 ap.add_argument("outfile", 
                 help="Output file name")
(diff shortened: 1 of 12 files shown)


sbailey:
Sigh, this is getting to be a bunch of special cases, but I do appreciate your efforts to maintain backwards compatibility while supporting DR8, which is important.  Merging now.

Issue #465, opened 2019-03-13
Standardize Gaia point source selection


sbailey:
For GFA target selection, it would be useful to standardize the cuts for how to select likely point-source targets from Gaia when Legacy Survey / Tractor results are not available (chip gaps, off the edge of the footprint...).

From Dustin via Adam:
```
gaia.pointsource = np.logical_or(
    (gaia.G <= 19.) * (gaia.astrometric_excess_noise < 10.**0.5),
    (gaia.G >= 19.) * (gaia.astrometric_excess_noise < 10.**(0.5 + 0.2*(gaia.G - 19.))))
```

Currently Gaia-only sources have MORPHTYPE='U' = undefined.  Consider making those that pass this cut have MORPHTYPE='GPSF' (other MORPHTYPES are 4-characters, so I'd rather not use a longer name).

geordie666:
Addressed in #467.

PR #467, opened 2019-03-15
Write Gaia morphologies and allow custom tilings for GFAs

geordie666:
This PR updates and streamlines the GFAs code. It includes: 

- Use a cut on `GAIA_ASTROMETRIC_EXCESS_NOISE` to define whether Gaia-only sources are point-sources (called `b'GPSF'`) or are not (called `b'GGAL'`). 
- Combine the `select_` executables for the commissioning and the main survey code into a single script (`select_gfas`) which can take a `--cmx` flag to select GFAs outside of the DESI footprint.
- Allow tile files to be passed directly to the GFA code to facilitate custom tilings (using the `--cmx` option).

Should address https://github.com/desihub/desitarget/issues/465.


py/desitarget/gfa.py
```diff
@@ -218,6 +218,36 @@ def add_gfa_info_to_fa_tiles(gfa_file_path="./", fa_file_path=None, output_path=
...
+    gaia: :class:`~numpy.ndarray`
+        Numpy structured array containing at least the columns `TYPE`,
+        `GAIA_PHOT_G_MEAN_MAG` and `GAIA_ASTROMETRIC_EXCESS_NOISE`.
+
+    Returns
+    -------
+    :class:`~numpy.ndarray`
+        Input array, with the ``TYPE`` column updated to "GPSF" or "GGAL"
+        based on a morphological cut with Gaia.
+    """
```
sbailey: This appears to modify the original input array, which is not clear from the docstring.

It would be useful to have a way to update the TYPE column only if it was not already set to something useful from the legacy surveys.  Consider having this function just return the array of GPSF/GGAL gaia types and leave it to the user to chose which rows of the original table to update.

py/desitarget/gfa.py
```diff
@@ -413,6 +448,10 @@ def select_gfas(infiles, maglim=18, numproc=4, cmx=False):
         Magnitude limit for GFAs in Gaia G-band.
     numproc : :class:`int`, optional, defaults to 4
         The number of parallel processes to use.
+    tilesfile : :class:`str`, optional, defaults to ``None``
+        Name of tiles file to load. If ``None`` or file doesn't exist,
+        then load the entire footprint. If no path to the file is
+        sent, then look in $DESIMODEL/data/footprint.
```
sbailey: This description isn't right.  tilesfiles is passed to `desimodel.io.load_tiles`, which has the following logic:
* if None: load the full footprint
* if no path:
  * if a file of that name is both local and in $DESIMODEL/data/footprint, print a warning and use the desimodel version
  * if only in $DESIMODEL/data/footprint or only local, use that
  * if in neither, raise an error
* with path: load it if it exists, or raise an exception if it doesn't.

For the purposes of this docstring, I think it's ok to just say this is passed to `desimodel.io.load_tiles()` and let that code document the subtleties.

py/desitarget/gfa.py
geordie666: I update the `gaia_morph` function to simply return an array of `GGAL` or `GPSF`, which should satisfy this suggestion. I then used that output elsewhere in the code to set the morphological type for Gaia-only sources.

py/desitarget/gfa.py
geordie666: Noted. Thanks for the clarification. I changed the docstring to simply point to the description in `desimodel.io.load_tiles()`.

geordie666:
@sbailey: I resolved the conflicts with master and incorporated your suggestions. I think this should be ready to merge.

sbailey:
Looks good, though one clarification question: if an object is in the sweeps (i.e. detected by tractor), does this retain the tractor morphology or replace it with the Gaia morphology?  I think the Gaia morphology (GPSF, GGAL) is only used for Gaia-only sources (e.g. off the edge of the footprint) but I wanted to confirm that before merging.

geordie666:
Clarification: Yes, the Gaia morphology is only ever used for Gaia-only objects, which are processed separately to objects in the sweeps. So, the Gaia morphology can never supplant the sweeps.

Specifically, [here](https://github.com/desihub/desitarget/blob/1aa5115548f28c6fd26032039c147a884facc2da/py/desitarget/gfa.py#L532-L535) the Gaia objects are limited to only bricks that are not present in the sweeps.

moustakas:
> Specifically, here the Gaia objects are limited to only bricks that are not present in the sweeps.

What about bricks that are just *barely* covered by the imaging, i.e., along the edges?

geordie666:
There was talk, at some point, about the imaging surveys seeding all Gaia sources, regardless of whether they were in DECaLS/BASS/MzLS. We would still have WISE in those cases, for instance. I don't think that's happening in legacypipe, or is likely to happen, though. It would, however, solve a lot of these edge cases.

In the edge case you describe, GFAs would be available one brick away from the targets (less than 0.25 degrees away from the targets). I'd hope that's sufficient. If not, I can perform a source-by-source match of all-Gaia-sources-in-the-sweeps to all-Gaia-source-in-the-sky, which will be expensive but plausible.

geordie666:
Thinking more, I think I can perform such a match on Gaia `REF_ID` instead of on `BRICKID`, meaning that we could hopefully recover the unique set of Gaia objects rather than the unique set of bricks. If @sbailey thinks that would be the preferred behavior, then I'll open an issue.

sbailey:
Good catch, @moustakas .  I spun this discussion off to new issue #473; please continue the discussion there.

PR #475, opened 2019-03-21
work around deprecated unsafe yaml.load

sbailey:
Starting with pyyaml 5.1, `yaml.load` without specifying a specific Loader is deprecated, due to security issues with the default loader potentially running arbitrary code when given untrusted input.  That doesn't apply to how we are using yaml to load our own files, but it generates scary warning messages and at some future version it would break.  This PR works around that by replacing `yaml.load` with `yaml.safe_load`.

See https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation for more background on this security issue.

### DIFF
--- a/py/desitarget/targetmask.py
+++ b/py/desitarget/targetmask.py
@@ -19,7 +19,7 @@ def load_mask_bits(prefix=""):
     fn = "{}/data/{}targetmask.yaml".format(prefix, prename)
     _filepath = resource_filename('desitarget', fn)
     with open(_filepath) as fx:
-        bitdefs = yaml.load(fx)
+        bitdefs = yaml.safe_load(fx)
         try:
             bitdefs = _load_mask_priorities(bitdefs, handle="priorities", prename=prename)
         except TypeError:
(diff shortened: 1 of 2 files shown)


geordie666:
Thanks. Looks fine to me (and I checked that you caught all instances of `yaml.load`. One of us can merge this when the tests pass.

weaverba137:
Are we certain that the pyyaml version installed *everywhere we use DESI software* supports `yaml.safe_load()`?

If not, the solution is:
```python
try:
    data = yaml.safe_load(file)
except AttributeError:
    data = yaml.load(file)
```

sbailey:
@weaverba137 good question.  I checked that `yaml.safe_load()` goes back to at least version 3.01 released in 2006, so I think we're safe.  i.e. `safe_load` wasn't added recently just to address this security concern.

Also note that this change in desitarget only impacts loading yaml files defining bitmasks in desitarget itself.  We should separately check our yaml usage elsewhere, but I suspect that if `safe_load` doesn't work for one of our yaml files, then we're using a feature of yaml that we hadn't really intended on using (allowing a yaml file to spawn arbitrary code...).

weaverba137:
OK, thanx for checking.  I thought it was more recent than that.

Issue #470, opened 2019-03-19
Add sky and bad_sky to all programs


tskisner:
Going forward, we intend to use a single target column (i.e. DESI_TARGET, SV1_DESI_TARGET, CMX_TARGET) for all input files to a single run of fiberassignment.  This means that sky and bad_sky bits should be defined for every program.  I think this is just a matter of updating the yaml files.

geordie666:
Fixed in https://github.com/desihub/desitarget/pull/477.

Issue #295, opened 2018-03-24
handling targets in north / south overlap regions


sbailey:
Develop a clear documented plan + code for how to handle targets that appear in both southern and northern photometry (i.e. same object with different RELEASE values).  Requirements:

* The input to fiberassign should have one and only one entry per object, i.e. something upstream of fiberassign chooses whether to pass forward the northern or southern version of each object.
* the survey footprint should be split into two areas (north and south) not 3 (north-only, south-only, and overlap).
* files are available to conveniently study the overlap, e.g. to cross compare the measured photometry and target selection efficiencies.

geordie666:
Fully addressed using `--resolve` and `--noresolve` options for all relevant target classes in #497.

PR #476, opened 2019-03-21
fix brightmask for numpy 1.16.2

sbailey:
This PR fixes some numpy view/reshape logic that works with numpy 1.13.3 but fails with numpy 1.16.2 .  I don't understand why it doesn't work, nor did I particularly understand the chain of view and reshapes to get from structured arrays with column names into vanilla 2D arrays.  Regardless, I think I know what you were trying to do for the actual masking and I think the new code reflects that (and tests pass).

This solution replaces fancy view/reshape logic with more straightforward (IMHO) loops over filter names.  Since this only involves loops over 2-3 elements, I don't think it is a significant performance hit (e.g. logic over targets is still vectorized).

But please double check my logic, @geordie666 .

With numpy 1.16.2, test_brightmask fails with:
```
======================================================================
ERROR: test_non_mask_targets (desitarget.test.test_brightmask.TestBRIGHTMASK)
Test that targets that are NOT in masks are flagged as not being in masks
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/sbailey/desitest/code/desitarget/py/desitarget/internal/sharedmem.py", line 377, in get
    return Q.get(timeout=1)
  File "/Users/sbailey/anaconda3/envs/desitest/lib/python3.6/multiprocessing/queues.py", line 105, in get
    raise Empty
queue.Empty

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/sbailey/desitest/code/desitarget/py/desitarget/internal/sharedmem.py", line 674, in map
    capsule = pg.get(R)
  File "/Users/sbailey/desitest/code/desitarget/py/desitarget/internal/sharedmem.py", line 380, in get
    raise StopProcessGroup
desitarget.internal.sharedmem.StopProcessGroup: StopProcessGroup

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/sbailey/desitest/code/desitarget/py/desitarget/test/test_brightmask.py", line 126, in test_non_mask_targets
    rootdirname=self.bsdatadir, outfilename=self.testmaskfile)
  File "/Users/sbailey/desitest/code/desitarget/py/desitarget/brightmask.py", line 502, in make_bright_source_mask
    objs = collect_bright_sources(bands, maglim, numproc, rootdirname, outfilename)
  File "/Users/sbailey/desitest/code/desitarget/py/desitarget/brightmask.py", line 242, in collect_bright_sources
[...]

PR #477, opened 2019-03-21
Update SV selections for DR8

geordie666:
This PR Includes updates to the SV selections and some additional changes for working with the DR8 (two-surveys-in-one-directory) directory structure. It includes:
- New SV targeting bits for QSOs and LRGs.
- New SV selection algorithms for QSOs, ELGs and LRGs.
- Updates to MTL to handle SV LRGs (which are now not 1PASS/2PASS LRGs in SV).
- Additions to how QA reads targets to interpret HEALPixel-split targeting files.
- Updated test files for the quasi-DR8 imaging data model.
- `SKY` and `BAD_SKY` bits have been added to commissioning yaml file.
- Randoms are now properly handled in overlap regions, and understand the DR8 dual directory structure.
- Overlap regions can be written instead of resolved sources for targets and randoms using a `--noresolve` flag.

This solves #470 and will probably also solve #295 (after some vetting to check the "noresolve" files are what is needed).


geordie666:
@sbailey: This is a *lot* of new functionality as I've been fielding many, many requests for updates prior to next week's workshop. I understand that we probably want to vet much of this functionality. But, it would also be nice to merge this and tag it (once I've also quickly vetted #476) so I can spend tomorrow making useful (tagged) targeting files for the workshop.

sbailey:
If you haven't already, please do a sanity check that this still works on dr7.1 and dr6.  Otherwise green light to merge once tests pass.

geordie666:
I've checked dr7.1, but I'll also check dr6. I'll also set-up a new tag so you won't have to.

PR #478, opened 2019-03-23
import speed of light from scipy

londumas:
Move to a common definition of speed of light. Step toward fixing issue https://github.com/desihub/desisim/issues/477.


geordie666:
@londumas: The code looks good, and I ran the QA through to completion. So, no need for a review from @sbailey.

Issue #469, opened 2019-03-19
Key error `BRIGHTBLOB` in select_mock_targets


forero:
Importing `desitarget` version `0.28.0.dev2909`  and running with the following command:

```
select_mock_targets --no-spectra --nproc 4 --nside 16 --seed 10 \
    -c $SCRATCH/quicksurvey_example/targets/no_spectra/dark/input.yaml \
    --output_dir $SCRATCH/quicksurvey_example/targets/no_spectra/dark \
    --tiles $SCRATCH/quicksurvey_example/survey/subset_tiles_dark.fits
```

with the following `input.yaml`

```
# mock target configuration file
targets:
    ELG: {
        target_type: ELG,
        mockfile: '{DESI_ROOT}/mocks/DarkSky/v1.0.1/elg_0_inpt.fits',
        format: gaussianfield,
        density: 2400,
    }
    LRG: {
        target_type: LRG,
        mockfile: '{DESI_ROOT}/mocks/DarkSky/v1.0.1/lrg_0_inpt.fits',
        format: gaussianfield,
        density: 480,
    }
    QSO: {
        target_type: QSO,
        mockfile: '{DESI_ROOT}/mocks/DarkSky/v1.0.1/qso_0_inpt.fits',
        format: gaussianfield,
        zmax_qso: 1.8,
        use_simqso: True,
        density: 120,
    }
    LYA: {
        target_type: QSO,
        mockfile: '{DESI_ROOT}/mocks/lya_forest/london/v4.0/master.fits',
        format: CoLoRe,
        nside_lya: 16,
        zmin_lya: 1.8,
        density: 50,
        use_simqso: True,
        balprob: 0.0,
        add_dla: True,
    }
    MWS_MAIN: {
        target_type: STAR,
        mockfile: '{DESI_ROOT}/mocks/mws/galaxia/alpha/v0.0.6/healpix',
        nside_galaxia: 8,
        format: galaxia,
    }
    MWS_NEARBY: {
        target_type: STAR,
        mockfile: '{DESI_ROOT}/mocks/mws/100pc/v0.0.4/mock_100pc.fits',
        format: mws_100pc,
    }
    WD: {
        target_type: WD,
        mockfile: '{DESI_ROOT}/mocks/mws/wd/v1.0.0/mock_wd.fits',
        format: mws_wd,
    }
    SKY: {
        target_type: SKY,
[...]

PR #480, opened 2019-03-26
first update of select_mock_targets to DR8 data model

moustakas:
Should address #469, but the DR8 data model is a moving target this week, so more changes may be needed.

Also fixes a caching bug identified by @julienguy.


geordie666:
Looks good, thanks!

PR #487, opened 2019-03-28
Add notebook

dkirkby:
This notebook studies the performance of the QSO random forest selection on DR8b, and also the interpretability of the random forest model.

Highlights:
 - The QSO selection is working well on DR8b.
 - A surprisingly large number of DR14 QSOs are not detected in DR8b.
 - The random forest gets most of its power from z-W2 and z-W1.
 - A simpler random forest model using only 4 colors instead of 10 colors + r performs equally well.


dkirkby:
To view the notebook before merging, use [this link](https://nbviewer.jupyter.org/github/desihub/desitarget/blob/qso-dr8b/doc/nb/qso-dr8.ipynb).

sbailey:
Nice.  Have you advertised this to the target selection working group?  Going to fewer features is desirably both for interpretability and speeding up the QSO random forest evaluation, which is our slowest target selection class (and biggest files in git).

Issue #473, opened 2019-03-20
Generate Gaia GFA targets for bricks that are only partially covered by imaging surveys


sbailey:
Spinning a post-merge discussion from #467 into a new ticket:

Currently `select_gfas` only generates Gaia targets for bricks that are not covered by the imaging surveys at all.  @moustakas asked what happens to bricks that are only partially covered, and indeed there is a gap in the targets generated:

![image](https://user-images.githubusercontent.com/218471/54652115-dca46800-4a72-11e9-875d-5a1d15830df7.png)

Thus if a GFA of an edge tile lands in this gap, it won't have as many targets (worst case is that it confuses acquisition and guiding because the fiberassign GFA_TARGETS will be missing potentially bright objects visible on the GFAs).

Find a way to fill these in with Gaia targets without requiring cross matching Gaia targets on fully covered bricks.

geordie666:
I think I can solve this with a relatively quick match on the imaging surveys `REF_ID` (the Gaia `SOURCE_ID`), although it's possible this will introduce Gaia-Tycho duplicates for sources where the Tycho ID takes precedence over the Gaia `SOURCE_ID`.

moustakas:
We handle this case in the pipeline [here](https://github.com/legacysurvey/legacypipe/blob/master/py/legacypipe/reference.py#L43-L61):  Tycho-2 stars that are also in Gaia are dropped (accounting for proper motion and using a 1 arcsec search radius).

Also note that stars from the two catalogs have different `ref_cat` entries (`T2` for Tycho and `G2` for Gaia/DR2).

geordie666:
I also thought of the `ref_cat` solution and will use it. Although, it's another column that we don't propagate into the sweeps (and probably should for DR8).

moustakas:
Let's be sure to review the columns that go into the sweeps during the Utah workshop -- I think we're propagating more columns than we should / need (and missing ones like `ref_cat`).

geordie666:
Excellent idea. We should definitely do that. I've made a note to myself to bring it up.

geordie666:
Addressed in #488.

PR #488, opened 2019-03-30
Remove GFA/Gaia duplicates using Gaia REF_ID

geordie666:
This PR updates the `select_gfas` framework to remove duplicates between Gaia and sweeps sources on `REF_ID` not `BRICKID`.

This addresses https://github.com/desihub/desitarget/issues/473.

An example file is at `/global/cscratch1/sd/adamyers/cmx-gfas-PR488.fits`.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,6 +5,7 @@ desitarget Change Log
 0.29.2 (unreleased)
 -------------------
 
+* Remove GFA/Gaia duplicates on ``REF_ID`` not ``BRICKID`` [`PR #488`_].
 * Various bug and feature fixes [`PR #484`_]. Includes:
     * Fix crash when using sv_select_targets with `--tcnames`.
     * Only import matplotlib where explicitly needed.
@@ -12,6 +13,7 @@ desitarget Change Log
 
 .. _`PR #480`: https://github.com/desihub/desitarget/pull/480
 .. _`PR #484`: https://github.com/desihub/desitarget/pull/484
+.. _`PR #488`: https://github.com/desihub/desitarget/pull/488
 
 0.29.1 (2019-03-26)
 -------------------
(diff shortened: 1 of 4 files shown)


geordie666:
Updated plot to compare to #473:
edge

moustakas:
Very nice!  Not merging because I can't check out the output file at this moment, but the diagnostic plot looks great.

sbailey:
Looks nice, though I am confused about the red points that leak into the blue regions, while I don't see any blue points leaking into the red regions.  Do you understand those?

geordie666:
Red points leaking into blue regions: Empty locations in DR7.1 that can be populated with a Gaia source that is missing in DR7.1.

No blue points leaking into red regions: Impossible to have DR7.1 sources outside of the DR7.1 footprint.

I did check whether we have any duplicates, as defined by close pairs of coordinates. The biggest effect is Gaia objects that are genuinely pairs within ~0.5" of each other (they have different Gaia `REF_ID`). There are several hundred of those across the ~25 million GFA locations.

geordie666:
One thought apropos the red points leaking into the blue regions: I cut zero (undefined) proper motion sources from the sweeps *before* adding the extra Gaia sources. So, the few red points could be Gaia sources with undefined proper motions that I've added back in. Let me check that (tomorrow) before anyone merges this PR.

geordie666:
I shifted the proper motion check to the end of the GFA selection code, and it did, indeed, clean up many of the red points leaking into blue regions:

newedges

The remaining "leaks" are near the very edge of the cross-over from sweeps-selected to Gaia-selected sources, where there are a few missing bricks. I think this is now ready to merge (once tests pass).

sbailey:
Thanks for tracking down and explaining the red/blue mingling.  Looks good; merging now.

Issue #485, opened 2019-03-27
Add PSF size to the QA page


rongpu:
It would be good to have PSF sizes (PSFSIZE_G, PSFSIZE_G and  PSFSIZE_Z) in the density vs systematics plots in the QA pages (e.g. https://portal.nersc.gov/project/desi/users/adamyers/0.22.0/desitargetQA-dr6.0-0.22.0/QSO.html). This would be informative on the seeing dependence of the targeting.

geordie666:
Addressed in #490.

PR #490, opened 2019-04-08
Further updates and enhancements for DR8

geordie666:
Includes:
- Resolve sky locations and SV targets by North/South region.
- Update sky and SV slurm scripts to facilitate DR8-style input (i.e two input directories).
- Write both of two input directories to any output file headers.
- Parallelize plot production to speed-up QA by factors of ~8.
- Add `PSFSIZE` to the random catalogs, pixweight maps and QA plots.
- QA and pixweight maps now have full functionality for SV-style files and bits.
- The code that produces pixel weight maps can now take HEALpixel-split targets as input.
- Add aperture-photometered background flux to randoms catalogs.
- An additional unit test module (function `test.test_geomask`).
- Deprecate `bin/make_hpx_density_file` and instead use `bin/make_imaging_weight_map`.
- `io.read_targets_in_a_box` can now read headers as well as data.
- Update the unit test data to add new DR8 columns and functionality.

Should address https://github.com/desihub/desitarget/issues/485 and https://github.com/legacysurvey/legacypipe/issues/304.


geordie666:
The last commit, here, merits some explanation. In the last few week, `fitsio` has been updated to [v1.0.1](https://github.com/esheldon/fitsio/releases). This release is [not](https://github.com/esheldon/fitsio/blob/master/CHANGES.md#version-101) backwards-compatible.

However, we're still running `v0.9.11` in our default environment at NERSC. Until we fully take the plunge and update to `v1.0+` at NERSC, I'll try to exist in a state where unit tests pass both `v0.9` and `v1.0` of `fitsio` to smooth the transition.

If this becomes burdensome, a simple alternative solution is to change to:
```
- PIP_ALL_DEPENDENCIES="coveralls fitsio==0.9.11"
```
in `.travis.yml`

This PR has been idling for a few weeks, so I'll merge on Monday unless there are objections.

Issue #491, opened 2019-04-29
Make masking cuts on MASKBITS not BRIGHTBLOB


geordie666:
For DR8, the `MASKBITS` column in the Tractor/sweeps files should include a one-to-one mapping to the mask bits in the coadd stacks used to define the randoms. So, we should now make cuts on the information in `MASKBITS` rather than using `BRIGHTBLOB` to ensure exact correspondence between the targets and random catalogs. 

This is to guard against the fact that (quoting @dstndstn via @moustakas):
```
`BRIGHTBLOB` gets set based on *initial* position (source detection integer pixel peak), while `MASKBITS` gets set after fitting has finished (i.e., on the final brick pixel position).
```
As a reminder, here are the `MASKBITS` values:
```
BITNM0  = 'NPRIMARY'           / maskbits bit 0: not-brick-primary
BITNM1  = 'BRIGHT  '           / maskbits bit 1: bright star in blob
BITNM2  = 'SATUR_G '           / maskbits bit 2: g saturated + margin
BITNM3  = 'SATUR_R '           / maskbits bit 3: r saturated + margin
BITNM4  = 'SATUR_Z '           / maskbits bit 4: z saturated + margin
BITNM5  = 'ALLMASK_G'          / maskbits bit 5: any ALLMASK_G bit set
BITNM6  = 'ALLMASK_R'          / maskbits bit 6: any ALLMASK_R bit set
BITNM7  = 'ALLMASK_Z'          / maskbits bit 7: any ALLMASK_Z bit set
BITNM8  = 'WISEM1  '           / maskbits bit 8: WISE W1 bright star mask
BITNM9  = 'WISEM2  '           / maskbits bit 9: WISE W2 bright star mask
BITNM10 = 'BAILOUT '           / maskbits bit 10: Bailed out of processing
BITNM11 = 'MEDIUM  '           / maskbits bit 11: Medium-bright star
BITNM12 = 'GALAXY  '           / maskbits bit 12: LSLGA large galaxy
BITNM13 = 'CLUSTER '           / maskbits bit 13: Cluster
```

geordie666:
Addressed in #494.

PR #494, opened 2019-05-08
Additional data model updates for DR8

geordie666:
This PR updates a few remaining elements of the data model to run on the DR8 imaging surveys. Includes:

- Add the `WISEMASK_W1` and `WISEMASK_W2` bitmasks to the random catalogs.
- Deprecate `BRIGHTBLOB` in favor of using `MASKBITS` for targets.

The second update is necessary because `MASKBITS` should be identical for the targets and for the randoms, whereas `BRIGHTBLOB` would be marginally inconsistent between targets and randoms. See #491, which this PR should address.


geordie666:
@moustakas: as I know you're about to start working on the mocks again, I wanted to push the (hopefully) final data model changes for DR8.

These will, in typical fashion, break `select_mock_targets`, although I think you probably just have to change occurrences of `BRIGHTBLOB` to `MASKBITS`.

If you were mocking up `BRIGHTBLOB` to be `2**0` for the mocks, note that you'll now have to make it `2**1`, or bit-shift it as I did [here](https://github.com/desihub/desitarget/blob/ADM-more-DR8-updates/py/desitarget/io.py#L252-L254).

geordie666:
@moustakas: I also restored the meaning of the `qso_selection==colorcuts` option for the SV cuts. It should now be straightforward to bypass the QSO Random Forest selection when creating SV mocks, e.g.:

```
select_targets(infiles, survey='sv1', qso_selection='colorcuts', ..., ...)
```

moustakas:
Thanks @geordie666.  The targeting catalogs do not contain the (new) `PSFDEPTH_W[1,2]` columns -- do we want these?

geordie666:
`PSFDEPTH_G, R, Z` and `GALDEPTH_G, R, Z` are not used for the targets. They *are* produced in the random catalogs, and then those randoms are used in QA, but they aren't touched in the target selection itself.

So, guessing that `PSFDEPTH_W1, W2` will be the same, and with a view to not further unnecessarily expanding the target files, I'm not going to propagate `PSFDEPTH_W1, W2`. For now.

sbailey:
Is this PR backwards compatible with running TS on DR6 and DR7, in addition to supporting the data model changes for DR8?

I'm ok with not including PSFDEPTH_* in this round, but we should revisit that in the broader picture of what should (not) be included.  The original guidelines for what should be included in TS:
1. columns needed for provenance (BRICKID, TARGETID, RELEASE, ...)
2. columns used by target selection
3. columns likely needed for key downstream analyses (QA and/or science)

PSFDEPTH_* seems in the (3) category.  Of course this was before the TS files ballooned, so now were revisiting what *really* needs to be included.

geordie666:
Trying to condense a bunch of discussion on-and-off-list into a conclusion:

I'm confident that this PR IS backwards-compatible with DR6 and DR7 files (I'll check one last time before merging). And, I won't propagate propagate `PSFDEPTH_W1, W2` for targeting, but I will add them to the sweeps in the legacypipe repo.

@moustakas: Unless you disagree I'll merge #493, #494, and #496 in an hour or two after running a few final tests.

Issue #492, opened 2019-05-07
support SV targeting bits in select_mock_targets


moustakas:

moustakas:
Done in #496.

PR #496, opened 2019-05-09
add SV support to select_mock_targets

moustakas:
Addresses #492, although still testing for corner cases.


moustakas:
OK with my tests and a standard yaml configuration file this branch runs to completion by simply passing `--survey sv1` to either `bin/select_mock_targets` or `bin/mpi_select_mock_targets`.

@sbailey Do you want to test or would you like me to simulate a chunk of sky?

I recommend that contaminants be excluded (commented out from the yaml file) since the densities have not at all been tuned for SV and so I think that would just waste compute time.  

Next step is to adjust the densities of the various nominal target classes, but that's outside of the updates here.

geordie666:
@sbailey: If you're going to merge this, please merge #494, upon which this PR builds, first.

PR #502, opened 2019-05-27
Fix I/O bug on some operating systems when using Python 3.6

geordie666:
This PR fixes an I/O bug resulting from how Gaia fields were being dropped in some rec arrays. The fix was to default to using `numpy.lib.recfunctions` (this will make the custom functions that needed to be fixed slower for large arrays, but the functions in question typically do not operate on large arrays).

The bug was identified by @Cyeche running `select_targets` on his laptop. The bug wasn't affecting Python 3.5 or Python 3.6 at NERSC, but I could recreate it in a custom environment for Python 3.6.8, on my Mac.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,8 @@ desitarget Change Log
 0.29.2 (unreleased)
 -------------------
 
-* Apply the same declination cut to the mocks as to the data [`PR #501`_]. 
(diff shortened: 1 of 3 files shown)


geordie666:
@Cyeche: can you check that this PR fixes the bug that was being triggered on your laptop? If this fix works, please merge the PR.

Cyeche:
@geordie666 
It works, Thanks!

PR #505, opened 2019-06-04
add random forest QSO selection and REF_CAT to select_mock_targets

moustakas:
Follow-up PR to #496 with changes/bug fixes requested/identified by @sbailey.

1. Select QSO targets using random forest (it's slower, but necessary for properly populating SV targeting bits.

2. Bug fix: `MWS` targeting bits were not being set because the `REF_CAT` column, while present, was not being populated correctly.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -10,6 +10,10 @@ desitarget Change Log
 0.30.0 (2019-05-30)
 -------------------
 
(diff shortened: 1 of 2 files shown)


moustakas:
These changes only touch `mock.mockmaker`, so I will self-merge after tests pass so @sbailey can continue pushing forward on the SV data challenge.

Issue #508, opened 2019-06-06
relative normalization of tracer and lya qsos is not correct in select_mock_targets


moustakas:
There is a bug in how the relative densities of the tracer and Lya QSOs are normalized in `select_mock_targets`, which you can clearly see in the figure below.  This is a bug arising from the fact that we *simulate* Lya QSOs all the way down to `z=1.8`, but the *target densities* are defined at a redshift cut-off of `z=2.1`.  So in the mocks we end up with too few Lya QSOs (e.g., relative to the expected redshift distribution, shown as the open gray histogram).

Fix this.

<img width="566" alt="Screen Shot 2019-06-06 at 6 13 57 AM" src="https://user-images.githubusercontent.com/1431820/59025606-8d283980-8822-11e9-8bf2-89abac5d6ff1.png">

moustakas:
Addressed in #509.

PR #509, opened 2019-06-06
fix normalization bug in QSO tracer/Lya mock target densities

moustakas:
Addresses #508.  

Below is the updated distribution from just a couple `nside=32` healpixels, which has the correct/expected shift (more *mock* Lya QSOs relative to *mock* tracer QSOs).

Screen Shot 2019-06-06 at 8 09 47 AM

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -10,6 +10,7 @@ desitarget Change Log
 0.30.0 (2019-05-30)
 -------------------
 
+* Fix normalization bug in QSO tracer/Lya mock target densities [`PR #509`_].  
 * Follow-up PR to `PR #496`_ with two changes and bug fixes [`PR #505`_]:
     * Select QSO targets using random forest by default.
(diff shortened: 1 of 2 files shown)


moustakas:
@sbailey Please merge this before you continue your SV data challenge tests.

geordie666:
@moustakas: I think the changes log is out-of-order. Both #505 and #509 should be after the `0.30.0` tag. For example, see the same file in [PR507](https://raw.githubusercontent.com/desihub/desitarget/eda1d36d81814689403665cbda1b2abe6d9e9acd/doc/changes.rst).

No particular need to fix this here, as I'll fix it in #507...just a note in case you add more quick PRs before #507 is merged.

moustakas:
Thanks @geordie666.  That was me rushing and not checking that a new tag had been cut.  I propose we merge this PR as-is (tests pass and changes only affect `select_mock_targets` once again) since you'll fix the log in #507.

PR #510, opened 2019-06-20
Add locations for "supplemental" (outside-of-footprint) sky locations

geordie666:
Adds (parallelized) support for sky locations that are outside of the Legacy Surveys footprint. Works by:
- Randomly populating sky area beyond some minimum Declination and beyond some Galactic b.
- Then avoiding all Gaia sources out to some specified radius.

Also fixes a bug where `desitarget.geomask.hp_in_box()` was using geodesics instead of small circles for Dec cuts.

This is the simplest approach where we don't have Legacy Surveys imaging. A further augmentation might be to sink apertures in the WISE imaging to check for bright stars. That would introduce other issues based on WISE-optical colors, but the code could at least report the WISE flux (if necessary).


moustakas:
I peaked at the code, but can you post some QA demonstrating the new functionality?  Great development!

geordie666:
@moustakas: What sort of QA would you like? It's fairly easy to demonstrate that these are random locations that are 2" from any Gaia source. Would you like a .png showing that?

moustakas:
Nmind, that sounds like busy work -- I was just trying to provide a more critical (read: useful) look at the new code, but if you're satisfied then let's run with it.

PR #513, opened 2019-06-27
Preliminary ELG cuts for DR8 imaging for main and SV

geordie666:
This PR is a preliminary update to the ELG selection in preparation for DR8.

It may be the case that we can deprecate the 4 "sub-bits" for the ELG SV selection, but I'm retaining those bit-slots in the SV yaml files, for now, in an attempt to not break downstream code. If the final ELG SV selection only demands one bit, then I'll update the SV yaml file at that point.

### DIFF
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -304,10 +304,10 @@ def isELG_colors(gflux=None, rflux=None, zflux=None, south=True, primary=None):
 
     # ADM work in magnitudes instead of fluxes. NOTE THIS IS ONLY OK AS
     # ADM the snr masking in ALL OF g, r AND z ENSURES positive fluxes.
-    g = 22.5 - 2.5*np.log10(gflux.clip(0))
-    r = 22.5 - 2.5*np.log10(rflux.clip(0))
-    z = 22.5 - 2.5*np.log10(zflux.clip(0))
-    
+    g = 22.5 - 2.5*np.log10(gflux.clip(1e-16))
+    r = 22.5 - 2.5*np.log10(rflux.clip(1e-16))
+    z = 22.5 - 2.5*np.log10(zflux.clip(1e-16))
+
     # ADM cuts shared by the northern and southern selections.
     elg &= g > 20                       # bright cut.
     elg &= r - z > 0.3                  # blue cut.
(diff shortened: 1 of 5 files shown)


geordie666:
@moustakas: A heads-up...when this is merged it will almost certainly break the mocks, as I'm now passing `MASKBITS` as an extra parameter in `cuts._prepare_optical_wise()`.

geordie666:
I've asked Anand for a review offline. I'll merge this once he confirms that I have the cuts correct.

geordie666:
@moustakas: I'm about to merge this, which, I think, will break the mocks.

At some point soon I'm going to need to make a tag with which to run DR8 targeting, and presumably @sbailey would like it if tagged versions didn't break his integrated unit tests. So, it would be excellent if you could make the few fixes required for the mocks relatively soon. 

Otherwise, this is a heads-up to @sbailey that he might have to bypass the mocks in his unit tests, which I think he's done before.

py/desitarget/cuts.py
```diff
@@ -255,91 +256,72 @@ def isLRGpass(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None,
...
 
-    elg &= notinELG_mask(gallmask=gallmask, rallmask=rallmask, zallmask=zallmask,
-                         brightstarinblob=brightstarinblob, primary=primary)
+    elg &= notinELG_mask(maskbits=maskbits, gsnr=gsnr, rsnr=rsnr, zsnr=zsnr,
+                         primary=primary)
 
-    elg &= isELG_colors(gflux=gflux, rflux=rflux, zflux=zflux, w1flux=w1flux, w2flux=w2flux,
-                        south=south, primary=primary)
+    elg &= isELG_colors(gflux=gflux, rflux=rflux, zflux=zflux, south=south,
+                        primary=primary)
```
moustakas: Yeah, these changes broke the mocks.  Even though `w1flux` and `w2flux` aren't used for ELG target selection, the way the targeting cuts are applied to the mocks is to pass all the *grz* photometry independent of the class of object.

geordie666: Oops, my bad. I should have remembered *that*. I'll restore this branch and fix that in the next 20 minutes.

geordie666: Actually, I'll just fix this in a separate PR for better transparency. I'll ask for your review in the next PR.

moustakas: I have the changes in a branch already and I'm testing other permutations of the code. I can open a simple PR (or even push the very minor change to master...).  Let me know.

geordie666: Will fix in #514. Thanks.

PR #514, opened 2019-06-28
restore w1/w2 flux argument to iselg_colors

moustakas:
Near-trivial fix for an unintended change to the `isELG` API introduced in #513.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,6 +5,8 @@ desitarget Change Log
 0.30.2 (unreleased)
 -------------------
 
+* Near-trivial fix for an unintended change to the isELG API introduced in `PR
+  #513`_ [`PR #514`_]. 
 * Preliminary ELG cuts for DR8 imaging for main and SV [`PR #513`_].
     * Don't deprecate wider SV bits, yet, ELGs may still need them.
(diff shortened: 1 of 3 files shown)


geordie666:
I think this will work, now. Strictly, for completeness, you might want to pass `w1flux` and `w2flux` in the main survey cuts code, too:

https://github.com/desihub/desitarget/blob/d151c9182b448a4fdea795010cce4320e7538918/py/desitarget/cuts.py#L259

That's your call (I think the code will work without failing, either way). Feel free to merge at your leisure.

moustakas:
Although I accidentally retriggered tests, I'm going to go ahead and merge.  I tested `select_mock_targets` with these changes with all target classes for both the main and sv1 surveys.

Thanks for your help, @geordie666.

geordie666:
Excellent. If you don't think the mocks are going to fail on `maskbits` now being passed out of `_prepare_optical_wise`:

https://github.com/desihub/desitarget/blob/01c930ae5388a3583e80f754bc8f63282046a18b/py/desitarget/cuts.py#L1938-L1947

Then we should be good to go.

Apologies for breaking this at such short notice.

geordie666:
Sorry, missed your last comment. If `select_mock_targets` works, then all is good.

Issue #517, opened 2019-07-11
Add CMX ELG, LRG, QSO, BGS target bits


sbailey:
Add CMX targeting bits for ELG, LRG, QSO, BGS for commissioning tests of galaxy/QSO targets for pipeline testing, before (pre-)SV.

geordie666:
Addressed in #519.

PR #519, opened 2019-07-16
Include further SV-like selections for commissioning

geordie666:
This PR adds the ELG/LRG/QSO/STD early SV ("SV0") selections for commissioning. 

It also includes a more minor update to ensure that (the few) MWS targets that have `NaNs` in columns that were added to the imaging data model as part of DR8 are not targeted.

### DIFF
--- a/py/desitarget/cmx/data/cmx_targetmask.yaml
+++ b/py/desitarget/cmx/data/cmx_targetmask.yaml
@@ -6,9 +6,12 @@ cmx_mask:
     - [STD_TEST,        2, "Very bright stars for early tests", {obsconditions: DARK|GRAY|BRIGHT}]
     - [STD_CALSPEC,     3, "Matches to CALSPEC stars", {obsconditions: DARK|GRAY|BRIGHT}]
 
-    # ADM BGS and MWS bits that should resemble at least the initial SV selections ("SV0"), 
+    # ADM targeting bits that should resemble at least the initial SV selections ("SV0"),
     - [SV0_BGS,         8, "Any SV-like BGS (North) bit is set (from initial SV selection)", {obsconditions: BRIGHT}]
     - [SV0_MWS,         9, "Any SV-like MWS (North) bit is set (from initial SV selection)", {obsconditions: BRIGHT}]
+    - [SV0_LRG,        10, "Any SV-like LRG (North) bit is set (from initial SV selection)", {obsconditions: DARK}]
+    - [SV0_ELG,        11, "Any SV-like ELG (North) bit is set (from initial SV selection)", {obsconditions: DARK|GRAY}]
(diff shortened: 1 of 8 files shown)


geordie666:
Addresses #517.

py/desitarget/cmx/data/cmx_targetmask.yaml
```diff
@@ -6,14 +6,19 @@ cmx_mask:
     - [STD_TEST,        2, "Very bright stars for early tests", {obsconditions: DARK|GRAY|BRIGHT}]
     - [STD_CALSPEC,     3, "Matches to CALSPEC stars", {obsconditions: DARK|GRAY|BRIGHT}]
 
-    # ADM BGS and MWS bits that should resemble at least the initial SV selections ("SV0"), 
-    - [SV0_BGS,         8, "Any SV-like BGS (North) bit is set (from initial SV selection)", {obsconditions: BRIGHT}]
-    - [SV0_MWS,         9, "Any SV-like MWS (North) bit is set (from initial SV selection)", {obsconditions: BRIGHT}]
-
+    # ADM targeting bits that should resemble at least the initial SV selections ("SV0").
```
sbailey: Nit-picking: is "SV0" effectively "CMX", and then when we start "real" SV we continue with "SV1"?  Or will `desitarget.sv0` appear to use as the "pre-SV-during-CMX" observations instead of committing ourselves to using the SV1 bits for those Nov/Dec observations?  It seems odd and potentially problematic to have "SV0" appear in the CMX target bit names, especially if SV0 might become a real thing with slightly different bit definitions.  My understanding is that these bits are for getting targets for commissioning the spectro pipeline even before the "pre-SV" observations, which would use different bits (including the extended selections).

i.e. would you object to just having these be "BGS", "MWS", "LRG", etc. and their presence in the CMX_TARGET bitmask is the indicator that they go with CMX?  Or "CMX_BGS", etc. if you want to be really clear?

geordie666: My goal is to distinguish classes that only appear in CMX (such as the `STD_TEST` and `STD_CALSPEC` classes) from target classes that are SV-like. So, I'd argue we need some `CMX_ONLY`, `CMX_PLUS` scheme. I'd rather label the classes to which you're referring `SV0` than `CMX_PLUS`, as they are "SV-like" selections. That's my logic, at least. Note that this scheme first appeared almost a year ago:

https://github.com/desihub/desitarget/commit/57959f039638fe080849b79965d3947fcd89f882

Not that that means we shouldn't change it, but it has been in circulation for a while.

geordie666: After an in-person discussion with @sbailey I'm leaving the bit names as-is and merging this.

Issue #400, opened 2018-10-12
Targets that are both LRG+BGS have NUMOBS_MORE always set to 1 even in dark time


akrolewski:
Hi all,

It appears that all BGS targets have NUMOBS_MORE reset to 1 in calc_numobs().  However, some BGS targets are overlaps with LRGs.  Therefore, these targets will continue to be observed in dark time (e.g. when running quicksurvey) and therefore you wind up with a tail of LRGs that are observed > 2 times in dark time (because they are BGS overlaps).  This doesn't make sense; either these are LRGs, in which case they should be observed once or twice in dark time, or they are BGS, in which case they should be observed in bright time.  The program information Dark, Gray, Bright should be passed to calc_numobs() and calc_priority() to take these cases into account.  More generally, it doesn't appear as if there is a way to properly treat overlapping targets (such as LRG+BGS).  After the first observation, one should know whether they are truly LRG or BGS and then the target can be observed appropriately.

I'm wondering what the best way is to solve this problem (or set of problems).

geordie666:
There is some additional discussion of this in #429. It is possible that we will deprecate this behavior (BGS targets always have `NUMOBS_MORE` reset to 1) completely. Otherwise, we'll have to think carefully about how to achieve the desired behavior in fiberassign.

geordie666:
Addressed by having MTL take an observing condition/layer in #531.

PR #523, opened 2019-07-25
Add OBSCONDITIONS to target files.

geordie666:
This PR adds the `OBSCONDITIONS` bitmask to output target files, instead of it being added in MTL. As part of this modification, it also:

- Splits target files explicitly into two sub-directories corresponding to "bright" and "graydark" surveys.
- Defaults to splitting target files in such a manner for SV and the Main Survey (not for commissioning).

Coincidentally, this PR:
- Adds an informational bit for supplemental sky locations.


geordie666:
@sbailey: Last week we decided that the idea of adding `OBSCONDITIONS` to the target files and splitting the targets into separate bright/graydark files had finally had time to percolate and so we should go ahead and do it. You noted that we should retain the bright-time targeting bits in the dark-time file and vice-versa for easier tracking of matching bright/dark targets. This PR makes those updates.

However, this done _not_ solve issues such as #400, or the issue that quasar targets can be set to have 4 observations in bright time. This is _because_ we retain the bright-time targeting bits in the dark-time file and vice-versa, and MTL operates on the targeting bits, oblivious to the `OBSCONDITIONS`.

If we want to ensure that only bright-time `PRIORITY`/`NUMOBS_MORE` decisions are applied to bright-time objects (and dark-time `PRIORITY`/`NUMOBS_MORE` decisions to dark-time objects) then I think there are only two solutions:

1. As also noted in #400, give MTL some knowledge of the desired `OBSCONDITIONS` that it can use to follow different logic for setting `PRIORITY`/`NUMOBS_MORE`.

2. Only write out the dark-time targeting bits to the dark-time file and only write out the bright-time targeting bits to the bright-time file. Any knowledge of a match would be retained in `TARGETID` but post-processing would be required to look for bright-time targets with dark-time spectra and vice-versa. We would also have to deprecate any ` {obsconditions: BRIGHT|GRAY|DARK}` choices for bright-time targets and fix them to only be observed in ` {obsconditions: BRIGHT}`.

Advice and guidance?

sbailey:
I think I prefer your option (1):

> 1. As also noted in #400, give MTL some knowledge of the desired OBSCONDITIONS that it can use to follow different logic for setting PRIORITY/NUMOBS_MORE.

Would it be meaningful to encode in a header keyword which OBSCONDITIONS each target file is for, and have MTL use that instead of trusting the user to remember (while still providing the ability to override it)?  Or otherwise make mtl require explicit obsconditions and fail if the user doesn't provide it?  i.e. let's avoid a situation where the default arguments result in the wrong behavior.

geordie666:
That would be meaningful, but aren't people passing tables/arrays to MTL rather than files?

I'll make it explicit that this needs to be passed (and fail without it), but trust the user to have read the appropriate header keyword (which is mostly coded up already [in the current PR](https://github.com/desihub/desitarget/blob/ADM-add-OBSCON/py/desitarget/io.py#L324-L336).

Can we merge this and I'll get back to fixing issue #400 later? I've started to work on other things while this PR was dangling. It's a complete PR, as it has improved MTL without addressing #400.

sbailey:
Good point.  This PR doesn't fix everything related to OBSCONDITIONS but it is a step in the right direction without a corresponding step backwards, so ok to merge it.

Issue #522, opened 2019-07-22
Extra targetting bit  BGS_FAINT_HIP


ShaunMCole:
In BGS, BGS_BRIGHT  (r<19.5) have higher targeting priority than BGS_FAINT (19.5<r<20.0) but we need (explained below) to make a random 10% subset of the faint class of objects the same priority as BGS_BRIGHT.   Adam suggested the way to do this is to assign an additional tragetting bit, say BGS_FAINT_HIP that will be assigned to 10% of the BGS_FAINT objects. Then have the fibre targeting code treat these objects the same as the BGS_BRIGHT objects.

Rationale:
     In BGS there are two (main) target classes BGS_BRIGHT and BGS_FAINT and these are given different priorities when passed to the fiber allocation software.When we first tried this out we realised that our targetting strategy had a flaw that meant the PIP weights (developed by Will and Davide) wouldn't work.  Basically in some overdense regions, i.e. galaxy clusters, there are so many BGS_BRIGHT objects that they saturate the fibers and no matter how many times you run the algorithm none of the BGS_FAINT objects receive a fiber.
This screws up the PIP weights as some pairs have zero probability of being observed and so no correction can be made. One ends up with large bias.

   The fix we came up with in Alex Smith's paper was to promote a random 10% of the BGS_FAINT things to have the same priority as BGS_BRIGHT. This solves the problem. In Alex's paper we did this by hand with our own branch of the target assignment code, but I think we need to find a way of doing this in the official version.

geordie666:
Addressed in #525.

PR #525, opened 2019-07-30
 Update Commissioning and Main Survey target classes.

geordie666:
Includes:
- New `SV0_WD` and `SV0_STD_FAINT` target classes for commissioning.
- Mild updates to the `SV0_BGS` and `SV0_MWS` target classes for commissioning.
- New `BGS_FAINT_HIP` Main Survey class. These are `BGS_FAINT` targets at the `BGS_BRIGHT` priority.
- Explicit checking on `ASTROMETRIC_PARAMS_SOLVED` for MWS targets that use proper motion and parallax.
- Adds a 3-sigma parallax slop in the `MWS_MAIN` survey target class.

Should address #522.


sbailey:
Looks like this also drops the BGS_KNOWN_XXX bits.  Any comments on that?  e.g. will those be brought in via secondary targets instead as needed?  Otherwise looks fine; I fixed merge conflicts in changes.rst, which has triggered another round of tests.

geordie666:
Yes, secondary targets is exactly the way to bring in KNOWN objects. My use cases for secondary targets are:

(1) Not based on sources in the sweeps files.
(1a Outside of the footprint).
(2) Based on an RA/Dec cross-match.

The `BGS_KNOWN` bits aren't defined anywhere, as far as I can tell. So, I decided that they were probably just placeholders.

sbailey:
Sounds good.  documentation build tests are failing due to docs.astropy.org being down (522, meaning sort of up but not really).  Will retry later.  OK to merge now if that is a blocking factor for proceeding with other work.

geordie666:
Let's hold off merging this for a day. The MWS folks may not want one of the flags I've added, but I'll change that tomorrow if needed (it's a one-liner).

sbailey:
Two days have passed without an outcry, so merging now.  We can revert later if needed, since I'm sure that there are other target cuts coming down the pipe too.

Issue #520, opened 2019-07-18
Add in UCAC/URAT proper motions where Gaia is missing


geordie666:
In 1.5% of the sky we have no proper motions from Gaia for our "typical" GFA objects. For GFA targets with missing Gaia proper motions, supplement with [the URAT catalog](http://cdsarc.u-strasbg.fr/viz-bin/Cat?I/329).

djschlegel:
Has anyone used this URAT catalog yet?  The only place I can seem to find the files are in the v12 directory of http://cdsarc.u-strasbg.fr/ftp/I/329/URAT1/ , but it looks like that's a straight binary file (not FITS binary, just raw binary).  Do-able, but really a pain.  And who knows, maybe they made these files on a VAX with a mix of big-endian and little-endian... :<

sbailey:
I haven't successfully parsed these files, but a bit of other info:

data model: http://cdsarc.u-strasbg.fr/ftp/I/329/ReadMe

directory with fortran code for reading them and 2 files extracted into ascii to cross check: http://cdsarc.u-strasbg.fr/ftp/I/329/URAT1/access/

astropy code that knows how to parse the ReadMe, but apparently is intended for reading the ascii version of the output not raw binary: https://docs.astropy.org/en/stable/api/astropy.io.ascii.Cds.html

Ugh.

geordie666:
I modified the provided fortran code to successfully parse the binary files, and verified that I could recreate the 2 files extracted into ascii.

@sbailey: Hopefully there will be no licensing problem, or DESI software standard, to prevent me committing my version of the fortran to git, when I'm done incorporating URAT into desitarget.gfa.

ameisner:
What about HSOY?

https://arxiv.org/pdf/1701.02629.pdf

It's a Gaia DR1 + PPMXL hybrid. That's what I used for my unWISE astrometric recalibrations before Gaia DR2 was available.

-Aaron

sbailey:
> I modified the provided fortran code to successfully parse the binary files, and verified that I could recreate the 2 files extracted into ascii.
> 
> @sbailey: Hopefully there will be no licensing problem, or DESI software standard, to prevent me committing my version of the fortran to git, when I'm done incorporating URAT into desitarget.gfa.

They don't provide an explicit license, but as long as you very clearly credit them with the original code, and note that you had to modify it and where the original is, I think we're fine.

[...]

PR #526, opened 2019-08-02
Add URAT catalog information

geordie666:
This PR adds [the URAT catalog](http://cdsarc.u-strasbg.fr/viz-bin/Cat?I/329) to desitarget. It includes:

- A new module `uratmatch.py` to retrieve URAT data from Vizier/FTP and reformat it.
- Code to efficiently match RAs/Decs to URAT, as part of the `uratmatch.py` module.
- Code to substitute URAT proper motions for GFAs where Gaia has not yet measured proper motions (the other Gaia information is retained for a GFA target, but zero/NaN proper motions are overwritten with URAT values, and the URAT_ID and separation is recorded).

This addresses #520. Here's an example of a general region we uncovered that had zero objects in a guide camera, which now has multiple objects from URAT:
URATGFAs


py/desitarget/fortran/README
```diff
@@ -0,0 +1,17 @@
...
+ADM-v1dump.f: FORTRAN code to parse UCAC/URAT files, adapted from
+              http://cdsarc.u-strasbg.fr/ftp/I/329/URAT1/access/v1dump.f
+              by Adam D. Myers (University of Wyoming).
+
+v1sub.f: Routines to help parse UCAC/URAT files, taken directly from
+         http://cdsarc.u-strasbg.fr/ftp/I/329/URAT1/access/v1sub.f
+
+v1dump: An executable compiled at NERSC via:
+
+        gfortran -o v1dump ADM-v1dump.f v1sub.f
```
sbailey: Thanks for including specific instructions for compiling v1dump.  Please also include documentation somewhere about how you ran it and were the outputs are that most
users should just use without having to run this themselves.

py/desitarget/uratmatch.py
```diff
@@ -0,0 +1,612 @@
...
+
+    Returns
+    -------
+    :class:`str`
+        The directory stored in the $URAT_DIR environment variable.
+    """
+    # ADM check that the $URAT_DIR environment variable is set.
+    uratdir = os.environ.get('URAT_DIR')
+    if uratdir is None:
+        msg = "Set $URAT_DIR environment variable!"
```
sbailey: Let's give the user a hint of where that directory should be when running at NERSC.  Please add it to `etc/desitarget.module` so that our code installations at NERSC will automatically set it.

py/desitarget/fortran/README
```diff
@@ -0,0 +1,17 @@
+The files in this directory are:
+
+ADM-v1dump.f: FORTRAN code to parse UCAC/URAT files, adapted from
+              http://cdsarc.u-strasbg.fr/ftp/I/329/URAT1/access/v1dump.f
+              by Adam D. Myers (University of Wyoming).
+
+v1sub.f: Routines to help parse UCAC/URAT files, taken directly from
+         http://cdsarc.u-strasbg.fr/ftp/I/329/URAT1/access/v1sub.f
+
+v1dump: An executable compiled at NERSC via:
```
sbailey: You probably didn't intend to add the compiled v1dump file to git.  Please remove it.
(Adding this comment to the README, where it is good to mention the executable, since I can't seem to github comment directly on a binary file itself).

geordie666: Ooops, sorry. Should I blast that with the BFG?

py/desitarget/fortran/README
geordie666: Done

geordie666:
Note that it will now be required to have the URAT catalog to run GFAs. But, it was already required to have the suitably-formatted Gaia catalog to run GFAs. The second (Gaia) is a far greater burden that the first (URAT), so I'm comfortable with it being a requirement.

py/desitarget/fortran/README
sbailey: It's only 30k so not super critical to blast with the BFG, but if you are comfortable with the BFG and want to use it, go for it.

geordie666:
I'd be 100% against moving the Fortran directory to the top level. I don't want to give the impression that we're expecting to add more Fortran code. In other words, Fortran is a "special case" under Python, here. The Python is a wrapper on it. Moving the Fortran to a top-level directory would imply that it has equal and substantial weight.

I also don't want to move everything into a urat directory, which would beg the question of why the `gaiamatch` code is in the `py` directory but the `uratmatch` code is in `py/urat/`. That seems structurally weird, to me.

py/desitarget/uratmatch.py
geordie666: Done.

py/desitarget/fortran/README
geordie666: In the end I decided to be lazy as it's such a small file and will never make it into master. I'll blast it down the road if I feel energetic.

geordie666: Oh, I did remove it though.

geordie666:
@sbailey: I believe I addressed all of your requests. I also: 
- moved the fortran code to `etc/fortran` as I liked that suggestion best (I've just noticed you actually suggested `etc/urat`. Let me know if you feel strongly about that distinction.
- added a `--nourat` command line argument in the binary, so if people *really* don't want to run without URAT, they don't have to.

I'm running a few tests of these changes. I'll let you know when I'm done.

geordie666:
@sbailey: OK, I'm done with internal tests. Let me know if you want more changes.

sbailey:
Looks good; thanks.  Merging now.

Issue #524, opened 2019-07-26
WDs as secondary targets


apcooper:
I wonder whether everyone is on the same page re. selecting white dwarfs. Two things have happened 'recently':

- Boris has stated a preference for targeting white dwarfs based on a pre-selected catalog rather than a bunch of Gaia cuts. He has the parent catalog and the code to select subsets of it that we want to target in SV and the main survey. 
- @geordie666 has added the secondary target scheme to `desitarget`. 

So it seems natural enough to use the secondary target code to deal with the WD science samples, but I'm worried that will cause problems or inconsistencies with the selection of WD_STDs. The alternative is to have the main code read the WD catalogs as a special case, but I guess that might be considered ugly given the existence of the secondary target scheme.

I don't think there's any distinction in code yet between the science WDs and the WD_STDs. I guess that won't be the final state. Who's responsible for progress on this?

I think swapping the WDs to secondary targets would break the code because of that hard-wiring.

Sorry for a fuzzy issue here but it seems like we need a focused discussion on this. Apologies if that's already happened/happening elsewhere. Pinging @moustakas, @sbailey and @CManser. I don't think Boris is on github so I'll point him to this by email.

BGaensicke:
Just a bit more details on the WD selection function: we have compiled a catalogue of white dwarf candidates (https://arxiv.org/abs/1807.03315) which provides a probability of a Gaia object being a white dwarf. That probability is mapped into Gaia parameter spaced using SDSS spectroscopy, and hence difficult to add to the DESI targeting software.

I have taken the key selection criteria that only make use of the Gaia data to have a selection that will run in the DESI targeting software, 

https://desi.lbl.gov/trac/wiki/TargetSelectionWG/TargetSelection#WhiteDwarfsMWS-WD

which will add a small number (few %) of contaminants (mainly quasars) as long as we stay away from crowded regions (that's the |b|>20 cut). 

If there is a way to provide the white dwarfs (as science and standard targets) simply in the form of a RA, Dec, proper motions, epoch (with associated Gaia IDs), that would work also well.

geordie666:
I think we will want the WD standards to be run through the primary selection code, and Boris' selection criteria are already coded up here:

https://github.com/desihub/desitarget/blob/master/py/desitarget/cuts.py#L786-L824

It's fine, I think, to have a different class of WDs for standards and for science targets. 

[...]

PR #530, opened 2019-08-29
Stellar SV

apcooper:
Tracking changes required for Stellar (MWS) SV. **Still work in progress, but probably close to final.**

@geordie666, if you have time to take a look at this and check that I'm on the right track that would be very helpful. I think we're just missing one or two additional 'special' classes from the secondary target mask and no more changes to the algorithms will be needed.

This is separate from the issue of getting the superbright time backup program implemented. This PR is only about the regular stellar SV, even though the secondary target mask preemptively includes a bunch of classes with `POOR|TWILIGHT` obsconditions.

Includes:
 - Move MWS_WD_SV (science WDs) to the secondary target mask but retain the algorithmic selection for WD standards (https://github.com/desihub/desitarget/issues/524)
 - Updates `sv1_cuts.py` to reflect that change.
 - Simplifies the MWS main selection in `sv1_cuts`, no need for a color selection now, makes 'sv' explcit in the function name for good measure.
 - Adds a missing bright limit to the nearby sample sv selection. 
 - Add lots of MWS_CALIB_* and MWS_SPECIAL_* target classes to the secondary target mask.

Notes to self on hacks that need to be undone before merging:
 - Standard secondary targets (VETO, BACKSTOP etc.) commented out for testing, otherwise slow.
 - Need to clean up PEP8 whitespace issues.


geordie666:
@apcooper: Apologies for the tardiness in getting to this review, but NERSC has been down for a few days and I want to run everything at something close-to-scale before providing a full review. Some thoughts while I wait for NERSC though:

- You're absolutely on the correct track, and I thank you for incorporating the MWS selections into the secondary mechanism.
- Feel free to delete the secondary bit `BACKSTOP` completely. This was a placeholder for the superbright/backup time program. As you noted, it's very slow to run this with the secondary mechanism, so let's get rid of it. Retain `VETO` as bit 0 and `DR16Q` as bit 1.
- Don't worry about implementing the superbright/backup time program directly yourself. I can incorporate this into SV and the Main Survey (although the MWS group will have to provide the cuts). I've already implemented something like it for commissioning, with parallelization across nodes at NERSC to speed it up. This will be the only Gaia-only program that I run through the "primary" mechanism.
- Related to the above point, don't forget that we also need cuts (using only Gaia) for standard stars for the superbright/backup program. I've already talked to Arjun about adding that to the wiki, too. I can also add this to the "primary" code (for you to review or iterate on), but I'll need the cuts from the MWS group.
- I'd prefer it if your bits in the `sv1_mws_mask` did not contain the text "SV1" so, e.g., `MWS_MAIN` instead of `MWS_MAIN_SV` etc. This makes some error-checking easier to automate (and these sources' targeting bits are in a column called `SV1_MWS_TARGET` so there's no ambiguity). I'm fine with you leaving the "SV1" snippet in the bit-names for the "secondary" part of SV.
- Thanks for catching the bright-end cut in the `MWS_NEARBY` sample! I left the note about that in the code a while back and then forgot to check up on it with you.

I'll review the code line-by-line (if needed) when NERSC is back up and I can run this to scale. But, this looks great for now. Remember to update the wiki with your new selections and to update the [changes file](https://github.com/desihub/desitarget/blob/master/doc/changes.rst) when you feel this PR is ready for final review.

geordie666:
@apcooper: One last thing. So I can test all of this properly, can you add your files (and associated documentation) to the official "secondary" directory at NERSC, which is:
```
/project/projectdirs/desi/target/secondary/sv1/indata/
/project/projectdirs/desi/target/secondary/sv1/docs/
```
These need not be final (we can update them later).

apcooper:
Thanks for the review @geordie666, those points should be easy to deal with.

I'm confused about the following related issue and would appreciate your help. This function
https://github.com/desihub/desitarget/blob/035b667c0c4a8072b961e0c88cd9bcdc4b86ccef/py/desitarget/targets.py#L262-L263
is called in two places; in `make_mtl`
https://github.com/desihub/desitarget/blob/035b667c0c4a8072b961e0c88cd9bcdc4b86ccef/py/desitarget/mtl.py#L161-L162
and in `finalize_secondary`:
https://github.com/desihub/desitarget/blob/8fd1513c5a97176b52875542b2eb29fdd32f1007/py/desitarget/secondary.py#L690-L691
When secondary targets are folded into the MTL, this means they get passed through `set_obsconditions` twice, first with `scnd=True` and then again in `make_mtl` with `scnd=False` (as part of the combined MTL table).

Since `scnd` changes the logic in `set_obsconditions` I'm wondering if this is a bug. What do you think?

Assuming that it's wrong, my guess for a fix is to call `set_obsconditions` twice in `make_mtl`, once with `scnd=False` and once with `sncd=True`, filtering the targets by `SCND_ANY = 0` and `!=0` respectively.

What is the reason for the `scnd` logic in `set_obsconditions`? Why do secondary target obsconditions only depend on the [-1] column returned by `main_cmx_or_secondary` but primary targets (including primary targets with secondary bits set?) depend on all the target bit columns?

Knowing that would also help me to understand why `set_obsconditions` is being called in `finalize_secondary` at all.

I'm looking into this because I found a bunch of secondary targets in my tests (notably those outside the LS footprint and hence not matched to primaries) are getting obsconditions = 0 (hence ignored by fiberassign) for no obvious reason.

apcooper:
> @apcooper: One last thing. So I can test all of this properly, can you add your files (and associated documentation) to the official "secondary" directory at NERSC, which is:
> 
> ```
> /project/projectdirs/desi/target/secondary/sv1/indata/
> /project/projectdirs/desi/target/secondary/sv1/docs/
> ```
> 
> These need not be final (we can update them later).

Double-checking: do you want the actual files copied or are symlinks ok?

geordie666:
@apcooper: Regarding your second question about symlinks...we absolutely need to copy the files across to the directory that I indicated rather than have symlinks. At some point we will need to "freeze" the secondary directory for SV, make it inviolate, and run check-sums. If we have symlinks, we run the risk of something changing across the link. So, let's copy everything to the official directory.

Your first set of questions are a little trickier. I've answered them one-by-one below:

_What is the reason for the scnd logic in set_obsconditions? Why do secondary target obsconditions only depend on the [-1] column returned by main_cmx_or_secondary but primary targets (including primary targets with secondary bits set?) depend on all the target bit columns?_

The reason that `set_obsconditions` is called with `scnd=True` in `secondary.py` is because the objects processed through `secondary.py` will not (I think) have values set for the `DESI_TARGET`, `BGS_TARGET` and `MWS_TARGET` columns. So, setting `OBSCONDITIONS` on those other columns will just fail with a non-existent column error. It's not actually important whether or not we incorporate primary information when setting `OBSCONDITIONS` as any secondary target that matches a primary (and that does not have `OVERRIDE` set) will just revert to the qualities of the primary, inheriting the `TARGETID` of the primary but leaving the primary unaffected. As it should.

_Knowing that would also help me to understand why set_obsconditions is being called in finalize_secondary at all._

We have to set the `OBSCONDITIONS` in advance of passing targets to MTL so that fiberassign can create custom MTL files for SV. We also have to set `OBSCONDITIONS` _in_ MTL because it is not a column that is stored downstream of targeting (the input to MTL only requires the `TARGETID`, `_TARGET` and `_INIT` columns).

_Since scnd changes the logic in set_obsconditions I'm wondering if this is a bug. What do you think?_

This is not a bug, per se, in that the intended behavior is correct for primary targets. I think, though, that you've identified an oversight on my part (**thank you!**) that crept in when I was simultaneously working on the `obscon` updates for both secondary targets and MTL. As you note, secondary-only targets will not have meaningful values of `OBSCONDITIONS` once they are passed through MTL. I guess that is a bug, really!


_Assuming that it's wrong, my guess for a fix is to call set_obsconditions twice in make_mtl, once with scnd=False and once with sncd=True, filtering the targets by SCND_ANY = 0 and !=0 respectively._

I think you are almost correct. We don't want to change the `OBSCONDITIONS` state of targets with `SCND_ANY != 0` because we don't want to affect the `OBSCONDITIONS` state of a primary target. The primary targets should be inviolate and their state should not be affected by the secondary targets.

I think the check has to be, instead, on targets that do _not_ have primary information set. If you're passing things to `mtl.make_mtl()` using the `scnd=True` keyword I set up, then in this loop:

https://github.com/desihub/desitarget/blob/1436ec373b2b24244631f909e8a7e6c6a6df0e5c/py/desitarget/mtl.py#L60-L70

you need some call like:
```
obsconmaskscnd = set_obsconditions(targets, scnd=True)
```
and then you need to propagate that additional mask through the code and (correctly) combine it with `obsconmask` here:

https://github.com/desihub/desitarget/blob/1436ec373b2b24244631f909e8a7e6c6a6df0e5c/py/desitarget/mtl.py#L161-L162

Alternatively, if it doesn't seem hack-y, I think that the secondary-only targets will always end up with `DESI_TARGET==0`. So, you could stick with your original suggestion but instead of having a check on `SCND_ANY != 0` check instead on `DESI_TARGET==0` to set `OBSCONDITIONS` for the secondary-only targets.

Or, for another alternative, in the loop of code I included above (lines 60-70), retain the `SCND_TARGET` or `SV1_SCND_TARGET` column in `targets`, pass that through the code, and then have a check on `SCND_TARGET !=0` for setting `OBSCONDITIONS` for the  secondary-only targets. Note that you can use:

```
col, _, _ = main_cmx_or_sv(scnd, scnd=True)
svcol = col[-1]
```
to rigorously retrieve the secondary column, whether its `SCND_TARGET` or `SV1_SCND_TARGET`.

Finally, before I forget, could you also correct my documentation mistake and remove ``DESI_TARGET`` from this part of the docstring of `mtl.make_mtl()`:

https://github.com/desihub/desitarget/blob/1436ec373b2b24244631f909e8a7e6c6a6df0e5c/py/desitarget/mtl.py#L39-L44

"DESI_TARGET" should not be a required column for the `scnd` input, I think.

Sorry this is such a lengthy conversation, but thanks for soldiering on and getting all of the details correct for the secondary targets.

apcooper:
Many thanks for the super-helpful write-up. It looks like the fixes you suggest will solve the problem I'm seeing with the Hyades. However, I'm now confused by this:

> We don't want to change the OBSCONDITIONS state of targets with SCND_ANY != 0 because we don't want to affect the OBSCONDITIONS state of a primary target. The primary targets should be inviolate and their state should not be affected by the secondary targets.

I was expecting the opposite. I thought that, since a high-priority secondary match could increase the priority of a primary (right?), it should also mean the obsconditions for the 'combined' target should be an `or` of the obsconditions for the primary and secondary versions of that target.

Does this mean a secondary can't have more permissive obsconditions than the primary it matches? Perhaps that's not a worry in practice, except maybe where we have secondary targets we want to allow in twilight/poor conditions that also match to MWS_MAIN (or BGS!) primary targets.

Maybe there's an issue where e.g. MWS unwittingly define a secondary class that matches a bunch of dark-time LRGs, which we don't want to pull out of the dark time survey to observe in bright time. Or is that taken care of in some other way?

Separately, I'm also a bit confused on this less critical point:

>>Knowing that would also help me to understand why set_obsconditions is being called in finalize_secondary at all.

> We have to set the OBSCONDITIONS in advance of passing targets to MTL so that fiberassign can create custom MTL files for SV.

Sorry, this hasn't quite clicked for me.... isn't the function that creates the custom MTL for SV then the place to set the obsconditions?  (My simple picture of how this works is that FA always operates on the output of make_mtl, never on the target files directly; could be wrong about that?)

geordie666:
@apcooper: To your first point. I think it's dangerous from the perspective of Large Scale Structure measurements to allow the nature of a primary target to change. The classic example is an interesting, high-priority, bright WD that is actually a quasar. The quasar could get shunted into extra observations with `BRIGHT` set, but it's `NUMOBS_MORE` would run down to zero in bright-time because `BRIGHT` had become an allowed observing condition for the target. I haven't thought through all of the repercussions carefully, but such an approach just seems fragile...there are almost certainly cases where we'd regret this, I'd guess.

To guard against this, I think that high-priority secondary targets should always be sent with `OVERRIDE=True`. Such a secondary would be its own target even if it matched a primary. Then, because it's a high-priority secondary target, it should bump the existing primary during fiberassign. This approach seems safer, because as far as the LSS measurements are concerned, the primary target simply did not receive a fiber.

So, judicious use of `OVERRIDE` is my preferred solution. In fact, you could set `OVERRIDE` for all of your targets in SV, if you'd like.

To your second point. I believe that Kyle and Sarah (and maybe Jaime, too) are directly running fiberassign on target files, without first running them through MTL. They are sometimes tweaking quantities such as priorities and numbers of observations by hand to test how fiber assignment is affected. You'd have to reach out to them for details, but, to help facilitate fiberassign for SV, I was asked to put "initial" versions of `PRIORITY`, `NUMOBS` and `OBSCONDITIONS` directly in the targeting files.

apcooper:
If `OVERRIDE=True` is set for every secondary catalog, does that mean that the secondary catalogs will also not be matched against each other? Do we then end up with one source per catalog for any duplicates?

We have quite a lot of sources that appear in multiple input catalogs, so (optimistically) it would be nice to have the option of standardizing all the secondaries against the primary catalog but in the end creating a (single) new target rather than updating the primary.

On the other hand, maybe just restricting `OVERRIDE=TRUE` to the `VERYBRIGHT` target classes in SV (so they can be observed outside MWS) will not create too much of a mess, so there is no need for a complicated fix. For now let's go with that solution.

apcooper:
**However**, just remembered (as a result of checking the output) that this affects `PRIORITY_INIT` as well, so the problem is more complicated. We really want to be able to bump up the priority of primary targets that are matched to secondaries -- that's kind of the point of having all these secondary classes in our SV sample.

For example, we want stars flagged as narrowband-selected giants to be observed at high priority, but a good fraction of them will match MWS-MAIN stars. 

This would be solved by `OVERRIDE` but at the cost of an awful lot of duplicated targets among the secondaries, plus added confusion later when it comes to figuring out which primaries actually got  observed as higher-priority secondaries.

I wonder if a simpler solution is to have a flag set on the primary class that allows it to be overridden by matched secondaries? So if all nonzero primary bits (MWS, BGS, DESI) for a matched target have this flag `TRUE`, then it gets the secondary `PRIORITY` (if any matched secondaries are higher priority) and `OBSCONDITIONS` are `OR`'d.

apcooper:
Less general but maybe easier is to hack this in for MWS only by allowing priority and obsconditions to be bumped by secondaries only if `DESI_TARGET == MWS_ANY` (although I'd want to include standards as well, but same idea).

apcooper:
On the same subject but not directly related, I wondered if the updating that happens in `targets.calc_priority` is consistent with the principle that secondary bits don't affect the priority of primaries? I can't see anything that protects against picking up a different `MORE_ZGOOD` etc. from secondary bits in the lines below.

https://github.com/desihub/desitarget/blob/1436ec373b2b24244631f909e8a7e6c6a6df0e5c/py/desitarget/targets.py#L521-L531

geordie666:
@apcooper: Just to keep you posted on this. I've reached out to @sbailey to get his opinion on whether we should allow primary targets to have their priorities (and `NUMOBS`, `OBSCONDITIONS`, etc.) changed by secondaries or not.

If @sbailey agrees with me that this is potentially dangerous for LSS measurements, then we'll change this clause you noted:

https://github.com/desihub/desitarget/blob/1436ec373b2b24244631f909e8a7e6c6a6df0e5c/py/desitarget/targets.py#L521-L531

and throughout, to only allow priorities to be bumped (as you suggested) if `DESI_TARGET == MWS_ANY`. Otherwise, we'll take up your other option and allow all higher-priority secondaries to substitute their `NUMOBS`/`PRIORITY`/`OBSCONDITIONS` for those of a corresponding primary.

I'm still really nervous about this second option of letting secondaries overwrite primaries, as I think we could end up shunting dark-time targets into bright-time conditions. But, I've stated that sufficiently often now that I decided it was time to punt to @sbailey.

Sorry this is so complicated!

geordie666:
@apcooper: I've yet to hear from @sbailey (understandably; it's a busy time since DESI went on-sky) so I'm going to make a decision myself.

Can you work to implement your idea to only allow priorities to be bumped if `DESI_TARGET == MWS_ANY`? We can always revisit that choice for the Main Survey, but for SV I think that's the way to make progress. The philosophy would be that none of the large-scale structure samples could be affected.

I think you'll need two hacks to the code. One is here (for the initial priorities):

https://github.com/desihub/desitarget/blob/17060bcdf7177ba2cd871a51897b932f04551187/py/desitarget/secondary.py#L497-L503

where, for any matching primary-secondary targets, you'll have to update the qualities of the primary where the primary has `DESI_TARGET == desi_mask.MWS_ANY`. I think that you'll also what to update `NUMOBS_INIT` and `OBSCONDITIONS` to record the values of the highest-priority target for those quantities.

The second one is in MTL (as you were already discussing):

https://github.com/desihub/desitarget/blob/1436ec373b2b24244631f909e8a7e6c6a6df0e5c/py/desitarget/targets.py#L521-L531

where you'd want to update the `if scnd_target in targets.dtype.names:` to also only affect/update sources for which `DESI_TARGET == desi_mask.MWS_ANY`.

Hopefully that sounds agreeable as a solution?

py/desitarget/secondary.py
```diff
@@ -494,6 +494,38 @@ def match_secondary(primtargs, scxdir, scndout, sep=1.,
...
+        # APC only the following bits set, in any combination.
+        update_from_scnd_bits = desi_mask['MWS_ANY'] & desi_mask['STD_BRIGHT'] & desi_mask['STD_FAINT'] & desi_mask['STD_WD']
+        scnd_update &= ((targs[desicols[0]] & ~update_from_scnd_bits) == 0)
+        log.info('Setting new priority, numobs and obsconditions from secondary for {} matched primaries'.format(scnd_update.sum()))
+
+        # APC Primary and secondary obsconditions are or'd
+        scnd_obscon = set_obsconditions(targs[scnd_update], scnd=True)
+        targs['OBSCONDITIONS'][scnd_update] &= scnd_obscon
+
+        # APC bit of a hack here
```
apcooper: Wasn't sure if this test was necessary, or if this was the best way to do it. At least seems harmless.

py/desitarget/secondary.py
```diff
@@ -494,6 +494,38 @@ def match_secondary(primtargs, scxdir, scndout, sep=1.,
...
+        # APC Check for _BRIGHT, _DARK split in column names
```
apcooper: Wasn't sure if this test was necessary, or if this was the best way to do it. At least seems harmless.

py/desitarget/targets.py
```diff
@@ -520,11 +520,22 @@ def calc_priority(targets, zcat, obscon):
 
         # ADM Secondary targets.
         if scnd_target in targets.dtype.names:
+            # APC Secondary target bits only drive updates to targets with specific DESI_TARGET bits
+            # APC See https://github.com/desihub/desitarget/pull/530
+            scnd_update = (targets[desi_target] & desi_mask['SCND_ANY']) != 0
```
apcooper: Line 525 not strictly necessary given that individual scnd_mask names are checked below, but seems harmless way to create `scnd_update` vs. `np.repeat(False,len(targets))`.

apcooper:
Thanks @geordie666, how does the above look to you?

geordie666:
@apcooper: This looks good to me. If you think you have all of the functionality that you need, do you want to start moving a few MWS target classes to:
`
/project/projectdirs/desi/target/secondary/sv1/indata/
/project/projectdirs/desi/target/secondary/sv1/docs/
`
So I can test this branch in earnest?

apcooper:
@geordie666: I've copied those files now.

geordie666:
@apcooper: I've been stress-testing this PR over the last week. I have 4 issues, but I think I can handle two of them (after we merge this) and I'd like to leave the other 2 for you to deal with before we merge.

First, the issues for me to handle after we merge:

* We now use `desitarget.io.find_target_files()` for standard I/O of targeting files, which has broken the default output locations for secondary targets. I need to update `desitarget.io.write_secondary()` to be consistent with this new formalism.
* This line
https://github.com/desihub/desitarget/blob/5c58a1208ee5d8e57a3eff876e3c33a81ae6d15e/py/desitarget/io.py#L695
needs to be conscious of the fact that we can pass both SV targets and Main Survey targets. We need a check on [`desitarget.targets.main_cmx_or_sv()`](https://github.com/desihub/desitarget/blob/5c58a1208ee5d8e57a3eff876e3c33a81ae6d15e/py/desitarget/targets.py#L183) instead of just importing the main survey masks. This appears to be a bug I introduced a while ago, but I only noticed it when you created new secondary files for testing SV.

Now, the issues that I'd like you to handle, or at least comment on:

* Your secondary files at NERSC have the wrong data model for some columns. In particular, `REF_EPOCH` should be `f4` but instead it is `S6` in your files. This is currently triggering a (non-malicious) `type` bug. Could you please add a check that asserts the data model for the mandatory columns? The mandatory columns would be those listed in `desitarget.secondary.indatamodel`:

https://github.com/desihub/desitarget/blob/5c58a1208ee5d8e57a3eff876e3c33a81ae6d15e/py/desitarget/secondary.py#L70-L73

* Currently, if we write out both `dark` and `bright` secondary targets (which is the default for `bin/select_secondary`) then the output files don't contain many of your `sv1_scnd_mask` `MWS_` targets. This is because those targets don't have `DARK` or `BRIGHT` specified as one of their `obsconditions`. I suspect that while developing this branch you've just been passing `--writeall` to `select_secondary` and ignoring the `dark/` `bright/` outputs completely. 

So...what do you think about just forcing the `--writeall` condition for secondary targets and never writing out the `dark` and `bright` individual directories? This would greatly simplify the outputs for secondary targets. I think that, in the cases we've been discussing, this is a safe choice, because other than your MWS targets, which you've already resolved, we can never have two secondary targets that match each other, one with `obsconditions` including `DARK` and one with obsconditions including `BRIGHT`? If you agree that this is the correct choice, then I think you just need to replace

https://github.com/desihub/desitarget/blob/44c0445f44f66f52b74112dbc639b2913fc573e6/bin/select_secondary#L59-L68

with:
```
ntargs, outfile = io.write_secondary(ns.dest, scx, primhdr=dict(hdr),
                                         scxdir=scxdir, obscon=None)
```
If you can handle these last two points, then let's merge-to-make-progress and I'll handle formatting issues next week.

apcooper:
@geordie666: regarding your first issue,

> * Your secondary files at NERSC have the wrong data model for some columns. In particular, `REF_EPOCH` should be `f4` but instead it is `S6` in your files. This is currently triggering a (non-malicious) `type` bug. Could you please add a check that asserts the data model for the mandatory columns? The mandatory columns would be those listed in `desitarget.secondary.indatamodel`:

I've added code to ensure the files are consistent with `desitarget.secondary.indatamodel` and written new files. Please let me know if I missed any!

(In the course of doing this it dawned on me that the notebooks for each secondary class that I'm putting under `docs` are not self-contained. Each notebook imports some common python code that lives under `/project/projectdirs/desi/external/mws/py/mwsecondary/`, via a `sys.path.append`. This is a bit ugly, maybe we can come up with a more sustainable solution.)

apcooper:
@geordie666: On the second issue:

> * Currently, if we write out both `dark` and `bright` secondary targets (which is the default for `bin/select_secondary`) then the output files don't contain many of your `sv1_scnd_mask` `MWS_` targets. This is because those targets don't have `DARK` or `BRIGHT` specified as one of their `obsconditions`. 

I guess those are the targets with only `TWILIGHT|POOR` etc. obsconditions, intended for 'backup' program SV. All my comments below are assuming that's the case.

> I suspect that while developing this branch you've just been passing `--writeall` to `select_secondary` and ignoring the `dark/` `bright/` outputs completely.

Perhaps confusingly, I have been passing `--writeall` but using the split files (just those under `bright`) for all recent tests, because they still get written with `--writeall`. 

Since we haven't had the regular backup program targets in the mix until very recently, I've just been ignoring the issue of backup-only special targets, assuming we'd come to that later.

> So...what do you think about just forcing the `--writeall` condition for secondary targets and never writing out the `dark` and `bright` individual directories? This would greatly simplify the outputs for secondary targets. I think that, in the cases we've been discussing, this is a safe choice, 

I'm not sure what the motivation was for having separate dark and bright files in the first place (sorry if you've told me and I've forgotten).

If it's easier to go back to a single file for secondaries rather than include backup-only secondaries in the separate-files scheme, then I'll make that change. However,

> because other than your MWS targets, which you've already resolved, we can never have two secondary targets that match each other, one with `obsconditions` including `DARK` and one with obsconditions including `BRIGHT`?

I guess that could still happen later, if e.g. another group selects some secondary QSOs that match some of our BHBs or WDs?

Also, although I think our current selections avoid this, I guess it's possible to have stars selected as both a regular secondary SV target and a backup-program SV target.

geordie666:
@apcooper: On the first point...I was hoping you'd incorporate a data-model-check directly into desitarget. But, no worries, let's merge this PR as soon as we can and I'll add that check myself. I'm not particularly worried about your use of common code in `/project/projectdirs/desi/external/mws/py/mwsecondary/` as the purpose of the documentation notebooks/files is to give a broad description of how the secondary samples are created. Actual code corresponding to the selection algorithm is fantastic, of course, but I'm more worried that the selections contain links to non-transparent files like `input_filename = 'apogee_xm_gaia_4DESI_v0.fits'`. What we really need is documentation of _how the file `apogee_xm_gaia_4DESI_v0.fits` was created_. Think of this at the level of a paragraph that you'd write in a science paper about each MWS secondary selection (which will hopefully save time when writing papers). I see you have placeholders for such descriptions. I'd say those descriptions are very important to fill-in and vet well before SV.

On the second point...I've been going back-and-forth on this, but I think you're correct that we should retain bright/dark sub-directories for secondary targets (in case of QSO/star mix-ups etc.). So, my suggested way forward is that all of the secondary target classes that have just `POOR|TWILIGHT_X' `obsconditions, e.g.: `{obsconditions: POOR|TWILIGHT12|TWILIGHT18, filename: 'MWS_WD_VERYBRIGHT_SV'}]` should be promoted to also allow `BRIGHT`. This wouldn't preclude such targets from being observed in `POOR` or `TWILIGHT` conditions, but it will guarantee that they appear in the `BRIGHT` file (and they could then also be observed in `BRIGHT` conditions if such conditions became available...hopefully that's not an issue?).

There may be some general misunderstanding about the "back-up/twilight" program. This needs to be something very simple that we can use in a "blanket" sense over the entire sky. It's currently as defined by Boris for for the [Main Survey](https://desi.lbl.gov/trac/wiki/TargetSelectionWG/TargetSelection#PrimarysampleBACKUP_BRIGHT) and [SV](https://desi.lbl.gov/trac/wiki/TargetSelectionWG/SurveyValidation#BrightTimeBackupStellarSurvey) and I don't think we should complicate it further.

So, in any case, if you're OK with the solution of promoting your "VERYBRIGHT" classes to also allow an `obsconditions` of `BRIGHT`, then go ahead and add that and I think we can then merge this PR.

apcooper:
@geordie666: Sorry to prolong this even further, but regarding the backup program and our twilight SV targets:

* We don't want the VERYBRIGHT_SV targets observed in BRIGHT conditions because many of them are way too bright to share an exposure with regular MWS targets. 

* When I say 'backup program' I should clarify: what we *really* want is to have our stellar SV tiles observed under the conditions of the backup program, to get the brighter stars in those fields. Relatively more of the very bright targets have pre-existing spectra, which serve as a check on our pipeline results and a test of performance under the conditions of the backup program.

* Where we observe stellar SV tiles in poor conditions to get the VERYBIRGHT_SV targets, I think we would want to fill the fibres with regular backup program targets. 

It would be a bonus to have the VERYBRIGHT_SV targets on *any* SV backup program tile where they might appear (i.e. any tile with POOR|TWILIGHT_X conditions) rather than just the Stellar SV tiles. That seems like the simpler route to me -- otherwise how do we stop them appearing on such tiles? There are not very many of those targets and for SV purposes they're valuable.

(The other point about documentation is noted, I'll follow up on that with the creators of each file.)

geordie666:
@apcooper: Trying not to modify the code too much to deal with special cases, how about this suggestion:

We change to running `select_sv_targets` with the `--writeall` option by default. Then, when tiles are being constructed for SV, if they are `POOR|TWILIGHT_X` tiles, then the secondary targets for those programs are taken from the "all" output file rather than from the `dark` or `bright` output file?

This requires a little wisdom downstream, as whoever is creating `POOR|TWILIGHT` tiles has to know which secondary targets to consider when merging targets to make tiles. But that's a detail, I think. Would that work? If you think it will, then either change the `--writeall` default yourself, or just indicate to me that this is a solution and we'll merge this PR and I'll add `--writeall` as the default in a subsequent branch.

apcooper:
@geordie666 : That solution sounds fine to me, if the data systems side can tolerate that way of working. 

If it's ok with you, let's merge this now and then we can patch up the outstanding items in another branch. I'm happy to help with that if you don't have time.

geordie666:
@apcooper: I agree. Merge this whenever you have the chance. I'll start working on updating the secondary targets for data model changes in a separate branch.

Issue #536, opened 2019-09-16
using zcat information to update numobs in mtl


forero:
The current implementation of `make_mtl` discards the zcat information when computing the `NUMOBS` that will be requested to fiberassign

https://github.com/desihub/desitarget/blob/master/py/desitarget/mtl.py#L129
```python
# ADM use passed value of NUMOBS_INIT instead of calling the memory-heavy calc_numobs.
    # ztargets['NUMOBS_MORE'] = np.maximum(0, calc_numobs(ztargets) - ztargets['NUMOBS'])
    ztargets['NUMOBS_MORE'] = np.maximum(0, targets_zmatcher['NUMOBS_INIT'] - ztargets['NUMOBS'])
```

We need to go back to using `calc_numobs` in the same way that we use `calc_priority`, i.e. to include spectroscopic information to update the MTL.

forero:
The main case that motivates this comment is the following. Imagine starting with a target that is labeled as a `QSO` and requests `NUMOBS=4` to start with (4 is the goal for lya qsos). After the first observation we confirm that its a low redshift quasar. In this case we need to request`NUMOBS=0` and not `NUMOBS=3` as it is the current case.

geordie666:
@forero: If you look at versions of `calc_numobs` prior to when I deprecated it, you'll see that it was never being used to set `NUMOBS=0` for LyA quasars:

https://github.com/desihub/desitarget/blob/0.31.0/py/desitarget/targets.py#L629-L630

That may have been the _intention_, but it has never been the _implementation_.

LyA quasars are set to `NUMOBS=0` based on their priorities dropping to what is appropriate for `MORE_ZGOOD`.

For example:

```
from astropy.table import Table, join
from desitarget.mtl import make_mtl
from desitarget.targetmask import desi_mask as Mx
from desitarget.targets import initial_priority_numobs
import numpy as np
# ADM set up a couple of QSO targets.
targets = Table()
types = np.array(['QSO', 'QSO'])
priorities = [Mx[t].priorities['UNOBS'] for t in types]
post_prio = [Mx[t].priorities['MORE_ZGOOD'] for t in types]
targets['DESI_TARGET'] = [Mx[t].mask for t in types]
targets['BGS_TARGET'] = np.zeros(len(types), dtype=np.int64)
targets['MWS_TARGET'] = np.zeros(len(types), dtype=np.int64)
n = len(targets)
targets['TARGETID'] = list(range(n))
[...]

PR #537, opened 2019-09-17
handles tracer QSO inside mtl

forero:
This solves the main problem around #536.

### DIFF
--- a/py/desitarget/mtl.py
+++ b/py/desitarget/mtl.py
@@ -138,7 +138,17 @@ def make_mtl(targets, obscon, zcat=None, trim=False, scnd=None):
         if (obsconditions.mask(obscon) & obsconditions.mask("BRIGHT")) != 0:
             ii = targets_zmatcher[desi_target] & desi_mask.BGS_ANY > 0
             ztargets['NUMOBS_MORE'][ii] = 1
+            
+        # If the object is confirmed to be a tracer QSO, then don't request more observations
+        if (obsconditions.mask(obscon) & obsconditions.mask("DARK|GRAY")) != 0:
+            if zcat is not None:
+                ii = ztargets['SPECTYPE']=='QSO'
+                ii &= (ztargets['ZWARN'] == 0 ) 
+                ii &= (ztargets['Z']<2.15)
+                ii &= (ztargets['NUMOBS']>0)
+                ztargets['NUMOBS_MORE'][ii] = 0
 
+            
     # ADM assign priorities, note that only things in the zcat can have changed priorities.
     # ADM anything else will be assigned PRIORITY_INIT, below.
     priority = calc_priority(targets_zmatcher, ztargets, obscon)
--- a/py/desitarget/mtl.py
+++ b/py/desitarget/mtl.py
@@ -144,7 +144,7 @@ def make_mtl(targets, obscon, zcat=None, trim=False, scnd=None):
             if zcat is not None:
                 ii = ztargets['SPECTYPE']=='QSO'
                 ii &= (ztargets['ZWARN'] == 0 ) 
-                ii &= (ztargets['Z']<2.15)
+                ii &= (ztargets['Z']<2.1)
                 ii &= (ztargets['NUMOBS']>0)
                 ztargets['NUMOBS_MORE'][ii] = 0
             
--- a/py/desitarget/mtl.py
+++ b/py/desitarget/mtl.py
@@ -140,7 +140,7 @@ def make_mtl(targets, obscon, zcat=None, trim=False, scnd=None):
             ztargets['NUMOBS_MORE'][ii] = 1
             
         # If the object is confirmed to be a tracer QSO, then don't request more observations
-        if (obsconditions.mask(obscon) & obsconditions.mask("DARK|GRAY")) != 0:
+        if (obsconditions.mask(obscon) & obsconditions.mask("DARK")) != 0:
             if zcat is not None:
                 ii = ztargets['SPECTYPE']=='QSO'
                 ii &= (ztargets['ZWARN'] == 0 ) 
--- a/py/desitarget/mtl.py
+++ b/py/desitarget/mtl.py
@@ -133,7 +133,7 @@ def make_mtl(targets, obscon, zcat=None, trim=False, scnd=None):
     # ADM (and only once) every time during the BRIGHT survey, regardless
     # ADM of how often they've previously been observed. I've turned this
     # ADM off for commissioning. Not sure if we'll keep it in general.
-    if survey != 'cmx':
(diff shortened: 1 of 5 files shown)


geordie666:
Thanks @forero, I'm glad this was a relatively simple fix.

Could you add a unit test in `test/test_mtl.py` that is designed to fail with the bug you were finding before this fix?

geordie666:
Oh, and also add a line in `doc/changes.rst` related to this PR?

forero:
Sure!

sbailey:
Fix looks good; I agree with @geordie666 test requests.  It looks like you'll also need to add a "SPECTYPE" column to the zcat table created in the test_priorities and test_mtl tests, since SPECTYPE is now used in mtl.

forero:
`SPECTYPE` is in, but the tests fail under `SV1` where tracer and lya qso are expected to have the same priority after being confirmed by zcat. That doesn't seem valid for the main survey. Should I then simply add `if survey != 'sv1'` inside the `make_mtl` logic?

geordie666:
I think you could just replace `if survey != 'cmx':` with `if survey == 'main':`. That might be easiest.

geordie666:
@forero: I'm happy to merge this but not until unit tests pass...let me know if you can't work out the issue and I'll take over this branch and try to get unit tests working.

forero:
@geordie666 I am busy today. Tomorrow I will have time to fix it for you to merge.

py/desitarget/mtl.py
```diff
@@ -138,7 +138,16 @@ def make_mtl(targets, obscon, zcat=None, trim=False, scnd=None):
...
-
+    if survey == 'main':
+        # If the object is confirmed to be a tracer QSO, then don't request more observations
+        if (obsconditions.mask(obscon) & obsconditions.mask("DARK")) != 0:
+            if zcat is not None:
+                ii = ztargets['SPECTYPE']=='QSO'
+                ii &= (ztargets['ZWARN'] == 0 ) 
+                ii &= (ztargets['Z']<2.15)
+                ii &= (ztargets['NUMOBS']>0)
+                ztargets['NUMOBS_MORE'][ii] = 0
```
sbailey: @geordie666 (or others): what should we do for spectype=QSO, z<2.15, and zwarn>0?  As implemented in this PR that case would get more observations but I'm not sure that is what we want.  I thought the default policy was "only continue observing if definitively confirmed to be QSO z>2.15" rather than "only stop observing if definitively confirmed to be QSO z<2.15 (i.e. keep observing ambiguous cases too)"

I know the LyA group (@andreufont in particular) have floated the idea of getting a second observation for ambiguous cases, but stopping if a second observation doesn't return a definitive answer.  Implementing that might need a deeper code update.

If we know what the current default plan is, let's implement that here.  If we still need discussion to refine that plan, we could go ahead and merge this PR as a step in the right direction (since this prevents confirmed z<2.15 QSOs from being re-observed) and then make additional updates in future PRs.

andreufont: Hi @sbailey, we have been meaning to look at this closely for years, but none has managed to provide a definite answer.

Some issues we have faced so far:
 - When running tests of the classifiers on DESI mocks, the performance is too good, at least when compared to BOSS/eBOSS data. We believe this is due to the quasar templates in desisim do not have enough diversity.
 - We haven't been able to integrate the classifiers to desihub, because they depend on Scikit Learn or TensorFlow. For the rest of the discussion, we could assume that we will only use Redrock (since this is already integrated).
 - We still haven't figure out how to integrate the result of the different classifiers to the main pipeline (fiber re-assignement). I believe @geordie666 mentioned that he could be able to help here.
 - We still haven't been able to run our survey simulations, including fiber re-assignement, to be able to test the software I mention above. Julian Bautista tried for a while, but he got stuck early on in some issues with mini-test.
 
Given the difficulties described above, I'd really like to have a baseline that does not assume that we will solve all the issues above in the next couple of months (we will try tough!). It would be great if we could have the following baseline:
 - stop observing if definitively confirmed that it is NOT a QSO z > 2.1

I'd like to test in survey simulations how many fibers we would need to do that, although again, the tests will be as good as the quasar templates...

sbailey: Thanks for the comments @andreufont .  In main operations we will (hopefully) have multiple codes checking and more human oversight, but the same policy questions will still apply (e.g. there will still be some ambiguous maybe-QSOs that we'll have to decide what to do).  I agree that the default pipeline+redrock should at the policy level do the right thing, such that other codes improve the correctness / purity of the decisions, but don't fundamentally alter the policies.

> * stop observing if definitively confirmed that it is NOT a QSO z > 2.1

That is what is implemented here (almost, z 2.15 instead of 2.1), with the consequence that
 
* continue observing if definitively confirmed that it *is* QSO z>2.1 (good)
* continue observing if the redshift answer is ambiguous (maybe that's what we want, maybe not, especially if it is still ambiguous after the second observation).

@andreufont can you confirm what you think we should do in the case of an ambiguous redshift?  (but no promises that we'll implement that right now depending on how complex it is...)

geordie666:
@sbailey: I think we should merge this as-is, for now. The current iteration of the code won't be exactly correct, but @forero has made useful progress.

I appreciate that we will need tweaks to the logic regarding how we partition QSOs into 1-versus-4 observations, but I don't think the Lyman-alpha WG has determined the exact algorithm to do that, yet.

@andreufont: We really need to converge on that algorithm soon. As I've indicated to you, I'm available and willing to help the Lyman-alpha group plug things into the code. I'd hope that we can determine the correct decision tree fairly soon...certainly not long after we get the first SV-like observations of quasars during commissioning.

geordie666:
@sbailey: I think we should merge this as-is, for now. The current iteration of the code won't be exactly correct, but @forero has made useful progress.

I appreciate that we will need tweaks to the logic regarding how we partition QSOs into 1-versus-4 observations, but I don't think the Lyman-alpha WG has determined the exact algorithm to do that, yet.

@andreufont: We really need to converge on that algorithm soon. As I've indicated to you, I'm available and willing to help the Lyman-alpha group plug things into the code. I'd hope that we can determine the correct decision tree fairly soon...certainly not long after we get the first SV-like observations of quasars during commissioning.

sbailey:
Thanks for comments @geordie666 .  I see that @forero just made a test update, but once tests pass again it's fine to merge this.  We can deal with updates to the QSO followup logic in other PRs.

forero:
I have added a new unit test that fails with the version on master but passes with the changes in `make_mtl`.

forero:
Do you have any other requests?

geordie666:
@forero: No more requests for now. I'll merge in about an hour (and once tests pass).

PR #543, opened 2019-09-24
update select_mock_targets to match latest data-based targets catalogs

moustakas:
A variety of updates to `select_targets` broke integration tests because `select_mock_targets` had fallen too far behind.


py/desitarget/targets.py
```diff
@@ -598,7 +598,8 @@ def resolve(targets):
 
 
 def finalize(targets, desi_target, bgs_target, mws_target,
-             sky=0, survey='main', darkbright=False):
+             sky=0, survey='main', darkbright=False,
+             targetid=None):
     """Return new targets array with added/renamed columns
```
moustakas: @geordie666, here...

py/desitarget/targets.py
```diff
@@ -655,10 +658,13 @@ def finalize(targets, desi_target, bgs_target, mws_target,
...
-                               brickid=targets['BRICKID'],
-                               release=targets['RELEASE'],
-                               sky=sky)
+    # allow TARGETID to be passed as an input (specifically for the mocks).
+    if targetid is not None:
+        targetid = encode_targetid(objid=targets['BRICK_OBJID'],
+                                   brickid=targets['BRICKID'],
+                                   release=targets['RELEASE'],
+                                   sky=sky)
+    assert ntargets == len(targetid)
```
moustakas: ...and here are the only places (so far) I've had to touch the nominal (non-mock) target selection code.  These changes should hopefully be non-controversial, but I wanted to explicitly alert you.

geordie666: Shouldn't this be
```
if targetid is None:
```
rather than
```
if targetid is not None:
```
?
In other words, the default behavior is for `targetid` to be `None` but currently, if that default is the case, then the clause:

```
 if targetid is not None:
        targetid = encode_targetid(objid=targets['BRICK_OBJID'],
```
will be skipped and a TARGETID will never be calculated?

sbailey:
I'm not sure if this PR was laying the groundwork, or intended to fix this, but I'm still getting this error from `join_mock_targets`:
```
...
INFO:build.py:1289:_merge_file_tables: Writing /global/cscratch1/sd/sjbailey/minitest-19.9/targets/truth.fits TRUTH_STAR
INFO:build.py:1289:_merge_file_tables: Writing /global/cscratch1/sd/sjbailey/minitest-19.9/targets/truth.fits TRUTH_WD
INFO:build.py:1403:join_targets_truth: Generating merged target list /global/cscratch1/sd/sjbailey/minitest-19.9/targets/mtl.fits
  File "/global/cscratch1/sd/sjbailey/desi/code/desitarget/bin/join_mock_targets", line 33, in <module>
    join_targets_truth(args.mockdir, outdir=args.outdir, overwrite=args.overwrite, comm=comm)
  File "/global/cscratch1/sd/sjbailey/desi/code/desitarget/py/desitarget/mock/build.py", line 1405, in join_targets_truth
    mtl = mtl.make_mtl(targets)
TypeError: make_mtl() missing 1 required positional argument: 'obscon'
```

moustakas:
WIP

moustakas:
@sbailey 

For the record: using `desitarget.io.write_targets` resulted in the obscure error
```
*** ValueError: unsupported type 'U8'
```
which turned out to come from `fitsio.write`.  

On my laptop I was running `fitsio/0.9.12` and upgrading to `fitsio/1.0.5` fixed the issue.

At NERSC it looks like we're using `fitsio/0.9.11` and since @geordie666 hasn't seen this issue we must be OK (and I'll be sure to test, of course).

But it may be something to keep an eye on.

moustakas:
OK, this branch now writes out `dark` and `bright` targets (and truth) catalogs for mocks identically how we write out the data-based catalogs.  (Unfortunately, @sbailey, this means you're going to have to rerun `select_mock_targets` for the `19.9` software release (but only I'm done testing!).

Next piece is to update the MTL-building and "joining" of the various catalogs.

moustakas:
I'll continue to test the code, but this should be ready for review.  

@geordie666 I did have to touch some additional code (e.g., `io.write_targets`) that was previously only called when building the "real data" target catalogs, but my changes (to handle the mocks) should hopefully be not too controversial.

moustakas:
@geordie666 One minor issue I ran into in this PR is the line here--
  https://github.com/desihub/desitarget/blob/master/py/desitarget/io.py#L393-L394

In the case of the mocks I set `RELEASE` equal to `9999`, and so `drstring` here ends up being `dr9`!  

Should I use a different `RELEASE` number?  And/or should we perhaps allow strings like `dr99` (or, for that matter, `dr9.1`)?  Other suggestions are welcome.

py/desitarget/mock/build.py
```diff
@@ -1391,7 +1394,7 @@ def join_targets_truth(mockdir, outdir=None, overwrite=False, comm=None):
         todo = comm.bcast(todo, root=0)
 
     if todo['sky']:
-        _merge_file_tables(mockdir+'/*/*/sky-*.fits', 'SKY_TARGETS',
+        _merge_file_tables(mockdir+'/*/*/sky-*.fits', 'SKY',
                            outfile=outdir+'/sky.fits', comm=comm,
                            overwrite=overwrite)
                            #addcols=dict(OBSCONDITIONS=obsmask.mask('DARK|GRAY|BRIGHT')))
```
moustakas: @sbailey FYI: I called this extension `SKY_TARGETS` to match the name of the extension written out by `io.write_skies` for "real" data.

sbailey: Thanks.  I incorrectly changed that when my minitest notebook was accidentally picking up desitarget master instead of this branch, and files were created with the old SKY name and then I was using this branch to try to merge them.  I reverted back to SKY_TARGETS.

I've made it past mock targets with this branch and am now hitting some fiberassign bookkeeping issues.  All my changes are pushed to github, but I'd like to resolve whether the fiberassign problems are on the fiberassign side or on the mock inputs side or user error before merging this.  I hope to get back to this later today after juggling some other flaming cats.

geordie666:
As regards your `RELEASE` number issue: This is just what is written to the output file header as `PHOTCAT`. I don't think it's critical if `dr9` is written to file for the mocks, but another option would be to have `RELEASE` be equal to 0 for the mocks.

Hopefully this wouldn't break other things, but with `RELEASE=0` and a correct call to `desitarget.targets.encode_targetid(mock=1)`, then `TARGETID`s should behave correctly and (I think) that `PHOTCAT` would be written as `dr0`.

I don't feel strongly about this, though. If the mocks are, ultimately, attempting to recreate the state-of-play for SV and the Main Survey, then `dr9` is presumably correct!

BTW, I'm fine with this PR being merged once @sbailey is happy that it passes integration tests. Other than the one `is None/is not None` that was already fixed, I couldn't find any issues.

sbailey:
We're getting close but please don't merge yet.  We need to make some updates to `desitarget.mock.io.findfile` to support the targets and truth files being grouped under dark/bright but the sky file not.  I'm chasing this and making updates in the context of downstream desisim newexp-mock.

sbailey:
I've made it far enough downstream in the minitest integration notebook that I think this contains what we need.  Thanks.  I'll merge now and open a new ticket for anything else that I trip over.  A companion desisim PR is forthcoming...

PR #544, opened 2019-09-28
Fixes missing discard of duplicate secondary targets

apcooper:
@geordie666 , I think this fixes the issue I raised by email: multiple rows in the output of `select_secondary` for (a) secondary targets that appear in more than one input catalog or (b) primary targets matched to more than one secondary.

Case (a) is fixed here:
https://github.com/desihub/desitarget/blob/203e392cf427c57391f1b6eabaec5808a7684bc3/py/desitarget/secondary.py#L648-L654

Case (b) is fixed here:
https://github.com/desihub/desitarget/blob/203e392cf427c57391f1b6eabaec5808a7684bc3/py/desitarget/secondary.py#L358

which was previously
https://github.com/desihub/desitarget/blob/203e392cf427c57391f1b6eabaec5808a7684bc3/py/desitarget/secondary.py#L357

**Question**
https://github.com/desihub/desitarget/blob/203e392cf427c57391f1b6eabaec5808a7684bc3/py/desitarget/secondary.py#L547-L554
I didn't see the `PRIORITY_INIT=-1` filtering implemented anywhere and it's not how I've handled case (a) in this PR. Wondering if I missed something about the intention here?

### DIFF
--- a/bin/select_secondary
+++ b/bin/select_secondary
@@ -27,10 +27,9 @@ ap.add_argument("-s", "--separation", type=float, default=1.,
 ap.add_argument("--scnddir",
                 help="Base directory of secondary target files (e.g. '/project/projectdirs/desi/target/secondary' at NERSC). "+
                 "Defaults to SCND_DIR environment variable.")
-ap.add_argument("--darkbright",
+ap.add_argument("--writeall",
                 action='store_true',
-                help=" If sent, then split `NUMOBS_INIT` and `PRIORITY_INIT` into `NUMOBS_INIT_DARK`, `NUMOBS_INIT_BRIGHT`, `PRIORITY_INIT_DARK`"+
-                "and `PRIORITY_INIT_BRIGHT` and calculate values appropriate to 'BRIGHT' and 'DARK|GRAY' observing conditions.")
+                help="Default behavior is to split targets by bright/dark-time surveys. Set this to ALSO write a file of ALL targets")
 ns = ap.parse_args()
 
 # ADM Sanity check that priminfodir exists.
@@ -50,7 +49,7 @@ scxdir = _get_scxdir(ns.scnddir)
 if surv != 'main':
     scxdir = os.path.join(scxdir, surv)
 
-scx = select_secondary(ns.priminfodir, sep=ns.separation, scxdir=scxdir, darkbright=ns.darkbright)
+scx = select_secondary(ns.priminfodir, sep=ns.separation, scxdir=scxdir, darkbright=(not ns.writeall))
 
 # ADM add the primary directory and matching radius to the header
 # ADM from the first primary file.
@@ -61,6 +60,8 @@ hdr['SEP'] = float(ns.separation)
 # ADM and dark-time targets written separately.
 obscons = ["BRIGHT", "DARK"]
 
+if ns.writeall:
+    obscons.append(None)
 for obscon in obscons:
     # ADM the dict() here is to make hdr immutable.
     ntargs, outfile = io.write_secondary(ns.dest, scx, primhdr=dict(hdr),
--- a/bin/select_secondary
+++ b/bin/select_secondary
@@ -27,7 +27,10 @@ ap.add_argument("-s", "--separation", type=float, default=1.,
 ap.add_argument("--scnddir",
                 help="Base directory of secondary target files (e.g. '/project/projectdirs/desi/target/secondary' at NERSC). "+
                 "Defaults to SCND_DIR environment variable.")
-
+ap.add_argument("--darkbright",
+                action='store_true',
+                help=" If sent, then split `NUMOBS_INIT` and `PRIORITY_INIT` into `NUMOBS_INIT_DARK`, `NUMOBS_INIT_BRIGHT`, `PRIORITY_INIT_DARK`"+
+                "and `PRIORITY_INIT_BRIGHT` and calculate values appropriate to 'BRIGHT' and 'DARK|GRAY' observing conditions.")
 ns = ap.parse_args()
 
 # ADM Sanity check that priminfodir exists.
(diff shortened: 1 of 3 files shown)


geordie666:
I'll try to look at this closely in the next day. But, the filtering on `PRIORITY_INIT` of -1 is done in the write-to-file, here:

https://github.com/desihub/desitarget/blob/master/py/desitarget/io.py#L608-L615

Are you perhaps running the functions from within Python rather than running the command-line executables (such as `select_secondary`)? If so, could you see if your original issue persists when running `select_secondary` instead of calling the `secondary.select_secondary()` function directly?

apcooper:
Ah, ok. Now I know where the filtering happens (I didn't check `io.write_secondary`) I'll just set `PRIORITY_INIT=-1` rather than filtering in `finalize_secondary`.

bin/select_secondary
```diff
@@ -27,7 +27,10 @@ ap.add_argument("-s", "--separation", type=float, default=1.,
 ap.add_argument("--scnddir",
                 help="Base directory of secondary target files (e.g. '/project/projectdirs/desi/target/secondary' at NERSC). "+
                 "Defaults to SCND_DIR environment variable.")
-
+ap.add_argument("--darkbright",
+                action='store_true',
```
geordie666: I think we want the default behaviour, here, to be to run with `darkbright=True`. In other words, we want the standard behavior to be to split everything into `bright-` and `dark-` file types. Can you change this so the argument is `--writeall` (like for the primary targets), and have the default be to split?

If you think `--writeall` is misleading, feel free to come up with a better term (but also change to that term in `select_targets` and `select_sv_targets` for the primary targets).

bin/select_secondary
```diff
@@ -47,7 +50,7 @@ scxdir = _get_scxdir(ns.scnddir)
 if surv != 'main':
     scxdir = os.path.join(scxdir, surv)
 
-scx = select_secondary(ns.priminfodir, sep=ns.separation, scxdir=scxdir, darkbright=True)
+scx = select_secondary(ns.priminfodir, sep=ns.separation, scxdir=scxdir, darkbright=ns.darkbright)
```
geordie666: With my requested default, the `darkbright` kwarg, here would be `darkbright=not(ns.writeall)` or some-such.

py/desitarget/secondary.py
```diff
@@ -346,26 +345,28 @@ def add_primary_info(scxtargs, priminfodir):
         primtargs = np.concatenate([primtargs, prim])
 
     # ADM make a unique look-up for the target sets.
-    scxbitnum = np.log2(scxtargs["SCND_TARGET"]).astype('int')
+    scxbitnum  = np.log2(scxtargs["SCND_TARGET"]).astype('int')
```
geordie666: In an number of places you introduce some extra spaces that don't conform to `PEP 8` code style. I'll probably catch these and fix them in a later PR, but if you want to fix them yourself you can see any issues in the [code style unit tests for your PR](https://travis-ci.org/desihub/desitarget/jobs/591940611#L973-L990).

py/desitarget/secondary.py
```diff
@@ -636,6 +641,15 @@ def finalize_secondary(scxtargs, scnd_mask, sep=1., darkbright=False):
     scxtargs = rfn.rename_fields(
         scxtargs, {'SCND_TARGET': prepend+'SCND_TARGET'}
     )
+
+    # APC Remove duplicate targetids from secondary-only targets
+    alldups = []
+    for _, dups in duplicates(scxtargs['TARGETID']):
+        dups = np.delete(dups, 0) # Retain the first
```
geordie666: I don't think we want to retain the first duplicate among matching secondary targets, I think we want to retain the duplicate with the _highest priority_. Is it possible to rewrite this to retain the highest priority secondary-only target?

bin/select_secondary
apcooper: OK, I was unsure what the default should be. Will fix this as suggested.

bin/select_secondary
apcooper: OK.

py/desitarget/secondary.py
apcooper: Bah, force of habit when parsing other people's code, sorry. Turns out my personal preference isn't in line (pun intended) with PEP8. Now I know, will fix.

py/desitarget/secondary.py
apcooper: Good catch, thanks -- will sort this out. Thanks for the review!

apcooper: Done, except I wasn't quite sure what was meant in the docstring by "the first encountered" . I wrote it to use the lowest index in the original table to break ties.

py/desitarget/secondary.py
```diff
@@ -647,8 +647,11 @@ def finalize_secondary(scxtargs, scnd_mask, sep=1., darkbright=False):
     # APC Remove duplicate targetids from secondary-only targets
     alldups = []
     for _, dups in duplicates(scxtargs['TARGETID']):
-        # Retain the duplicate with highest priority
-        dups = np.delete(dups, np.argmax(scxtargs['PRIORITY_INIT'][dups]))
+        # Retain the duplicate with highest priority, breaking ties
+        # on lowest index in list of targets (see docstring)
```
apcooper: Breaking ties on priority_init on the lowest index in scxtargs.

geordie666:
@apcooper: I think this looks good, now. If you agree I'll merge this PR.

Thanks for putting this effort into getting the secondary targets correct. It's important work and I'm swamped at the moment.

py/desitarget/secondary.py
geordie666: Oh, "the first encountered" here just means that if we find two secondaries with the same priority, we just keep the first one in the table/array/loop. In other words, we don't break ties in any clever way once we've broken ties to keep the highest-priority secondary. That's not critical though. The way we write everything back to the input files and generate consistent TARGETIDs should keep track of what was matched and how.

If you _don't_ think the code does this, just remove the sentence in the docstring. As I noted, it's not critical how we break ties other than on priority.

apcooper: OK, in that case I'll just use np.argmax directly, since that breaks ties on the lowest index in the list of duplicates. Done now.

apcooper:
OK, I pushed one last commit to simplify the tie-break part. Ready to merge now.

geordie666:
One last request for good practice:

Can you add a line or two to the [changes.rst](https://github.com/desihub/desitarget/blob/master/doc/changes.rst) file to describe this PR. Then I'll merge. Thanks!

apcooper:
Done, hope that's enough info in `changes.rst`.

Issue #541, opened 2019-09-24
Gaia catalog density cuts vs. l,b not just b


sbailey:
Rather than a simple cut on galactic latitude, Boris suggests the following cuts to the Gaia target catalog to allow ourselves to get closer to the plane without overwhelming ourselves with target density.

![image](https://user-images.githubusercontent.com/218471/65536123-a29b9f00-deb7-11e9-9148-7d2665027de1.png)
![image](https://user-images.githubusercontent.com/218471/65536127-a4fdf900-deb7-11e9-8f52-b2e7b6833b4c.png)

geordie666:
I've already embarked on a branch where I split GFAs and skies by HEALPixel so that we genuinely have all-sky catalogs. As part of that branch I will make sure that I also provide code so target classes can be read as a function of tile. This neatly solves some I/O slowness while maintaining the possibility of placing a tile anywhere on the sky.

But, Boris' work should be extremely useful in determining where we should or should not place *tiles*.

geordie666:
Addressed in #548 and #558.

PR #548, opened 2019-10-05
Update skies, GFAs and CMX targets to facilitate all-sky observing

geordie666:
This PR updates all classes except for SV and the Main Survey to produce targets across the entire sky. It includes:
- GFAs, skies and CMX targets can now be processed and written split-by-HEALPixel.
- New "bundling" scripts to parallelize GFAs, skies and CMX targets by HEALPixel.
- Target classes are now bundled across all HEALPixels (not just those that touch sweeps files).
- Pixel information is now added to file headers for GFAs, skies and CMX targets, allowing these flavors of files to be read in by RA/Dec box, cap and list of HEALPixels.
- All-sky CMX targets are written separately from in-footprint CMX targets.
- Back-up and first light target classes have been added for commissioning.
- A New `TARGETID` encoding scheme has been implemented to distinguish Gaia-only and first light targets from in-footprint targets.
- Commissioning targets that are processed from the Legacy Surveys are now "resolved".
- The `desitarget.io.read_` functions can now process SKY and GFA target files.
- A new function `desitarget.io.read_targets_in_tiles()`to read in targets restricted to a set of DESI tiles (or the entire tiling footprint).
- Boris Gaensicke's geographical cuts for Gaia (see #541) have been implemented (for CMX, but using a general function).


geordie666:
I'll merge this in about 24 hours, as tagging and producing new commissioning files is time-critical. There are a lot of changes here, so I expect unit test might fail, but I'll work to resolve any failures.

PR #552, opened 2019-10-14
fixed overlap in faint_ext bit

qmxp55:
There is a huge overlap between the faint_ext bit and the fibermag bit in BGS SV. This fixed that overlap by excluding the overleaped objects in the faint_ext.


geordie666:
@qmxp55: Please also add this new cut to the [SV targeting wiki](https://desi.lbl.gov/trac/wiki/TargetSelectionWG/SurveyValidation#bgs). Once you've done that, I'll confirm and merge this.

qmxp55:
@geordie666: I've updated the wiki and included the newest target densities too.

PR #553, opened 2019-10-14
Updates for ELG selections for SV

geordie666:
Includes, related to ELG selections for SV:
- Four new bit names: `ELG_SV_GTOT`, `ELG_SV_GFIB`, `ELG_FDR_GTOT`, `ELG_FDR_GFIB` that replace the old ELG selections.
- A new set of cuts for each of these new selections, which include north/south differences.
- The `FIBERFLUX_G` column is now propagated from the sweeps to facilitate SV ELG cuts.

Also includes some minor updates for commissioning work:
- Increase the default sky densities by a factor of 4x.
- Relax the limit of the `BACKUP_FAINT` CMX targets to G < 21.


geordie666:
I'll request a review of the ELG selection changes from Anand before merging.

Issue #557, opened 2019-10-18
consider adding obscon to target files


moustakas:
Spinning an off-list discussion with @sbailey and @geordie666 into an issue, the proposal is to explicitly include `obscon` in the target filenames i.e.,
`targets/dark/targets-*.fits`
would become
`targets/dark/targets-dark-*.fits`

And ditto for `bright` targets and the mock output files.

sbailey:
Related: currently `desitarget.io.write_targets` and `_bright_or_dark` do some rearrangement of the directories (and filenames?) to get the obscon and healpix numbers in there.  We need something like `findfile` that can reconstruct the canonical filepath given input like base directory, nside, healpix, obscon, etc.  Currently that is in `desitarget.mock.io.findfile`, but that isn't actually specific to mocks and it would be better for `desitarget.io` to provide that directly and mocks and desisim just use it.

geordie666:
Also related: Howsoever this `findfile` function is written, it would be convenient to have it stop at the directory-level if `healpix` is passed as `None`.

This is because the `desitarget.io.read*` convenience functions (`read_targets_in_hp`, `read_targets_in_tiles`, etc.) operate at the directory-level.

I can get to this on Monday, if nobody beats me to it.

geordie666:
Addressed in #558.

Issue #350, opened 2018-08-23
Backstop selection for targets


geordie666:
Several problems in assigning fibers beyond the edges of the survey and in CCD gaps could be solved with a very-low-priority set of targets. This would be different to `BAD_SKY` in that these would be potential targets rather than potential blank locations.

One option would be to just provide a Gaia target everywhere, in some sensible magnitude range.

Related to https://github.com/desihub/desitarget/issues/276.

geordie666:
Addressed in #548 and #558.

PR #558, opened 2019-10-22
General all-sky/backup targets and a new output data model/directory structure

geordie666:
Includes:
- Add all-sky/backup/supplemental targets for SV.
- Add all-sky/backup/supplemental targets for the Main survey.
- Write files split by `obscon` (dark/bright files) using a format like, e.g. `targets/main/dark/targets-dark*.fits`.
- A new output file for the backup/allsky targets (writes to, e.g. `targets/main/supp-targets/targets-*.fits`).
- A new function `io.find_target_files()` that can parse the typical output directory structure for targets to find files (and directories) of different types.


geordie666:
This PR should address #557. When combined with #548 it should also address #541 and #350.

py/desitarget/io.py
```diff
@@ -330,6 +330,8 @@ def _bright_or_dark(filename, hdr, data, obscon, mockdata=None):
     if not os.path.exists(newdir):
         os.mkdir(newdir)
     filename = os.path.join(newdir, os.path.basename(filename))
+    # ADM modify the filename with an obscon prefix.
+    filename = filename.replace("targets-", "targets-{}-".format(obscon.lower()))
```
moustakas: Are we punting on a `desitarget.io.findfile` routine in this PR? Just askin'.

geordie666: No punting here:

https://github.com/desihub/desitarget/pull/558/commits/1f135c6805acb811b8ff9e5dea71860eaa1ac58d

moustakas: Nice! However, neither `select_targets` nor `io.write_targets` seem to be using this function, and in fact `io.write_targets` is creating the filename on-the-fly rather than using the output of `io.find_target_files`.  It's exactly this behavior that originally broke the integration tests (specifically, `desisim.simexp` couldn't find the output files written by `desitarget.mock.build.targets_truth`.

To clarify this request, take a look at https://github.com/desihub/desitarget/blob/master/py/desitarget/mock/io.py#L30-L51 and how `desitarget.mock.io.findfile` is called by `desitarget.mock.build.targets_truth`: https://github.com/desihub/desitarget/blob/master/py/desitarget/mock/build.py#L971-L974

What both @sbailey and I were thinking is that `io.write_targets` would accept a filename that would not be modified internally, that way the same function could be used both for the data and the mocks.  And indeed, the two little functions in `mock.io` could in fact be moved to `desitarget.io` which could handle the mocks as well as the data.

I hope this is all clear and if not I'm happy to chat off-list.  Also: this request is low-priority relative to some of the other changes in this PR, so we shouldn't necessarily hold up this PR while we hammer this out.

geordie666: Ah, I see...

The existing functions in `desitarget.mock` won't work as-is, as they don't have a `survey` kwarg to distinguish between main survey ("main"), commissioning ("cmx") and SV ("sv1") files and file structures. So, some modification would be required. The mocks also don't write allsky ("supp") targets.

Leave that with me, though and I'll see what I can hack together.

Incidentally, there's a lot of functionality on the data side that isn't in the mocks (randoms, commissioning, supplemental file types, gfas, etc.). So, any resolution just to fix the mock/data integration tests is sort of cheating...just modifying desitarget.io.write_targets() is an oversimplification of the data model and directory structure for all of the target types.

geordie666: @moustakas: I've updated all of the principal `write` and `select` functions on the non-mock side to use `io.find_target_files`. The interface into `io.write_targets` is (necessarily) more complicated than for your mock functions, as I have to deal with `resolve` directories, `cmx/sv/main` directories, `supp` directories, etc.

I think it should be straightforward for the mock functions to call the refactored `select/write` functions, though. Let me know if this looks good enough, and we can merge so you can call the new functions for the mocks. Or, if you'd prefer, hijack this branch and update the mock side of the code directly before merging.

geordie666:
@moustakas: I updated the Main Survey/SV ``BACKUP`` targets in this branch based on Boris' final selections. It's tough to test the ``BACKUP`` selections without a full Gaia directory, though...so let's not merge this until NERSC returns and I can run full tests.

moustakas:
Thanks @geordie666, understood. 

I'm hoping to test this branch on `select_mock_targets` tomorrow or by Wednesday at the latest and will report back here.

geordie666:
@moustakas: I'm done with testing this branch, and I'd like to merge it by the middle of the day tomorrow to produce new targeting files for commissioning and SV.

If you don't find time to update the mocks by this time tomorrow, I'd like to go ahead and merge this PR. Note that @weaverba137 is intending to tag `desitarget` today while the end-to-end integration tests are working. So, it's probably not critical if you don't get to this by tomorrow. You could fix any problems in a separate branch.

moustakas:
@geordie666 I don't think I'll be able to test this today so please feel free to merge and I'll pick up a new branch from there.  Thanks.

PR #559, opened 2019-10-24
Update sv1_cuts.py

jinyiY:
Add cuts for z~4.3-4.8 quasar into the z5QSO selection

### DIFF
--- a/py/desitarget/sv1/sv1_cuts.py
+++ b/py/desitarget/sv1/sv1_cuts.py
@@ -784,7 +784,7 @@ def isQSOz5_cuts(gflux=None, rflux=None, zflux=None,
 
     Notes
     -----
-    - Current version (09/25/19) is version 101 on `the SV wiki`_.
+    - Current version (10/26/19) is version 101 on `the SV wiki`_.
     - See :func:`~desitarget.cuts.set_target_bits` for other parameters.
     """
     if not south:
@@ -810,7 +810,8 @@ def isQSOz5_cuts(gflux=None, rflux=None, zflux=None,
     if south:
         morph2 = dcs < 0.01
     else:
-        morph2 = dcs < 0.005
+    #currently identical, but leave as a placeholder for now
+        morph2 = dcs < 0.01
     qso &= _psflike(objtype) | morph2
 
     # ADM SV cuts are different for WISE SNR.


geordie666:
@jinyiY: This looks good to me:

I think there's one discrepancy between your cuts and what's reported on [the wiki](https://desi.lbl.gov/trac/wiki/TargetSelectionWG/SurveyValidation#QuasarsQSO44:ColorSelectionforz4.8Quasars). Currently, in the code, we make this cut on morphology in the northern imaging surveys:
```
morph2 = dcs < 0.005
```
but the wiki records just:
```
(dchisq[1] - dchisq[0])/dchisq[0] <0.01 
```
(presumably for both "north" and "south"). Can you confirm that you want the different morphology cuts in the north and the south? If so, update the wiki to reflect that and then I'll merge this PR.

Normally I'd ask you to update the `doc/changes.rst` file detailing this PR (which is good practice). But, don't worry, in the case of this short PR, I'll just update that in my current branch.

geordie666:
@jinyiY: I suggest that @Cyeche answer your question about morphology. He has worked most closely with the morphology cuts for the high-z quasars.

I agree, though, that it will only make a small difference. For SV (for your z~5 sample) we might as well just use the most permissive cut to be more complete, which would be dcs < 0.01?

Cyeche:
Yes  (type== ‘’PSF'’ or delta_chi2<0.01 is fine.

geordie666:
@jinyiY: One last thing: Once you've changed [the SV wiki](https://desi.lbl.gov/trac/wiki/TargetSelectionWG/SurveyValidation#QuasarsQSO44:ColorSelectionforz4.8Quasars) and/or the code based on your decision for the morphological cut, please update this line in the docstring of your function:

https://github.com/desihub/desitarget/blob/e66ff2e6f75bbf218fb66413c5dee753cfdefe0e/py/desitarget/sv1/sv1_cuts.py#L787

to the correct version from the wiki.

jinyiY:
@geordie666 @Cyeche : Thanks! Then I will change the code to use <0.01 only.

geordie666:
Thanks @jinyiY: If you could update this line in the docstring of your function:

https://github.com/desihub/desitarget/blob/e66ff2e6f75bbf218fb66413c5dee753cfdefe0e/py/desitarget/sv1/sv1_cuts.py#L787

to the correct version from the wiki, then I can merge this.

jinyiY:
@geordie666 : Thanks. I have updated the date of version but I recently can not log into the wiki. So could you please let me how to check the version from wiki? I am in China now and will be on the way to Chile. The internet connection may be limited until Oct 29.

geordie666:
@jinyiY: The wiki is down because of a power outage in California. Don't worry: I'll merge this this morning and then I'll update the version in a different branch.

PR #562, opened 2019-11-05
BGS main updates

qmxp55:
Added the Large galaxies and Globular cluster bitmasks. We have include a fibre magnitude cut too. This updates were already implemented in the wiki: https://desi.lbl.gov/trac/wiki/TargetSelectionWG/TargetSelection#BrightGalaxySurveyBGS

### DIFF
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -953,7 +953,7 @@ def _check_BGS_targtype_sv(targtype):
         raise ValueError(msg)
 
 
-def isBGS(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None,
+def isBGS(rfiberflux=None, gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None,
           gnobs=None, rnobs=None, znobs=None, gfracmasked=None, rfracmasked=None, zfracmasked=None,
           gfracflux=None, rfracflux=None, zfracflux=None, gfracin=None, rfracin=None, zfracin=None,
           gfluxivar=None, rfluxivar=None, zfluxivar=None, maskbits=None, Grr=None,
@@ -991,7 +991,7 @@ def isBGS(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None,
                          gfluxivar=gfluxivar, rfluxivar=rfluxivar, zfluxivar=zfluxivar, Grr=Grr,
                          gaiagmag=gaiagmag, maskbits=maskbits, targtype=targtype)
 
-    bgs &= isBGS_colors(gflux=gflux, rflux=rflux, zflux=zflux, w1flux=w1flux, w2flux=w2flux,
+    bgs &= isBGS_colors(rfiberflux=rfiberflux, gflux=gflux, rflux=rflux, zflux=zflux, w1flux=w1flux, w2flux=w2flux,
                         south=south, targtype=targtype, primary=primary)
 
     return bgs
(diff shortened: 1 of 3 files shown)


qmxp55:
FYI, I did a sanity check running desitarget for a couple of bricks and everything seems fine.

geordie666:
@qmxp55: Unit tests are failing because you updated the inputs for the `isBGS()` function without also updating them for the unit tests. You'll need to update the inputs here:

https://github.com/desihub/desitarget/blob/a5f5dd6eb7f41a8bd04d20e9ed790e0c8ac0dc0b/py/desitarget/test/test_cuts.py#L172-L179

to reflect your new `rfiberflux` input to `isBGS()`, e.g.:

https://github.com/desihub/desitarget/pull/562/files#diff-8e3b412655b1463f27d2f61e6a918bc6R956

Also, could you update the `changes.rst` file in:

https://github.com/desihub/desitarget/tree/master/doc

with a one-line description (e.g. `add bitmasks and rfiberflux cut for BGS Main Survey selection`).

geordie666:
Oh, also note that here:

https://github.com/desihub/desitarget/blob/a5f5dd6eb7f41a8bd04d20e9ed790e0c8ac0dc0b/py/desitarget/test/test_cuts.py#L103-L104

You'll need to add `rfiberflux = flux['RFIBERFLUX']`.

qmxp55:
@geordie666 Done. Thanks for the help!

geordie666:
@qmxp55: Unit tests are still failing. You need one more fix. In the `changes.doc` file:

https://github.com/desihub/desitarget/pull/562/commits/267c26a3dc64dba0671176888f68697d3719fe5b

you need to add a url for the link `[`PR #562`_]`. If you change your current text in that file to:

```
 * Add bitmasks (1,12,13) and rfiberflux cut for BGS Main Survey selection [`PR #562`_].

.. _`PR #562`: https://github.com/desihub/desitarget/pull/562
```
Then that should fix the unit tests.

py/desitarget/cuts.py
```diff
@@ -1062,6 +1069,18 @@ def isBGS_colors(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None,
         bgs &= rflux < gflux * 10**(4.0/2.5)
         bgs &= zflux > rflux * 10**(-1.0/2.5)
         bgs &= zflux < rflux * 10**(4.0/2.5)
+        
+    g = 22.5 - 2.5*np.log10(gflux.clip(1e-16))
+    r = 22.5 - 2.5*np.log10(rflux.clip(1e-16))
+    z = 22.5 - 2.5*np.log10(zflux.clip(1e-16))
+    rfib = 22.5 - 2.5*np.log10(rfiberflux.clip(1e-16))
```
moustakas: This change broke integration tests (currently debugging in https://github.com/desihub/desisim/pull/514) because we're not currently passing mock fiberfluxes to the targeting functions, and we don't test here against the possibility that `fiberflux is None` (the default input).

geordie666:
Apologies @moustakas: Would you like me to add a workaround today (this afternoon) or will you handle it?

moustakas:
If you could implement a quick fix I'd be grateful. And apologies I didn't catch this when the PR was up.

BTW, could you add a unit test where `fiberflux` is *not* passed, whether to the BGS or the LRG or the ELG (if relevant) selection functions?

geordie666:
OK, I'll do that.

geordie666:
@moustakas: Do you need this compatibility with the mocks for *just* the Main Survey, or also for SV?

moustakas:
Hmm, my inclination is that it's just for the main survey (which is all we're using for integration testing), although `svdc2019` was based on SV selections, so maybe we should handle that case, too.  But definitely no CMX.

py/desitarget/cuts.py
geordie666: This should now be addressed in #563.

PR #563, opened 2019-11-09
Strict NoneType checking for fiberfluxes

geordie666:
This PR checks for `NoneType` fiberfluxes and substitutes `FLUX_X` when `FIBERFLUX_X` is not available. The goal is to make it easier for the mocks to directly use `desitarget` cuts without having to create mock fiberfluxes. 

Also included are new unit tests to explicitly pass `fiberflux=None` to the relevant `desitarget.cuts()` functions, to try to ensure continued compatibility with the mocks.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,11 +5,14 @@ desitarget Change Log
 0.34.1 (unreleased)
 -------------------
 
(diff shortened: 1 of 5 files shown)


moustakas:
@geordie666 the test failures in https://github.com/desihub/desisim/pull/514 do not appear to be related to this PR, so please go ahead and merge when ready / happy.  And thank you.

Issue #568, opened 2019-12-04
test_mock_build failures


sbailey:
tests are failing at NERSC on desitarget master:

```
======================================================================
ERROR: test_targets_truth (desitarget.test.test_mock_build.TestMockBuild)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/global/common/software/desi/cori/desiconda/20190804-1.3.0-spec/code/desitarget/master/py/desitarget/test/test_mock_build.py", line 43, in test_targets_truth
    targets_truth(params, healpixels=[99737,], nside=256, output_dir=self.outdir, no_spectra=True)
  File "/global/common/software/desi/cori/desiconda/20190804-1.3.0-spec/code/desitarget/master/py/desitarget/mock/build.py", line 976, in targets_truth
    qso_selection='colorcuts', mockdata=mockdata)
  File "/global/common/software/desi/cori/desiconda/20190804-1.3.0-spec/code/desitarget/master/py/desitarget/io.py", line 436, in write_targets
    drint = int(indir.split("dr")[1][0])
AttributeError: 'NoneType' object has no attribute 'split'
```

@moustakas could you fix this quickly or otherwise punt it back?

geordie666:
I think the fix is just to add `AttributeError` to the tuple of exceptions here:

https://github.com/desihub/desitarget/blob/57e151e037d452078d3a848eeaf508b27ec4f390/py/desitarget/io.py#L435-L439

Some context: As Gaia-only targets don't have a release number, I needed a way to find the intended data release outside of the Legacy Surveys footprint, or fail gracefully. Looks like I forgot to catch the `NoneType` possibility in the exception.

moustakas:
What @geordie666 said.

geordie666:
I have fixed this in my working branch, which I hope to have a PR for on Friday afternoon. I can fix this earlier, if needed.

sbailey:
NERSC is out today so it isn't critical to get a PR today, but one tomorrow morning would be tangibly more useful than one tomorrow afternoon, if possible.  Once NERSC is back I'd like to proceed with integration testing and get a clean slate of tags for the OSU workshop, and this issue is currently a blocking factor for the integration testing.

geordie666:
This should be fixed in #570 (which also includes a number of other minor bug-fixes).

PR #570, opened 2019-12-06
Some further clean-up of glitches and minor bugs

geordie666:
Includes:
- Deprecate the Python 3.5 unit tests.
- Catch the `AssertionError` if a `NoneType` input directory is passed in `desitarget.io` when writing files (this should address #568).
- Assert the data model when reading secondary target files.
- Use `io.find_target_files()` to name the priminfo file (the file of primary matches) for secondaries.
- Allow N < 16 nodes when bundling files for slurm.
- Use the DR14Q file for SV, not the DR16Q file. By agreement with SDSS-IV/eBOSS we can only use DR16Q once the DR16Q fits file is made public.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -106,16 +106,16 @@ matrix:
         # Coverage test.  Note that installing the coverage software can
         # change the set of packages installed by conda, so we do
         # separate 'test' and 'test --coverage'.
-        - os: linux
-          env: PYTHON_VERSION=3.5 SETUP_CMD='test --coverage'
-               CONDA_DEPENDENCIES=$CONDA_ALL_DEPENDENCIES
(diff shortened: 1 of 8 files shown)


sbailey:
Thanks!

sbailey:
Two errors about errors:
  * in io.py you meant to add AttributeError rather than AssertionError
  * while testing this PR, I forgot that expensive mock tests only run if DESITARGET_RUN_MOCK_UNITTEST is set, and thus I wasn't testing the fix I thought I was testing.

I fixed AssertionError -> AttributeError directly in master, and confirmed that I could get my test to fail before the fix and have it work after the fix.

geordie666:
Oops, sorry. Thanks for catching my `AssertionError/AttributeError` error.

PR #573, opened 2019-12-13
Preparation for DR9, particularly the DR9c imaging files.

geordie666:
Includes:
- Update the data model for DR9c imaging files, while maintaining backwards-compatability with DR8.
- Don't set the `SKY` bit in `DESI_TARGET` when setting the `SUPP_SKY` bit.
- Users can now input a seed (1, 2, 3, etc.) to `bin/select_randoms`:
    - This user-provided seed is also added to the output file name and file header.
    - This facilitates generating a range of numbered random catalogs.
    - For instance, `randoms/randoms-dr8-hp-127,191-4.fits` for a seed of `4`.
- Write out final secondaries using `desitarget.io.find_target_files()`

### DIFF
--- a/bin/select_randoms
+++ b/bin/select_randoms
@@ -56,13 +56,16 @@ ap.add_argument("--noresolve", action='store_true',
 ap.add_argument("--aprad", type=float,
                 help="Radius of aperture in arcsec in which to generate sky background flux levels (defaults to 0.75; the DESI fiber radius)",
                 default=0.75)
+ap.add_argument("--seed", type=int,
+                help="Random seed passed to desitarget.randoms.select_randoms()",
+                default=1)
 
 ns = ap.parse_args()
 # ADM build the list of command line arguments as
 # ADM bundlefiles potentially needs to know about them.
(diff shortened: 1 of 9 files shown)


geordie666:
@sbailey: people are eager to have DR9-related data files, so I'd like to merge this in 24 hours or so and run some new target files.

Note that this should also address https://github.com/desihub/fiberassign/issues/212 (at the simple level of turning off the `SKY` bit when `SUPP_SKY` is set).

But, I'm not sure where you are with integration tests. So, feel free to tell me not to merge this in 24 hours if you're worried it will break the mocks. I tried to maintain backward compatibility, but you never know.

moustakas:
> But, I'm not sure where you are with integration tests. So, feel free to tell me not to merge this in 24 hours if you're worried it will break the mocks. I tried to maintain backward compatibility, but you never know.

I'm around and can fix integration test issues due to this PR quickly.

sbailey:
@geordie666 go ahead and merge when ready and I will retest integration afterward and submit tickets/PRs as needed.

PR #574, opened 2019-12-15
Updates for stellar SV and fixes to secondary target code

apcooper:
PR #530 had a couple of bugs with "&"s that should have been "|"s when checking which primary targets should be updated on the basis of their secondary bits. SCND_ANY should have been in the bitmask that identifies those updatable targets. 

Another bug: when secondary targets are merged into the MTL, they need to have SCND_ANY set in the DESI_TARGET bitmask for the subsequent logic to work correctly.

This PR also adds SV secondary target classes for BHBs, filler targets in the Hyades and the GD1 and Orphan streams.


sbailey:
@geordie666 could you review this today to include in our next tag?  Thanks.

geordie666:
Yep. I'll review this by this afternoon.

doc/changes.rst
```diff
@@ -26,18 +26,19 @@ desitarget Change Log
     * Useful to ensure ongoing compatibility with the mocks.
 * Bitmasks (1,12,13), rfiberflux cut for BGS Main Survey [`PR #562`_].
 * Implement a variety of fixes to `select_mock_targets` [`PR #561`_].
-* Fixes and updates to ``secondary.py`` [`PR #530`]:
+* Fixes and updates to ``secondary.py`` [`PR #530`_] and [`PR #574_`]:
     * Fix a bug that led to incorrect ``OBSCONDITIONS`` for secondary-only
```
geordie666: This way of updating the `changes` file is quick, but it also leads to PRs being out-of-order with versioning (i.e. `PR #574` should be included beneath version `0.35.0`, or a `no changes yet` version).

Please rewrite this so that `PR #574` has its own comment beneath the correct version (or `no changes yet` version) tag (which might just say, `fixes bugs in PR #530`).

py/desitarget/secondary.py
```diff
@@ -508,7 +508,8 @@ def match_secondary(primtargs, scxdir, scndout, sep=1.,
     if np.any(scnd_update):
         # APC Allow changes to primaries if the DESI_TARGET bitmask has
         # APC only the following bits set, in any combination.
-        update_from_scnd_bits = desi_mask['MWS_ANY'] & desi_mask['STD_BRIGHT'] & desi_mask['STD_FAINT'] & desi_mask['STD_WD']
+        log.info('Testing if secondary targets can update {} matched primaries'.format(scnd_update.sum()))
+        update_from_scnd_bits = desi_mask['SCND_ANY'] | desi_mask['MWS_ANY'] | desi_mask['STD_BRIGHT'] | desi_mask['STD_FAINT'] | desi_mask['STD_WD']
```
geordie666: Thanks for catching this general `&` instead of `|` bug.

My only concern: Doesn't this mean that _all_ secondary targets are allowed to update the priorities of the primaries? 

I thought we'd discussed that it was OK to update the priorities for primaries if they were _just_ MWS targets? This seems like any primary object that has `SCND_ANY` set will be allowed to trump primaries, which is dangerous for large-scale-structure targets that match a secondary. What would happen for a target with, say, just `LRG` and `SCND_ANY` set?

py/desitarget/mtl.py
```diff
@@ -70,13 +75,12 @@ def make_mtl(targets, obscon, zcat=None, trim=False, scnd=None):
         # APC Propagate a flag on which targets came from scnd
         is_scnd = np.repeat(False, len(targets))
         is_scnd[-nrows:] = True
+        # APC Assign the SCND_ANY bit to secondary targets. We need to do this
+        # APC before the NUMOBS_MORE logic below
+        log.info('Updating DESI_TARGET for secondaries...')
+        targets[desi_target][is_scnd] |= desi_mask.SCND_ANY
```
geordie666: I'm nervous about updating target bits within MTL itself. The target bits should be updated in the `cuts/secondary` functions, and MTL should just respond to those bits.

Could you assign the `SCND_ANY` bit to the relevant targets in `secondary.py` instead of in `mtl.py`? This would mean including the `DESI_TARGET` output column for the secondary files, but I'm OK with that. 

Sorry! It's just cleaner to only update the target bits outside of MTL.

py/desitarget/secondary.py
apcooper: My intention was to allow updates to a target with SCND_ANY set ONLY if that target **does not** have any other bits (e.g. LRG) set, *except* the specified MW bits. As written, I think this is what happens. .

The important thing is how that mask gets used in the next line, `scnd_update &= ((targets[desi_target] & ~update_from_scnd_bits) == 0)`. The (A &~B) is true if any bit is set in A that is *not* set in B, so checking `== 0` leaves the complement (no bit set in A that isn't set in B). 

Maybe this could be written in a less abstruse way, but I think it works as it is. Without the SCND_ANY in the 'updatable' mask, this logic would not update any secondary targets

For example:

```python
test_target = desi_mask['LRG'] | desi_mask['SCND_ANY']
print('Target bits:', test_target)
scnd_update = (test_target & desi_mask['SCND_ANY']) != 0
print('Is secondary:', scnd_update)
update_from_scnd_bits = (desi_mask['SCND_ANY'] | desi_mask['MWS_ANY'] | desi_mask['STD_BRIGHT'] | desi_mask['STD_FAINT'] | desi_mask['STD_WD'])
scnd_update &= ((test_target & ~update_from_scnd_bits) == 0)
print('Is updatable:', scnd_update)
```
```
Target bits: 4611686018427387905
Is secondary: True
Is updatable: False
```
And the positive case:
```python
test_target = desi_mask['MWS_ANY'] | desi_mask['STD_BRIGHT'] | desi_mask['SCND_ANY']
print('Target bits:', test_target)
scnd_update = (test_target & desi_mask['SCND_ANY']) != 0
print('Is secondary:', scnd_update)
update_from_scnd_bits = (desi_mask['SCND_ANY'] | desi_mask['MWS_ANY'] | desi_mask['STD_BRIGHT'] | desi_mask['STD_FAINT'] | desi_mask['STD_WD'])
scnd_update &= ((test_target & ~update_from_scnd_bits) == 0)
print('Is updatable:', scnd_update)
```
```
Target bits: 6917529062000820224
Is secondary: True
Is updatable: True
```

doc/changes.rst
apcooper: OK, thanks for the correction. 

I'm not up to speed on the protocol about whether to put the new PR under `0.35.0` vs. `no changes yet`. I merged with  the latest changes master and put this PR under `no changes yet`, hope that's ok.

py/desitarget/targets.py
```diff
@@ -526,9 +526,10 @@ def calc_priority(targets, zcat, obscon):
             if np.any(scnd_update):
                 # APC Allow changes to primaries if the DESI_TARGET bitmask has any of the
                 # APC following bits set, but not any other bits.
-                update_from_scnd_bits = (desi_mask['MWS_ANY'] &
-                                         desi_mask['STD_BRIGHT'] & desi_mask['STD_FAINT'] & desi_mask['STD_WD'])
+                update_from_scnd_bits = (desi_mask['SCND_ANY'] | desi_mask['MWS_ANY'] |
```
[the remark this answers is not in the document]
apcooper: See comment above.

py/desitarget/mtl.py
apcooper: OK -- sorry I might not have time to get this done and tested today. 

For my own reference (@geordie666 , would be useful to have a second opinion): on quick inspection it seems like I need to add a `DESI_TARGET` column to the `outdatamodel` in `secondary.py` and then populate this with `SCND_ANY` somewhere, maybe in the loop over input files in `read_files`.

I need to make sure the `DESI_TARGET` column doesn't overwrite anything for  the primaries and ends up with the right 'SV_' or whatever prefix. 

In the `|` vs `&` changes elsewhere in this PR I'm using `from desitarget.targetmask import desi_mask` to get the mask that will give `SCND_ANY`. Probably I shouldn't be doing this but rather getting the program-specific `desi_mask`?

apcooper:
Thanks for the review @geordie666, I think I've dealt with the points except for where to set `SCND_ANY`. I'll need to find time to do that tomorrow. I can handle it but further suggestions for where/how to do it properly are welcome.

py/desitarget/targets.py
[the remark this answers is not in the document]
geordie666: I confess, you did confuse me with the `&` of a -ve bit number. But, I've tested this again and it seems to work.

py/desitarget/secondary.py
geordie666: I confess, you did confuse me with the `&` of a -ve bit number. But, I've tested this again and it seems to work. Thanks for the informative demonstration, here.

doc/changes.rst
geordie666: Yes, this should be under `no changes yet` (or, actually, should _replace_ the phrase `no changes yet`, as the first change for the next tag).

Can you fix a few malformed links, please? You have links like
```
[`PR #530_`]  [`PR #574_`]
```
these should be 
```
[`PR #530`_]  [`PR #574`_]
```

py/desitarget/mtl.py
geordie666: Slightly more defensible than adding the `DESI_TARGET` column in MTL might be to do it when we write the files? That's not ideal, but we already add a lot of necessary information on the write (such as the `SUBPRIORITY` column) so I think we're stuck with that model.

This line:

https://github.com/desihub/desitarget/blob/f6369d5bcd04f536e0087068e971d410fc91a513/py/desitarget/io.py#L725

already gives you the correct output column (just change the `_` to `cols` and then use `cols[0]` as `DESI_TARGET` (or, it should correctly also revert to `SV1_DESI_TARGET` when appropriate).

Best, I think, would be to add the `DESI_TARGET` column in `secondary.finalize_secondary()`. Maybe just above this line:

https://github.com/desihub/desitarget/blob/f6369d5bcd04f536e0087068e971d410fc91a513/py/desitarget/secondary.py#L731-L732

I _think_ my `main_cmx_or_sv` trick from:

https://github.com/desihub/desitarget/blob/f6369d5bcd04f536e0087068e971d410fc91a513/py/desitarget/io.py#L725

will work here, too (although it depends on which exact columns are set).

My preference would be to add `DESI_TARGET` in `secondary.finalize_secondary()`, but if that proves convoluted, I'd be OK with adding it in `io.write_secondary()`, too

apcooper:
Back to work now -- should all be done. The `SCND_ANY` bit in `DESI_TARGET` is set in `finalize_secondary`.

doc/changes.rst
geordie666: Thanks, looks good now.

py/desitarget/mtl.py
geordie666: Much better to add `SCND_ANY` and `DESI_TARGET` in the new locations. Thanks.

geordie666:
@apcooper: This all looks good to me now. I'll merge this in the next hour or so.

PR #579, opened 2020-01-31
Further preparation for DR9, particularly the DR9d imaging files.

geordie666:
Includes:
- Add the new `SERSIC` columns for the DR9 data model.
- Force columns read from the bricks file to be lower-case in `desitarget.skyfibers.get_brick_info()` because, for the dr9c/dr9d test runs, columns in the bricks file have been both upper- and lower-case.
- Set the default `nside` to `None` when generating randoms, so the code will fail if `nside` is necessary but is not specified.
- Fix a numpy future/deprecation warning.

### DIFF
--- a/bin/select_randoms
+++ b/bin/select_randoms
@@ -37,7 +37,7 @@ ap.add_argument("--numproc", type=int,
                 help='number of concurrent processes to use [{}]'.format(nproc),
                 default=nproc)
 ap.add_argument('--nside', type=int,
-                help='Process random points in parallel in bricks that have centers within HEALPixels at this resolution (defaults to 4)',
+                help='Process random points in parallel in bricks that have centers within HEALPixels at this resolution (defaults to None)',
                 default=None)
 ap.add_argument('--healpixels',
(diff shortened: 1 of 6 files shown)


moustakas:
I feel like a decision was made somewhere, sometime in ancient history that all FITS table columns in data systems would be uppercase. Now, I know that we're inheriting these tables from the imaging surveys, but I wonder if we should cast the columns to uppercase, just so the *code* at least always works with uppercase.  My $0.02 -- certainly your call!

geordie666:
I'm choosing to cast to lower-case to maintain backwards compatibility with earlier iterations of the imaging. For example, [the DR8 "bricks" files](http://www.legacysurvey.org/dr8/files/#region-survey-bricks-dr8-region-fits-gz) have lower-case columns (see, e.g. `/global/project/projectdirs/cosmo/data/legacysurvey/dr8/south/survey-bricks-dr8-south.fits.gz`).

But, in actuality, it doesn't matter which may I cast things, as the upper-or-lower-case column that I read is only needed for one line of code. I'll think about your suggestion over the weekend and maybe make the change! I agree that your suggestion would be more aligned with _Data Systems_ conventions.

Issue #445, opened 2018-12-17
Fix QA density histogram binning


sbailey:
Something went wrong with the density histogram binning of the desitarget/0.27.0 QA on DR 7.1, e.g.

![](http://portal.nersc.gov/project/desi/users/adamyers/desitargetQA-dr7.1-0.27.0/histo-ELG.png)

I'm guessing that there is some pixel with extremely high density that is messing up the auto-binning (and should be inspected too).

geordie666:
Addressed in #584.

PR #584, opened 2020-02-10
Updates and fixes to QA for DR9

geordie666:
Includes:

- Command-line options to pre-process and downsample input files. This facilitates making smaller input files for speedier testing.
- Better labeling of QA output, including cleaning up some long-standing labeling bugs.
- Change the color of the points in scatter plots to black so they contrast with the blue hex-contours.
- Smarter clipping of over-dense (and under-dense) pixels in the 1-D histogram plots and sky maps (addresses #445).
- Print out the densest pixels for each target class. The RA/Dec of these pixels includes links to the Legacy Surveys viewer.

### DIFF
--- a/bin/run_target_qa
+++ b/bin/run_target_qa
@@ -37,6 +37,9 @@ ap.add_argument('--nosystematics', action='store_true',
 ap.add_argument("--numproc", type=int,
                 help='number of concurrent processes to use when generating plots [defaults to {}]'.format(nproc),
                 default=nproc)
+ap.add_argument("--downsample", type=int,
+                help='downsample the targets by (roughly) this value, as a speed-up.',
+                default=None)
 
 ns = ap.parse_args()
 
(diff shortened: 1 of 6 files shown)


geordie666:
Example output QA pages are [here](https://portal.nersc.gov/project/desi/users/adamyers/PR584).

sbailey:
Please check the image links and/or fix Apache permissions (e.g. via fix_permissions.sh -a on the test QA dir).  I'm getting broken image links on pages like https://portal.nersc.gov/project/desi/users/adamyers/PR584/BACKUP_BRIGHT.html

geordie666:
The broken images are deliberate. This occurs when a targeting class will be present in DR9 but was not run as part of DR8. I used the randoms and pixweight files from DR8 to test pixel-weights, as running DR9 is uninformative for weighting pixels (because DR9 covers so little of the sky).

These broken links should only be present for the few classes that have had their bit-names introduced or changed between DR8 and DR9...just the `BACKUP` classes and a few `MWS` classes, I think.

geordie666:
Actually, thinking some more, maybe I should just remove the `BACKUP_` classes completely. They aren't in the Legacy Surveys footprint (so the pixel-weights will mess them up) and there's little we can do about their selection, as they're just a uniform cut in Gaia. Plus, the `BACKUP_` classes tend to overwhelm the "standard" QA plots near the Galactic Plane. 

Let me know if you have any objection to me just removing them from the QA. If not, I'll excise them before we merge this PR.

sbailey:
No objections; OK to remove QA of `BACKUP_` classes.  Otherwise looks good to me.

PR #586, opened 2020-02-11
Minimal hack to get desitarget running from refcat workaround.

michaelJwilson:
Minimal hack to get desitarget running from refcat workaround.  Likely not an acceptable approach, but illustrates what needs fixed.  Maybe useful in the meantime.  Runs to completion.

PR #588, opened 2020-02-12
Fix bug when adding LSLGA galaxies into Main Survey BGS

geordie666:
The new process to add LSLGA galaxies as BGS targets in the Main Survey was failing for mocks because `REF_CAT` was a zero-length string. This PR updates the code to catch the 3 possible flavors of `REF_CAT` string:

- zero-length.
- has length and is of type `str`.
- has length and is of type `bytes`.

This supplants PR #586 and should fix issues with unit tests. Once everyone agrees that this works, I'll merge this and close PR #586.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,6 +5,8 @@ desitarget Change Log
 0.35.4 (unreleased)
 -------------------
 
+* Fix bug when adding LSLGA galaxies into Main Survey BGS [`PR #588`_]:
+    * Catch cases of bytes/str types as well as zero-length strings.
 * Updates and fixes to QA for DR9 [`PR #584`_]. Includes:
     * Options to pre-process and downsample input files to speed testing.
     * Better labeling of QA output, including cleaning up labeling bugs.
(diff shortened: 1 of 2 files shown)


geordie666:
@michaelJwilson: It would be excellent if you could check this works with the mocks.

@sbailey: I think this will fix end-to-end integration tests. It would be great if you could check that.

@qmxp55: It would be splendid if you could check that I didn't break your intended algorithm for including LSLGA galaxies as BGS targets.

michaelJwilson:
As a fix this works, and it runs for BGS generation - I don't think there's a need to test it with the rest.  Looks consistent with the intended LSLGA selection for BGS aswell.

That said, REFCAT in the astropy table is defined as S2.  Really, this should be propagated correctly as b'  ', as intended.  But we can catch the symptoms.

sbailey:
Between the "excellent", "great", and "splendid" requests I will confirm the "great" request that this fixes integrations tests at NERSC.  In particular:
```
export DESITARGET_RUN_MOCK_UNITTEST=True
python py/desitarget/test/test_mock_build.py
```
fails on master and works on this branch.

bytes vs. strings and fits libraries auto-promoting or not ... what a pain.

qmxp55:
@geordie666 : A bit late but I can confirm that it works properly for DR8 ('L2') and DR9d ('L4')

PR #596, opened 2020-03-06
Fixing b sleak

qmxp55:
This fix the leak of bright stars in BGS MAIN and BGS SV. The leak in Main was caused by the LSLGA objects we manually added back to BGS regardless of any of the BGS cuts and hence ignoring the Bright star mask. In SV we were introducing bright stars in the LOWQ (low quality) super set as this includes a sample of objects rejected by the bright star mask in BGS.


geordie666:
Thanks @qmxp55: This looks good to me. I'll leave this PR open for a while in case anyone else has comments, and then merge it later today.

PR #598, opened 2020-03-10
Update SV0 class in CMX to specifically flag the high-z quasar selection.

geordie666:
This PR creates a new targeting class in commissioning (`SV0_QSO_Z5`) that corresponds to the high-redshift (z~4.5-5.0) quasar selection. This high-z class was (and remains) a subset of `SV0_QSO`. But, the class is now also given its own bit to make it easier to track.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,6 +5,7 @@ desitarget Change Log
 0.36.1 (unreleased)
 -------------------
 
+* Flag the high-z quasar selection in CMX (as `SV0_QSO_Z5`) [`PR #598`_].
 * Leak of Bright Stars in BGS Main Survey and BGS SV fixed [`PR #596`_].
 * Restrict skies to the geometric boundaries of their brick [`PR #595`_].
 * Changes in CMX after running code for Mini-SV [`PR #592`_]. Includes:
@@ -17,6 +18,7 @@ desitarget Change Log
 .. _`PR #592`: https://github.com/desihub/desitarget/pull/592
 .. _`PR #595`: https://github.com/desihub/desitarget/pull/595
 .. _`PR #596`: https://github.com/desihub/desitarget/pull/596
+.. _`PR #598`: https://github.com/desihub/desitarget/pull/598
 
 0.36.0 (2020-02-16)
 -------------------
--- a/py/desitarget/cmx/cmx_cuts.py
+++ b/py/desitarget/cmx/cmx_cuts.py
@@ -749,7 +749,7 @@ def isSV0_QSO(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None,
     # ADM The individual routines return arrays, so we need
     # ADM a check to preserve the single-object case.
     if _is_row(rflux):
-        return qso_north[0]
+        return qso_north[0], qsoz5_north[0]
 
     return qso_north, qsoz5_north
 
--- a/py/desitarget/cmx/cmx_cuts.py
+++ b/py/desitarget/cmx/cmx_cuts.py
@@ -693,7 +693,7 @@ def isSV0_QSO(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=None,
     -----
     - Current version (02/19/20) is version 51 on `the cmx wiki`_.
     - Current version (03/10/20) for the high-z (QSO_Z5 selection)
-      is version 58 on `the cmx wiki`_.
+      is version 59 on `the cmx wiki`_.
     - Hardcoded for south=False.
     - Combines all QSO-like SV classes into one bit.
     """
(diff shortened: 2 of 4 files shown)


geordie666:
I think this is the last change prior to rerunning targeting for next week (for the March, 2020 dark time observing runs). I intend to merge this tomorrow morning and re-tag the code for a new run of targets.

geordie666:
@Cyeche: It would be good if you could briefly check that this looks like what you wanted. 

I've already checked that I recover 686 of @jinyiY's targets using the new bit and running `select_cmx_targets` on your working file (`/project/projectdirs/cosmo/data/legacysurvey/dr8/south/sweep/8.0/sweep-160p025-170p030.fits`), e.g.:

```
select_cmx_targets /project/projectdirs/cosmo/data/legacysurvey/dr8/south/sweep/8.0/sweep-160p025-170p030.fits $CSCRATCH --nobackup
```
then:

```
import fitsio 
import numpy as np 
targs = fitsio.read("/global/cscratch1/sd/adamyers/targets/cmx/cmx-targets-dr8-hp-X.fits") 
from desitarget.cmx.cmx_targetmask import cmx_mask 
np.sum(targs["CMX_TARGET"] & cmx_mask["SV0_QSO_Z5"] != 0)

Out[]: 686
```

jinyiY:
> @Cyeche: It would be good if you could briefly check that this looks like what you wanted.
> 
> I've already checked that I recover 686 of @jinyiY's targets using the new bit and running `select_cmx_targets` on your working file (`/project/projectdirs/cosmo/data/legacysurvey/dr8/south/sweep/8.0/sweep-160p025-170p030.fits`), e.g.:
> 
> ```
> select_cmx_targets /project/projectdirs/cosmo/data/legacysurvey/dr8/south/sweep/8.0/sweep-160p025-170p030.fits $CSCRATCH --nobackup
> ```
> 
> then:
> 
> ```
> import fitsio 
> import numpy as np 
> targs = fitsio.read("/global/cscratch1/sd/adamyers/targets/cmx/cmx-targets-dr8-hp-X.fits") 
> from desitarget.cmx.cmx_targetmask import cmx_mask 
> np.sum(targs["CMX_TARGET"] & cmx_mask["SV0_QSO_Z5"] != 0)
> 
> Out[]: 686
> `

> @Cyeche: It would be good if you could briefly check that this looks like what you wanted.
> 
> I've already checked that I recover 686 of @jinyiY's targets using the new bit and running `select_cmx_targets` on your working file (`/project/projectdirs/cosmo/data/legacysurvey/dr8/south/sweep/8.0/sweep-160p025-170p030.fits`), e.g.:
> 
> ```
> select_cmx_targets /project/projectdirs/cosmo/data/legacysurvey/dr8/south/sweep/8.0/sweep-160p025-170p030.fits $CSCRATCH --nobackup
> ```
> 
> then:
> 
> ```
> import fitsio 
> import numpy as np 
> targs = fitsio.read("/global/cscratch1/sd/adamyers/targets/cmx/cmx-targets-dr8-hp-X.fits") 
> from desitarget.cmx.cmx_targetmask import cmx_mask 
> np.sum(targs["CMX_TARGET"] & cmx_mask["SV0_QSO_Z5"] != 0)
> 
> Out[]: 686
> ```

@geordie666 Thanks so much for this new bit! But I am a little confused about the number. I ran my selection with this sweep and got more than 800 targets which is similar to the result from Christophe. So I am not sure why 686 targets are selected here. Is it caused by the difference between select_sv and select_cmx?

geordie666:
@jinyi: Yes, `select_sv_targets` will run the southern selection on southern sweeps files. `select_cmx_targets` always runs the _northern_ selections on _all_ of the sweeps. So, if you pass a southern sweep file to `select_cmx_targets` you will get a slightly "strange" result!

But, the `SV0_` classes were set up to correspond to the northern selections a long time ago, and it's hard-coded in `desitarget`.

jinyiY:
@geordie666 Thanks! But I think in my color selection code, northern and southern selections have the same cuts. So is that mean there are other differences not only in color cuts code ?

geordie666:
@jinyiY I think you _do_ have different color cuts in the north and south? This is because you shift the photometry using the `shift_photo_north` function:

https://github.com/desihub/desitarget/blob/87db47c3135b8cb28356f6cb13e78c51db7061e3/py/desitarget/cmx/cmx_cuts.py#L1157-L1158

(this shift is used in both `select_sv_targets` and `select_cmx_targets`).

If the code is running a "southern" selection, this shift isn't triggered, so your cuts are effectively different in the north and the south.

jinyiY:
> @jinyiY I think you _do_ have different color cuts in the north and south? This is because you shift the photometry using the `shift_photo_north` function:
> 
> https://github.com/desihub/desitarget/blob/87db47c3135b8cb28356f6cb13e78c51db7061e3/py/desitarget/cmx/cmx_cuts.py#L1157-L1158
> 
> (this shift is used in both `select_sv_targets` and `select_cmx_targets`).
> 
> If the code is running a "southern" selection, this shift isn't triggered, so your cuts are effectively different in the north and the south.

 Yes. I forgot it. That's the point. Thanks!

Cyeche:
Fine with me!

Issue #578, opened 2020-01-31
re-enable test_qa by updating basemap


alxogm:
Travis test "test_qa" has been disabled since we require to update the basemap version it uses as current version (basemap/1.0.7) is not longer available from anaconda (see comment in [PR 565](https://github.com/desihub/desitarget/pull/565#issuecomment-579598769))

If basemap is not present. test_qa fails with the error below, raised by:  

`QA.py: line 1740, in make_qa_plots`  

Traceback (most recent call last):
2322  File "/home/travis/build/desihub/desitarget/py/desitarget/test/test_qa.py", line 57, in test_qa_main
2323    imaging_map_file=self.pixmapfile, tcnames=tcnames)
2324  File "/home/travis/build/desihub/desitarget/py/desitarget/QA.py", line 2016, in make_qa_page
2325    weight=weight, imaging_map_file=imaging_map_file, mocks=mocks)
2326  File "/home/travis/build/desihub/desitarget/py/desitarget/QA.py", line 1740, in make_qa_plots
2327    pool.map(_generate_plots, list(targdens.keys()), reduce=_update_status)
2328  File "/home/travis/build/desihub/desitarget/py/desitarget/internal/sharedmem.py", line 678, in map
2329    raise pg.get_exception()
2330desitarget.internal.sharedmem.SlaveException: No module named 'mpl_toolkits.basemap'
2331Traceback (most recent call last):
2332  File "/home/travis/build/desihub/desitarget/py/desitarget/internal/sharedmem.py", line 265, in _slaveMain
2333    self.main(self, *self.args)
2334  File "/home/travis/build/desihub/desitarget/py/desitarget/internal/sharedmem.py", line 563, in _main
2335    r = realfunc(work)
2336  File "/home/travis/build/desihub/desitarget/py/desitarget/internal/sharedmem.py", line 629, in realfunc
2337    else: return func(i)
2338  File "/home/travis/build/desihub/desitarget/py/desitarget/QA.py", line 1693, in _generate_plots
2339    weights=weights[ii], max_bin_area=max_bin_area)
2340  File "/home/travis/build/desihub/desitarget/py/desitarget/QA.py", line 447, in qaskymap
2341    basemap = init_sky(galactic_plane_color='k', ax=ax[0])
2342  File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/desiutil/plots.py", line 378, in init_sky
2343    from mpl_toolkits.basemap import pyproj
[...]

PR #141, opened 2017-01-31
updated mock routines handling split between wd and mws_nearby files

forero:
Solves #137

PR #605, opened 2020-04-16
Update QA now basemap dependency has been removed from desiutil

geordie666:
This PR updates the `desitarget.qa`module now the basemap dependency has been removed by [desiutil PR #141](https://github.com/desihub/desiutil/pull/141).

The changes I had to make were relatively minor and involved:
- Reintroducing the QA unit tests (which addresses #578)
- Moving any global imports from `desiutil.plots` inside of `desitarget.qa` functions. These were interfering with the `matplotlib.use('Agg')` command that was still required in `desitarget.qa` to spoof the display. Most likely the matplotlib back end is already set somewhere in `desiutil.plots`, now.
- Removing any `fig, ax = plt.subplots(1)` commands and just using `plt` directly. I'm not sure why this wasn't working, but I may previously have been passing around `ax` sub-optimally when using the old basemap calls.

I tested this PR worked for real data [reasonably extensively](https://portal.nersc.gov/project/desi/users/adamyers/PR605/).


geordie666:
@sbailey: This PR integrates `desitarget` QA with [desiutil PR #141](https://github.com/desihub/desiutil/pull/141). The unit tests are failing for two reasons:
- The tests won't pass until [desiutil PR #141](https://github.com/desihub/desiutil/pull/141) is merged (I think). Otherwise the `basemap` dependency will persist.
- `from desispec.io.util import makepath` now imports a dependency on `specter`. This dependency is new to me, and must have occurred (I think) in the time since we deprecated QA unit tests back in January. I can try to fix this, but you may just know how to fix the problem (likely by updating `.travis.yml`)...
- ...alternatively, it's not clear that `makepath` is needed at all, to me (I inherited this code) so we could just remove it from [here](https://github.com/desihub/desitarget/blob/3c9ff37e667c49d51e9cc9fea6073fbbdb523d53/py/desitarget/QA.py#L1902) and [here](https://github.com/desihub/desitarget/blob/3c9ff37e667c49d51e9cc9fea6073fbbdb523d53/py/desitarget/QA.py#L1953). That should completely remove any dependency on `specter`.

sbailey:
Thanks.  I will proceed down the chain of dependencies to get these Travis tests to pass again.  We (I?) didn't intended desispec.io to require specter (which has come in via using the TraceSet class) just to import it.  I'll see if that can be easily disentangled.

sbailey:
Looks good; using desiutil/3.0.0 and desispec/0.34.1 Travis tests now pass too.  Thanks for updating this.

Issue #597, opened 2020-03-10
Add columns to randoms required to run through fiberassign


ashleyjross:
AJR added the following to prepare random files for fiberassign (in https://github.com/desihub/LSS/blob/master/Sandbox/miniSVcat/cattools.py)

	rmtl['TARGETID'] = np.arange(len(rall))
	rmtl['DESI_TARGET'] = np.ones(len(rall),dtype=int)*2
	rmtl['SV1_DESI_TARGET'] = np.ones(len(rall),dtype=int)*2
	rmtl['NUMOBS_INIT'] = np.zeros(len(rall),dtype=int) #these numobs aren't quite correct
	rmtl['NUMOBS_MORE'] = np.ones(len(rall),dtype=int)
	rmtl['PRIORITY'] = np.ones(len(rall),dtype=int)*3400
	rmtl['OBSCONDITIONS'] = np.ones(len(rall),dtype=int) #this needs to be made to work for all tile types
	rmtl['SUBPRIORITY'] = np.random.random(len(rall))

geordie666:
Thanks @ashleyjross. Can you clarify:

Do you actually want values _exactly_ like ```rmtl['PRIORITY'] = np.ones(len(rall),dtype=int)*3400``` or do you want more sophisticated behavior, such as a random being assigned an actual, realistic target type (such as "QSO", "ELG", overlapping "QSO/ELG" etc.) and the associated `PRIORITY` and `OBSCONDITIONS` for that target type?

Perhaps a reasonable answer is "let's do the simple thing in my comment for now, and work towards something more sophisticated later"?

ashleyjross:
For these, my idea is that we will purely be probing geometry, so all points get treated exactly the same (and it is important, e.g., to set things so that this is the case no matter the tile type). Above is just what I put together to make things work in my first attempt. The important information is what gets written to the FAVAIL HDU (not the actual assignment).

geordie666:
@ashleyjross: I now add the following columns automatically to the randoms:
```
TARGETID
DESI_TARGET
NUMOBS_INIT
NUMOBS_MORE
PRIORITY
OBSCONDITIONS
SUBPRIORITY
```
So, for DR9 (and henceforth), those columns will exist automatically in the random catalogs.

Hopefully these will be useful, because they'll have realistic values that are assigned in the same way they're assigned for the targets (for example, `TARGETID` will be correctly assembled from the brick information).

geordie666:
I'll close this soon.

[...]

PR #606, opened 2020-04-16
Add some new features and columns to the random catalogs

geordie666:
This PR adds extra columns and features to the random catalogs. It includes:
- Better error messages and default behavior for `bin/supplement_randoms`.
- Don't calculate the `APFLUX` (and `APFLUX_IVAR`) quantities if `aprad=0` is passed (as a speed-up).
- Pass the randoms through the `desitarget.targets.finalize()` and `desitarget.mtl.make_mtl()` functions to add the extra columns needed to run `fiberassign` on the randoms.
- Add the `BRICKID` column to the random catalogs, for easier indexing on bricks.
- Add a realistic `TARGETID` column to the randoms (and the associated `RELEASE` and `BRICK_OBJID` columns).
- Recognize failure modes more quickly (and fails more quickly), particularly in the `select_randoms` executable.
- Always write out both the `resolve` and the `noresolve` catalogs, so that separate random catalogs are always available for the northern and southern imaging.

Addresses #597.

### DIFF
--- a/py/desitarget/io.py
+++ b/py/desitarget/io.py
@@ -1055,20 +1055,6 @@ def write_randoms(targdir, data, indir=None, hdr=None, nside=None, supp=False,
         except (ValueError, IndexError, AttributeError):
             drint = None
 
-    # ADM add HEALPix column, if requested by input.
-    if nside is not None:
-        theta, phi = np.radians(90-data["DEC"]), np.radians(data["RA"])
-        hppix = hp.ang2pix(nside, theta, phi, nest=True)
-        data = rfn.append_fields(data, 'HPXPIXEL', hppix, usemask=False)
-        hdr['HPXNSIDE'] = nside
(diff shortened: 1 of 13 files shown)


py/desitarget/io.py
```diff
@@ -2208,3 +2217,13 @@ def _check_hpx_length(hpxlist, length=68, warning=False):
         else:
             log.critical(msg)
             raise ValueError(msg)
+
+
+def _check_both_set(hpxlist, nside):
```
sbailey, line 2222: Normally an `_underscore()` function is hidden and used only within the file that defines it, but this function seems useful enough that you are importing it in multiple bin/ scripts.  Let's upgrade it to a non-underscore function.

geordie666: Done. I fooled myself by making it local to `io.py` and then adopting it in basically every single executable in the `bin` directory! Bad form.

PR #615, opened 2020-05-05
add masks for qso_random_forest in correct function

echaussidon:
Hi all, 

I made a little mistake in my last commit. I set maskbit in wrong function : `isQSO_cuts`. There is no problem because I think this function is not used now... So I remove it.

And I set correct maskbit in the function : `isQSO_randomforest`

Sorry for that..

Edmond


echaussidon:
@Cyeche @geordie666

echaussidon:
@geordie666 If i want to update changes.rst file, I have to create a pull request ?

geordie666:
@echaussidon : The `isQSO_cuts` function is definitely still used, if only by the mock catalogs which can't easily access the RF. So I would retain the maskbits cuts in that function to make sure it has the same flag logic as the RF.

Also, please update the [Main Survey wiki page](https://desi.lbl.gov/trac/wiki/TargetSelectionWG/TargetSelection) to reflect the new maskbits choices (both for the color cuts and for the RF).

You can "guess" the next PR number by looking at the highest current number for any (closed or open) [issue](https://github.com/desihub/desitarget/issue) or [PR](https://github.com/desihub/desitarget/pulls) and incrementing by one. If your PR ends up having a different number than you thought it would, then you can just change it locally, commit "minor update to changes doc" and push the new commit to the PR using the name of your branch.

echaussidon:
@geordie666 OK I re-add masks for `isQSO_cuts`.

geordie666:
@echaussidon : Looks good, thanks. I'm just running my standard checks at NERSC and then I'll merge this.

Issue #616, opened 2020-05-07
Should QSOs start with NUMOBS_INIT=4 or NUMOBS_INIT=1?


geordie666:
Currently, all QSOs start with 4 observations (`NUMOBS_INIT=4`). This number is altered by MTL according to the following rules after 1 pass of spectroscopic observations:

- For confirmed LyA quasars (`z >= 2.1`; `ZWARN = 0`) the number of additional observations (`NUMOBS_MORE`) is set to `NUMOBS_INIT`-`NUMOBS`, where `NUMOBS` is the number of observations so far.
- For objects with dubious spectra (`ZWARN > 0`), `NUMOBS_MORE` is set to `NUMOBS_INIT`-`NUMOBS`.
- For confirmed "tracer" quasars (`z < 2.1`; `ZWARN = 0`), `NUMOBS_MORE` is set to zero.

We could change to `NUMOBS_INIT=1` with the following rules:

- For confirmed LyA quasars (`z >= 2.1`; `ZWARN = 0`) the number of additional observations (`NUMOBS_MORE`) is set to `4-NUMOBS`, where `NUMOBS` is the number of observations so far.
- For objects with dubious spectra (`ZWARN > 0`), `NUMOBS_MORE` is set to `4-NUMOBS`.
- For confirmed "tracer" quasars (`z < 2.1`; `ZWARN = 0`), `NUMOBS_MORE` is set to zero.

As fiberassign [does not know anything about the PASS structure](https://github.com/desihub/fiberassign/issues/265), targets on any given PASS with `NUMOBS_MORE > 1` could receive multiple fibers in different tiles. So, the choice outlined above really amounts to whether we want QSO targets to be able to have multiple observations on the first PASS.

A benefit of the `NUMOBS_INIT=4` choice is that we "get ahead" on LyA quasars.

A drawback of the `NUMOBS_INIT=4` choice is that we could "waste" a second fiber in a given pass on confirmed `z < 2.1` QSOs.

Another upshot of the `NUMOBS_INIT=4` choice, which might be considered a benefit or a drawback, is that we obtain an early second observation of targets with ambiguous classifications and redshifts.

geordie666:
We ended up using `NUMOBS_INIT=4` for QSO targets.

PR #617, opened 2020-05-07
Ensure NUMOBS_MORE=1 for tracer QSOs that are also other targets

geordie666:
This PR makes (mostly minor) fixes and updates to MTL. 

The major change is to ensure that tracer QSOs that are in both the QSO target class and another target class drop to `NUMOBS_MORE = 0` for the Main Survey. Also:
- Separates out the calculation of `NUMOBS_MORE` into its own function called `desitarget.targets.calc_numobs_more()`.
    - There was enough special logic in `desitarget.mtl.make_mtl()` that I felt a separate function was justified.
- Consistently uses `zcut = 2.1` to define Lyman-Alpha QSOs. 
    - The codebase used two different redshifts in two different places.
- Adds a new unit test to enforce the check that `NUMOBS_MORE` always drops to 0 for tracer QSOs.
- Adds a new unit test to check BGS always gets `NUMOBS_MORE = 1` in the `BRIGHT` layer.
- Enforces a maximum seed in `desitarget.randoms.randoms_in_a_brick_from_edges()`.

### DIFF
--- a/py/desitarget/targets.py
+++ b/py/desitarget/targets.py
@@ -399,12 +399,12 @@ def calc_numobs_more(targets, zcat, obscon):
     targets : :class:`~numpy.ndarray`
         numpy structured array or astropy Table of targets. Must include
         the columns `DESI_TARGET`, `BGS_TARGET`, `MWS_TARGET`
-        (or their SV/cmx equivalents).
+        (or their SV/cmx equivalents) `TARGETID` and `NUMOBS_INIT`.
     zcat : :class:`~numpy.ndarray`
-        numpy structured array or Table of redshift information. Must
-        include 'Z', `ZWARN`, `NUMOBS` and be the same length as
-        `targets`. May also contain `NUMOBS_MORE` if this isn't the
-        first time through MTL and `NUMOBS > 0`.
+        numpy structured array or Table of redshift info. Must include
+        `Z`, `ZWARN`, `NUMOBS` and `TARGETID` and BE SORTED ON TARGETID
+        to match `targets` row-by-row. May also contain `NUMOBS_MORE` if
+        this isn't the first time through MTL and `NUMOBS > 0`.
     obscon : :class:`str`
         A combination of strings that are in the desitarget bitmask yaml
         file (specifically in `desitarget.targetmask.obsconditions`), e.g.
@@ -425,8 +425,8 @@ def calc_numobs_more(targets, zcat, obscon):
           of 1 in bright time and QSO "tracer" targets which always get
           NUMOBS_MORE=0 in dark time.
     """
-    # ADM check the input arrays are the same length.
-    assert len(targets) == len(zcat)
+    # ADM check input arrays are sorted to match row-by-row on TARGETID.
+    assert np.all(targets["TARGETID"] == zcat["TARGETID"])
 
     # ADM determine whether the input targets are main survey, cmx or SV.
     colnames, masks, survey = main_cmx_or_sv(targets, scnd=True)
@@ -483,12 +483,12 @@ def calc_priority(targets, zcat, obscon):
     targets : :class:`~numpy.ndarray`
         numpy structured array or astropy Table of targets. Must include
         the columns `DESI_TARGET`, `BGS_TARGET`, `MWS_TARGET`
-        (or their SV/cmx equivalents).
+        (or their SV/cmx equivalents) and `TARGETID`.
     zcat : :class:`~numpy.ndarray`
-        numpy structured array or Table of redshift information. Must
-        include 'Z', `ZWARN`, `NUMOBS` and be the same length as
-        `targets`. May also contain `NUMOBS_MORE` if this isn't the
-        first time through MTL and `NUMOBS > 0`.
+        numpy structured array or Table of redshift info. Must include
+        `Z`, `ZWARN`, `NUMOBS` and `TARGETID` and BE SORTED ON TARGETID
+        to match `targets` row-by-row. May also contain `NUMOBS_MORE` if
+        this isn't the first time through MTL and `NUMOBS > 0`.
     obscon : :class:`str`
         A combination of strings that are in the desitarget bitmask yaml
         file (specifically in `desitarget.targetmask.obsconditions`), e.g.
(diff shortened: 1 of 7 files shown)


geordie666:
@forero: Can you check my new logic for MTL in this branch? I could easily be missing something.

The critical update is that tracer QSOs no longer require a check on `SPECTYPE=QSO` to drop to `NUMOBS_MORE=1`. The check on `SPECTYPE=QSO` meant that mixed (e.g.) `QSO|ELG` targets (i.e. targets that were both in the `QSO` and the `ELG` target classes) for which we obtained (good) _galaxy_ spectra at `z < 2.1` would revert to `NUMOBS_MORE=3` because they were not confirmed tracer _QSOs_ based on `SPECTYPE`.

Consider this snippet of code:
```
# ADM Redshift cut-off for Lyman-alpha QSOs. 
zcut = 2.1 
import numpy as np  
from astropy.table import Table, join  
from desitarget import mtl  
from desitarget.targets import initial_priority_numobs  
from desitarget.targetmask import desi_mask  
# ADM 6=(merged ELG/QSO target).
for bit in [6]:  
    targets = Table()  
    # ADM 4 targets of this kind. 
    classes = [bit, bit, bit, bit]  
    n = len(classes)  
    targets['DESI_TARGET'] = classes  
    # ADM the BGS and MWS target bits need to be set, but they don't affect dark-time MTL.  
    targets['BGS_TARGET'] = np.zeros(n, dtype=np.int64)  
    targets['MWS_TARGET'] = np.zeros(n, dtype=np.int64)  
    # ADM this needs to be a unique TARGETID. We'll just use the integers 0-4.  
    targets['TARGETID'] = np.arange(n)  
    # ADM determine the initial PRIORITY and NUMOBS for the input target classes.  
    pinit, ninit = initial_priority_numobs(targets, obscon="DARK|GRAY")  
    targets["PRIORITY_INIT"] = pinit  
    targets["NUMOBS_INIT"] = ninit  
  
    zcat = Table()  
    zcat['TARGETID'] = targets['TARGETID']      
    # ADM pick two redshifts above the Lyman-Alpha cutoff and two below.  
    zcat['Z'] = [zcut+0.1, zcut+0.1, zcut-0.1, zcut-0.1]  
    # ADM make everything a GALAXY spectrum.  
    zcat['SPECTYPE'] = ['GALAXY', 'GALAXY', 'GALAXY', 'GALAXY']  
    # ADM flag warnings in one Lyman-alpha QSO and one tracer.  
    zcat['ZWARN'] = [4, 0, 4, 0]  
    # ADM each of our targets has one spectroscopic observation.  
    zcat['NUMOBS'] = [1, 1, 1, 1]  
     
    print("...{}...".format(desi_mask.names(bit)))  
    mtltargets = join(mtl.make_mtl(targets, obscon="DARK|GRAY", zcat=zcat), zcat)  
    print(mtltargets['TARGETID', 'Z', 'ZWARN', 'PRIORITY_INIT', 'NUMOBS_INIT', 'PRIORITY', 'NUMOBS_MORE']) 
```
In `master`, the output is:
```
TARGETID  Z  ZWARN PRIORITY_INIT NUMOBS_INIT PRIORITY NUMOBS_MORE
-------- --- ----- ------------- ----------- -------- -----------
       0 2.2     4          3400           4     3400           3
       1 2.2     0          3400           4     3500           3
       2 2.0     4          3400           4     3400           3
       3 2.0     0          3400           4     3000           3
```
In this branch (`ADM-MTL-fixes`) the output is:
```
TARGETID  Z  ZWARN PRIORITY_INIT NUMOBS_INIT PRIORITY NUMOBS_MORE
-------- --- ----- ------------- ----------- -------- -----------
       0 2.2     4          3400           4     3400           3
       1 2.2     0          3400           4     3500           3
       2 2.0     4          3400           4     3400           3
       3 2.0     0          3400           4        2           0
```
I think this branch is correct (i.e. once we know something is at `z < 2.1` with `ZWARN=0` we want to drop to `NUMOBS_MORE` to `0`)?

geordie666:
@forero: I'll merge this PR at the same time I merge #619 (for which I've requested multiple reviews from other people). So, you still have some time to review this PR #617 (if you want to).

forero:
@geordie666 apologies for the late reply! I will review this tomorrow afternoon. Thanks.

py/desitarget/targets.py
```diff
@@ -387,6 +390,90 @@ def initial_priority_numobs(targets, scnd=False,
...
+    -----
+        - Will automatically detect if the passed targets are main
+          survey, commissioning or SV and behave accordingly.
+        - Most targets are updated to NUMOBS_MORE = NUMOBS_INIT-NUMOBS.
+          Special cases include BGS targets which always get NUMOBS_MORE
+          of 1 in bright time and QSO "tracer" targets which always get
+          NUMOBS_MORE=0 in dark time.
+    """
+    # ADM check the input arrays are the same length.
+    assert len(targets) == len(zcat)
```
forero, line 429: The function assumes that the items in `targets` and `zcat` come in the same order. There has to be some check on that internal to the function (just in case this function ends up being used outside `make_mtl`)

py/desitarget/targets.py
```diff
@@ -387,6 +390,90 @@ def initial_priority_numobs(targets, scnd=False,
...
+        desi_target, bgs_target, mws_target, scnd_target = colnames
+        desi_mask, bgs_mask, mws_mask, scnd_mask = masks
+    else:
+        cmx_mask = masks[0]
+
+    # ADM Default is 0 , i.e. no more observations.
+    numobs_more = np.zeros(len(targets), dtype='i8')
+
+    # ADM main case, just decrement by NUMOBS.
+    numobs_more = np.maximum(0, targets['NUMOBS_INIT'] - zcat['NUMOBS'])
```
forero, line 444: We have to keep in mind that this logic will give zero `numobs_more` to all quasars if we start QSOs with NUMOBS_INIT=1. (https://github.com/desihub/desitarget/issues/616).

py/desitarget/targets.py
geordie666: This is an excellent point; thanks for catching this. I've updated the `assert` statement to check that the `targets` and `zcat` catalogs have the same order on `TARGETID`. This was _such_ a good point, that it made me also realize that we need the same check in the `targets.calc_priority()` function, so I added it there, too!

py/desitarget/targets.py
geordie666: Agreed. If we do change to `NUMOBS_INIT=1` for QSOs, we will need more special logic in the new `targets.calc_numobs_more()` function (which we would have needed anyway, previously, in the `mtl.make_mtl()` function).

geordie666:
@forero: Thanks for the useful review. 

I addressed your first (really excellent) point with new code. 

To your second point: I agree that we will need new logic if we decide to switch to `NUMOBS_INIT=1` for QSOs, but let's deal with that new logic in its own PR if we _do_ make the decision to switch.

Let me know if you want further changes, or if I can merge this PR.

forero:
@geordie666 I am happy with the latest update. I agree that introducing new logic for `NUMOBS_INIT=1` in QSOs can wait.

Issue #486, opened 2019-03-27
Should unobserved quasar targets have higher priority than Lyman-alpha quasars?


geordie666:
Documenting a conversation with Daniel Eisenstein, @nathalieGit and @sbailey at a targeting workshop. Patching in @andreufont for an opinion.

Currently, an unobserved quasar has lower priority than a Lyman-alpha quasar:

https://github.com/desihub/desitarget/blob/master/py/desitarget/data/targetmask.yaml#L166

This will mean that for close pairs of Lyman-alpha quasars one of the pair of quasars will *never* be observed, unless we happen to have more than 4 passes in a region with a pair.

Is this the desired behavior?

A simple change would be to instead give an unobserved quasar a *higher* priority than a Lyman-alpha quasar. This would mean that for close pairs, one quasar would be observed and if it's a Lyman-alpha quasar then the *other* quasar would be observed. If the *other* quasar is a Lyman-alpha quasar then it would receive 3 more observations.

So, the trade-off is that, if we make the proposed simple change then we would recover more close pairs of quasars with 3+1 observations instead of 4+0 observations. Behavior would be unaffected where there are no pairs of quasar targets within the fiber patrol radius.

geordie666:
Note: I appreciate that ultimately the Lyman-alpha working group may give us a more complex schema for prioritizing quasars. But, this simple change would ensure, at least, 3+1 instead of 4+0.

andreufont:
Hi @geordie666 - We had a similar discussion with Kyle in an e-mail exchange back in August, I copy paste it here. I don't think we ever finished the discussion, probably good to get it started again.

On 14 August 2018 at 22:41, Kyle Dawson <kdawson@astro.utah.edu> wrote:

    Hello all, particularly clustering and lyman alpha folks,

    We have several conditions on fiber collisions that we need to address in the fiber 
    assignment code. We need input from you to decide which object to give a fiber in 
    the case that two high priority objects compete for the same fiber and we already 
    have some information from prior observations. The obvious cases are:

    (a) two quasar targets, one is confirmed z>2.1 with only one epoch and the other 
    has not been observed.  Do we choose deeper observations of the known z>2.1 
    quasar or do we choose to increase sample size by taking an epoch of the unknown 
    quasar target?  
    ***PROPOSAL:  assign fiber to the known quasar.  
    LOGIC:  the second target has <20% likelihood of being a z>2.1 QSO, so spend the 
    time collecting data on the sure thing

[...]

Issue #349, opened 2018-08-23
Separate out the QSO bits that are based on the random forest


geordie666:
The random forest quasar selection is currently, effectively, two different selections, one targeted at high-z quasars and one targeted at tracer quasars. Even though these will be treated the same for initial fiber assignment, the high-z flavor should really have a separate bit set for provenance.

geordie666:
Addressed in #619.

PR #619, opened 2020-05-12
Various updates to targeting bits and MTL

geordie666:
This PR makes a range of updates, mostly to how targets are prioritized in MTL. It includes:
- Don't select any `BGS_WISE` targets in the Main Survey.
- Always set BGS targets that have a bad spectroscopic observation (`ZWARN > 0`) to a (very low) priority of `DONE`.
- Add an informational bit for the Main Survey indicating which QSOs are selected with the high-redshift Random Forest (addresses #349 ).
- MWS/BGS joint targets are now driven by the MWS priority _for the second observation_. Unobserved MWS targets will still always have a higher priority than repeat (`MORE_ZGOOD`) BGS targets, but targets that have both MWS and BGS bits set will be driven by the BGS `MORE_ZGOOD` priority rather than by the MWS priority.
- Update the default priorities for reobserving Lyman-alpha QSOs (as described in #486, which this addresses).


geordie666:
@apcooper: I'd appreciate your review on whether I've messed up MTL for the MWS in this PR. In `master`, every time a joint BGS/MWS target requested an additional observation it would be at the (higher) MWS initial (`UNOBS`) priority. This would mean that for a fiber-collided pair of BGS targets, one of which was also an MWS target, we'd always observe the joint BGS/MWS target, _even if `ZWARN > 0` for the spectroscopic observation_. The way I've set things up in this branch, we'd instead split observations between the MWS/BGS joint target and the BGS target. I'm happy to change back to the behavior in `master` if you think my new approach is incorrect, or we could make new-and-improved choices for the `MORE_ZGOOD` and `MORE_ZWARN` priorities for MWS targets.

@Cyeche: It would be great if you can confirm that my new "informational" bit indicating which QSO targets are from the _high-z_ RF doesn't affect target selection behavior for the QSOs at all (other than turning on one new informational bit).

@michaelJwilson: Can you confirm that these are the two changes requested by the BGS Working Group that you communicated to me:
1. Turning off `BGS_WISE`.
2. Ensuring repeat observations of BGS targets with `ZWARN>0` are always set to a very low priority (i.e. are effectively `DONE`).

michaelJwilson:
@geordie666 I understand the intention to be the correct one.  I'll ask for any last minute objections at the BGS telecon on Friday and hopefully have a look at the implementation aswell.

apcooper:
Just to add:

> we could make new-and-improved choices for the MORE_ZGOOD and MORE_ZWARN priorities for MWS targets.

This is probably something we (MWS) will want to look into again in the near future, as we shift attention from SV planning back to the main survey.

geordie666:
Thanks for your review, @apcooper. I was most concerned that I was messing up the MWS behavior in some way, so I'll merge this tomorrow even if I receive no further reviews.

michaelJwilson:
@geordie666 I finally got an opportunity to bring this up at a regular telecon.  

--  BGS_WISE should be parked. 
--  BGS does not want to re-observe failed redshifts, as this couples the assignment probabilities to redshifting & target type, which leads to a much harder correction.  

So yes to both.  I think 'effectively done' is fine in terms of implementation, but really we want done done.

geordie666:
Thanks @michaelJwilson. I believe that both of those BGS requests were implemented in this (now merged) PR. Of course, if you find otherwise, or want additional different behavior, let me know!

Issue #534, opened 2019-09-11
Discard SUPP_SKY near SKY


geordie666:
After discussions with @sbailey about not having `SUPP_SKY` where we don't need them:

In a HEALPixel of `NSIDE` that is appropriate to the density at which `SKY` fibers were generated, discard any `SUPP_SKY` sources if there is already an existing `SKY` in that HEALPixel.

geordie666:
This will ultimately be done in fiberassign instead, I think, via a different category of sky fiber. @sbailey: can we close this?

sbailey:
I'd like to keep this open for now though it's ok to prioritize other tickets instead.  We would like to have fiberassign handle this directly, but we're also near our memory limits for fiberassign so it could be helpful to trim out SUPP_SKY targets that we know fiberassign won't use anyway to save fiberassign time and memory.

geordie666:
Addressed in #621.

PR #621, opened 2020-05-22
Repartition sky files so sky locations lie within HEALPixel boundaries

geordie666:
Currently, sky targets are processed in bricks for which the _brick centers_ lie within a HEALPixel or HEALPixels written to the output file header. This was a deliberate choice to help parallelize brick look-ups but produces files that have a different convention to other target classes. For other classes, the targets _themselves_ are constrained to lie within the HEALPixel(s) written to the file header.

This PR provides an after-burner to repartition sky files such that the sky locations themselves lie within the HEALPixels listed in the file header, which aligns the skies with the convention for other target files. Tools added in this PR include:
- `desitarget.io.is_sky_dir_official()` to check that skies are partitioned correctly.
- `bin/repartition_skies` to reassign skies to the correct HEALPixels.

Alongside these changes, I updated functionality to make it easy to remove supplemental skies that share a HEALPixel with an existing sky location. So, this PR also addresses #534. 

Finally, this PR includes some speed-ups; notably `desitarget.io.read_targets_in_hp()` is now 5-10x quicker for most use cases.


geordie666:
A few plots of HEALPixels to show this addresses #534. Sky locations are on the left (red), the new supplemental sky locations are on the right (blue). In master, the supplemental sky locations would densely populate the entire pixel.
Screen Shot 2020-05-22 at 4 12 29 PM
Screen Shot 2020-05-22 at 4 10 38 PM

PR #623, opened 2020-05-29
MWS fall SV

apcooper:
Updates to the MWS `sv1_targetmask` to handle a few changes arising from the move to a later SV.

* Make a few target class names more generic
* Remove ORPHAN target class and add CV class
* Lower the priority of SEGUE targets

This requires an update to the  secondary target input files. That can be done as soon as the PR is merged.


geordie666:
@apcooper: This looks excellent. Thanks for continuing to be proactive on the MWS targeting bits. Feel free to merge this whenever you're ready.

Remember that if you change the actual secondary selections themselves, then you'll also need to update the documentation that describes the selection in `$SCND_DIR/sv1/docs/`

PR #624, opened 2020-05-29
 Functionality to allow iterations of SV beyond sv1

geordie666:
At some point we will need iterations of SV beyond the "sv1" formalism, if only for the One Percent Survey. This PR makes the necessary updates to generalize to svX (allowing sv2, sv3, sv4 etc.). It includes:

- A script `make_new_sv` to create the necessary files and directories for new iterations of SV. The script will build an `svX` directory from an existing `sv(X-1)` directory (`sv2` from an existing `sv1` directory, `sv3` from an existing `sv2` directory, etc.).
- Generalized handling of `"bitmask.yaml"` and `"cuts.py"` files for all iterations of `svX` (`X` being any integer).
- Updates to `desitarget.targets.main_cmx_or_sv()` to handle any SV file (e.g. that function will automatically return `sv4`-style masks and `SV4_DESI_TARGET`-like column descriptors for a file constructed using `SV4` cuts files and bitmasks).
- Altered output target directory names for SV files. SV targets are written to a directory that begins `/targets/svX/` instead of just `targets/sv/`.
- A new `TARGETID` formalism for secondary targets that will yield unique IDs for different iterations of SV.
    - The schema is `RELEASE=(X-1)*100 + SCND_BIT` for SVX-like surveys...
    - ...and `RELEASE=5*100 + SCND_BIT` for the Main Survey...
    - ...it was previously just `RELEASE=SCND_BIT`, which would have conflicted if we have new secondary target bits for different iterations of SV (and the Main Survey).

### DIFF
--- a/bin/make_new_sv
+++ b/bin/make_new_sv
@@ -1,0 +1,55 @@
+#!/usr/bin/env python
+
+import os
+import numpy as np
+
+#import warnings
+#warnings.simplefilter('error')
+
+from pkg_resources import resource_filename
+from glob import glob
+from time import time
+start = time()
+
+from desiutil.log import get_logger
+log = get_logger()
+
+from argparse import ArgumentParser
+ap = ArgumentParser("Make a new sv directory, based on the highest existing sv directory (i.e., if sv1 exists, make sv2)")
+
+# ADM the standard root name for a desitarget SV directory.
+svroot = resource_filename('desitarget', 'sv')
+
+# ADM retrieve the highest current sv directory.
+fns = glob(svroot+'*')
+svlist = [int(fn[-1]) for fn in fns if os.path.isdir(fn)]
+maxsv = np.max(svlist)
+
+# ADM copy the highest previous SV directory to make a new one.
+newdir = "{}{}".format(svroot, maxsv+1)
+cmd = 'cp -R {}{} {}'.format(svroot, maxsv, newdir)
+success = os.system(cmd)
+if success != 0:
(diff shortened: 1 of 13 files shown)


geordie666:
The `make_new_sv` utility script makes it easy to test this branch works. Potentially useful notes for interested parties who want to try the new functionality:

- run `make_new_sv`, which will create a new `desitarget.sv2` directory from the existing `desitarget.sv1` directory.
- run `select_sv_targets` with the `--iteration 2` flag to utilize the new `sv2` functionality, e.g.:
```
drdir=/global/cscratch1/sd/adamyers/dr9j
select_sv_targets $drdir/south/sweep $CSCRATCH/blat -s2 $drdir/north/sweep --nosecondary --nobackup --iteration 2
```
- As `sv2` was built directly from the `sv1` directory, and so has identical functionality, the output files from running `select_sv_targets` should be the same as those created by running without the `--iteration` flag:
```
drdir=/global/cscratch1/sd/adamyers/dr9j
select_sv_targets $drdir/south/sweep $CSCRATCH/blat -s2 $drdir/north/sweep --nosecondary --nobackup
```
- Finally, you can read in the files and check that `main_cmx_or_sv` works with the new files, e.g.:
```
import fitsio
from desitarget.targets import main_cmx_or_sv
objs = fitsio.read("/global/cscratch1/sd/adamyers/blat/targets/sv2/resolve/dark/sv2-targets-dr9-hp-X.fits")
columns, masks, survey = main_cmx_or_sv(objs)
```

PR #626, opened 2020-06-30
avoid absolute path in resource_filename

sbailey:
This PR fixes a warning about deprecated absolute paths in `resource_filename`, which could arise when `load_mask_bits` was called with the default `prefix=""`, which would evaluate to searching for `/data/targetmask.yaml` instead of `data/targetmask.yaml`.


geordie666:
This looks good to me. Feel free to merge it whenever you're ready.

PR #631, opened 2020-07-24
support for py3.8 and np1.18

sbailey:
This PR cleans up some warnings from python 3.8 + numpy 1.18.5:

  * numpy is getting more picky about not using `where` on scalars, thus necessitating
     more verbose `if np.isscalar(x) ... else ...` code.
  * avoid calling `np.median` on empty slices (previously that would return a NaN, so
     still using that if the slice was empty).

Tests pass at NERSC using both the current desi environment with python/3.6 + numpy/1.16.4 and a test python/3.8 + numpy/1.18.5 environment, but I'm not sure if the tests cover the Gaia NaN and QA empty slice cases so please take a look.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,8 @@ env:
         - NUMPY_VERSION=1.16.5
         # - SCIPY_VERSION=0.16
         - ASTROPY_VERSION=2.0.16
+        #- Old healpy version needed as long as we pin old astropy too
+        - HEALPY_VERSION=1.12.9
         # - SPHINX_VERSION=1.6.6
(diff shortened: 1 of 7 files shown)


sbailey:
Thanks @geordie666 .  In this case I do want to leave the "warnings as errors" for the tests, since I think if we accidentally do re-introduce this warning we want to fix it since it is triggered by a deprecated "feature" that will change in the future.  We can turn it back off if we hit a legitimate case where that "nonzero on 0d arrays" warning is triggered that we don't want to work around (I'm not sure what that case would be...)

PR #634, opened 2020-07-30
support mock QA with fitsio>1.0 (bytes vs. str)

### LINKED ISSUE #632: mock QA failing with fitsio 1.x
mock QA is failing with fitsio 1.x due to `QA.py` [line 413](https://github.com/desihub/desitarget/blob/e17821a5f435262371ff7907e34f57caf015d3d1/py/desitarget/QA.py#L413):
```python
oo = objtype.decode('utf-8').strip().upper()
```
because now `objtype` is a `numpy.str_` instead of `numpy.bytes_` object.

One potential fix is straightforward:
```python
try:
    oo = objtype.decode('utf-8').strip().upper()
except:
    oo = objtype

extname = 'TRUTH_{}'.format(oo)
```

This highlights a piece of code that is used by the integration tests but not covered by the unit tests.  In addition to the above fix, I suggest adding a lightweight truth.fits file to the test data directory and use that to at least verify that `desitarget.QA.read_data(targfile, mocks=True)` works, and perhaps even more.

But before launching into adding new binary files and mixing mocks and real data testing, I'm pausing for comment.  @geordie666 @moustakas ?

sbailey:
This PR fixes #632 so that mock QA works with fitsio > 1.0 by adding extra bytes vs. str checks.  It adds some trimmed down mock targets+truth test data and uses that to test mock QA.  I have tested with with both fitsio/0.9.11 and fitsio/1.1.2 .

As a cleanup measure, this PR also wraps all `test/make_test*.py` code in `if __name__ == "__main__":` blocks so that importing the code (e.g. by some documentation scraper) doesn't accidentally trigger trying to regenerate the test datasets.


geordie666:
I'll merge this tomorrow morning if you don't beat me to it.

sbailey:
Thanks for the additional checks.  Merging now.

FWIW, the code previously passed unit tests, and now I'm going through the integration test and for anything that is failing there, I'm going back and making sure that there is a unit test that would have caught the problem.  I think I am done with desitarget, but am now working on other repos.

PR #635, opened 2020-08-06
Refactor MTL to implement ledgers

geordie666:
This PR is a major refactor of MTL to implement ledgers. It includes:
- Code to make a set of initial HEALPixel-split ledger files from the standard target files.
   - An example of the output is at `/global/cscratch1/sd/adamyers/egledger`.
   - Functionality to produce ledgers for each set of observing conditions (`"BRIGHT"`, `"GRAY"`, `"DARK"`, etc.).
   - An easy-to-use binary executable script.
   - A new data model that includes a state for each target, a timestamp for when a target's state was updated, and the `desitarget` code version that was used to update the target's state.
- Code to rapidly update the MTL information for a set of targets by appending to a ledger.
    - An example of the output is at `/global/cscratch1/sd/adamyers/egupdates`.
    - Requires a set of targets to update and a `zcat`...these currently have the standard set of columns (i.e. the required columns are unchanged by this PR).
- Functionality that works (throughout) with either `FITS` or `ECSV` files (although we'll need to choose one or the other).
- Columns are now automatically trimmed to a minimum necessary set in `desitarget.mtl.make_mtl()`.
    - This should save factors of 4-5 in memory, which may help with the processing of the mocks.
- A function `desitarget.mtl.inflate_ledger()` that can be used to re-obtain (any subset of) the trimmed target columns.
- A function to write MTL ledgers in a standard format.
- Utility functions to read (FITS or ECSV) MTL ledgers, including:
    - In a set of HEALPixels (`desitarget.io.read_mtl_in_hp()` or `desitarget.io.read_targets_in_hp()` with `mtl=True`)
    - In a set of tiles (`desitarget.io.read_targets_in_tiles()` with `mtl=True`)
    - In a box (`desitarget.io.read_targets_in_box()` with `mtl=True`)
    - In a cap (`desitarget.io.read_targets_in_cap()` with `mtl=True`)
- All of the "read" functions can read the entire ledger, or just the most recent entry for each ``TARGETID``.


geordie666:
To fully implement ledgers will likely also require changes downstream (to `fiberassign`, and maybe even how the `zcat` is produced). So, here are some additional notes on how to use this new functionality, and the overall workflow, to illustrate the underlying philosophy:

- Use `mtl.make_ledger()` or `/bin/make_initial_mtl_ledger` to make the ledger. See the example at `/global/cscratch1/sd/adamyers/egledger`.
    - The parallelized version of `mtl.make_ledger()` makes an initial (HEALPixel-split) ledger in about 30 minutes.
- Use `mtl.update_ledger()` to update the ledger. See the example at `/global/cscratch1/sd/adamyers/egupdates`. 
    - Currently, this takes `targets` and a `zcat` to inform the updates as was the case before this PR. *However*, really all that is needed is a `zcat` that includes `"RA"`, `"DEC"` and "`TARGETID`". If those columns were present in the zcat, then the `targets` input wouldn't be needed, as the full state could be looked up in the existing MTL ledger.
    - Updating 250 tiles-worth of targets (so, say, about a month of observations) takes less than a minute for ECSV files (which are quick to append to).
- When planning the next set of tiles (say, for the subsequent month), read the targets back in from the ledger file using `desitarget.io.read_targets_in_tiles()` (with `mtl=True`).
- Where needed downstream, or for analysis, grab any missing target columns using `desitarget.mtl.inflate_ledger()`.

geordie666:
One note: This doesn't yet implement a separate ledger for secondary targets. Or, at least, that aspect has not been extensively tested. Secondary targets will need an additional PR.

sbailey:
@geordie666 and I walked through this in some detail.  Please update `make_mtl(trimcols=True)` to be False by default for backwards compatibility (at least for now).  Otherwise it looks like this adds new functionality for the ledgers, but doesn't otherwise break how fiberassign and integration tests would use the mtl, so ok to merge.

We can then update fiberassign to use desitarget.io functions to read the ledgered MTLs, but that can happen after merging this.

Other notes from my conversation with Adam:
  * If we ad RA,DEC into the zcat, then `update_ledger` may not need the targets anymore
   (i.e. by trusting that the pre-existing ledgers are right).
  * Although updating the MTL should now be much faster (appending rows of what changed instead of rewriting the whole thing), the big-FITS to N>>1 small-ASCII may have made the fiberassign I/O much slower when replaying fiberassign on the whole survey.  That doesn't matter for operations, but could matter for mocks.

moustakas:
I apologize I don't have the cycles at the moment to provide useful (or not useful!) feedback, but I'm just wondering if this impacts `select_mock_targets` in any way? I don't think so but I wanted to get a sense of whether any critical mock updates were going to be needed.

geordie666:
@moustakas: In the short-term, I don't think so either. I've deliberately not made any critical changes in `make_mtl()`, which, I believe, is the only relevant function for the mocks. If I've somehow broken that it should be a quick fix.

In the longer-term, yes, I think we'd need refactoring of some of the infrastructure for the mocks depending on how (or even if) we decide to adopt a ledger-based scheme.

geordie666:
@forero: Before I merge this PR, I'd appreciate feedback from you regarding one point for which I'm unsure if my changes to `mtl.make_mtl()` could break the end-to-end simulations.

I've introduced a `NUMOBS` column in the _output_ MTL from `mtl.make_mtl()` which seemed like a column that we'd want to record in an informative ledger. However I remember that in #405 you encountered problems when `NUMOBS` was an _input_ to `mtl.make_mtl()`. Is it a problem that `mtl.make_mtl()` now includes `NUMOBS` (and `Z` and `ZWARN`) as outputs? 

If this _is_ a problem, I can implement a short-term fix before merging this PR by including a keyword in `mtl.make_mtl()` that doesn't return the newly added `NUMOBS` column (and `Z` and `ZWARN`).

forero:
Hi @geordie666. I am running a set of small tests. So far I've only got a warning `mtl.py:348: SyntaxWarning: "is" with a literal. Did you mean "=="?
  ecsv = get_mtl_ledger_format() is "ecsv"`.

geordie666:
Thanks, @forero: That warning, I think, is tied to more careful checking of singletons that came as part of the upgrade to Python 3.8. Sloppy coding on my part. I'll fix it before I merge. 

I appreciate you stress-testing this PR. Let me know if you find anything else today and, if not, I'll merge later tonight.

forero:
@geordie666 I didn't find anything weird. The end-to-end tests went fine and the results look perfect.

Issue #633, opened 2020-07-29
Update filenames for targets to make them more unique


geordie666:
Currently, targets (and the associated directory structure) look like:

`target/catalogs/dr8/0.39.0/targets/sv1/resolve/bright/sv1-targets-dr8-hp-21.fits`

To make the filenames unique but free of extraneous information, make the following updates in `io.find_target_files()`:

- Remove the trailing hyphen between sv1 and targets in the filename (`sv1targets` instead of `sv1-targets`).
- Remove the imaging release in the filename completely (`targets-` instead of `targets-dr8`).
   * Should probably also add a DR keyword to the header, to ensure that information is recorded somewhere in the file.
- Main Survey filenames should (still) look like `targets-` without any trailing `main`.
- Add the observing "layer" (`bright`/`dark`) to the filenames.
- Add `noresolve` to the filenames where needed for the rare cases that we release targets that are not resolved across the "North"/"South" imaging.

Some example results should look like:

```
target/catalogs/dr8/0.41.0/targets/main/resolve/dark/targets-dark-hp-33.fits
target/catalogs/dr8/0.41.0/targets/main/resolve/bright/targets-bright-hp-33.fits
target/catalogs/dr8/0.41.0/targets/sv1/resolve/dark/sv1targets-dark-hp-33.fits
target/catalogs/dr8/0.41.0/targets/main/noresolve/bright/targets-dark-noresolve-hp-33.fits
```

sbailey:
Thanks.  Expanding on this point more generally:

> Should probably also add a DR keyword to the header, to ensure that information is recorded somewhere in the file

Indeed.  As a general principle throughout all our code and files, any metadata that is encoded in the directory structure and filename should also be present in the header keywords.  i.e. the directories and filenames are useful for finding the files and keeping ourselves organized, but the contents of the file should also be sufficient to re-derive the associated metadata to recreate the filepath and filename.

geordie666:
Addressed in #637.

PR #637, opened 2020-08-14
Update the data directory structure and filenames

geordie666:
This PR updates the data model to address #633. The filenames and directory structure for targets are now aligned with the discussion in #633 and the updates in [desidatamodel #85](https://github.com/desihub/desidatamodel/pull/85).

The data model for other target classes (_skies, gfas, randoms, masks, mtls_) has also been updated to better reflect the directory structure and filenames that are now used for targets. In addition, a few informative keywords have been added to FITS headers, in particular, the Gaia Data Release has been added for supplemental targets and skies.


geordie666:
This is a relatively simple PR that should be uncontroversial. I'll merge it over the weekend or on Monday morning.

Issue #356, opened 2018-08-29
add units to target catalog table


moustakas:
None of the columns in the output targeting catalog have units.  Do we care?
```python
import fitsio
from astropy.table import Table
tt = Table(fitsio.FITS('/global/projecta/projectdirs/desi/target/catalogs/dr7.1/0.23.0/targets-dr7.1-0.23.0.fits')[1].read(rows=0))
for col in tt.colnames:
    if tt[col].unit is not None:
        print('Column {} has units!'.format(col))
```

sbailey:
Ideally we would include units in the fits files themselves, as well as in the data model.  Low priority.

geordie666:
Addressed in #638.

PR #638, opened 2020-09-09
Add units to all output files

geordie666:
This PR adds units to all output files, where "all" means targets, mtl ledgers, mask files, GFAs, skies and randoms. New functionality includes:

- Units for all output quantities are stored in a new yaml file named `data/units.yaml`.
- Unit tests check that any new output quantities that are added to the code have associated units.
- Tests also check that all units are valid astropy units using `astropy.units.Unit()`.

This should address #356.

This PR also includes some more minor cleanup and speedups, such as:
- Fixes some Future/Deprecation Warnings associated with environment upgrades.
- Uses the `io.find_target_files()` functionality to write output files in parallelization scripts for randoms.
- Ensures randoms parallelized across HEALPixels on different nodes are re-randomized before they are combined.
- Smarter traversing of imaging directories when searching for sweeps and Tractor files
    * The imaging directories can contain coadd/log/metrics/etc. files that could slow down an `os.walk()`

### DIFF
--- a/bin/gather_targets
+++ b/bin/gather_targets
@@ -32,6 +32,8 @@ ap.add_argument("--norandomize", action='store_true',
 ap.add_argument("--numchunks", type=int,
                 help='number of chunks in which to write to save memory [defaults to {}]'.format(nchunks),
                 default=nchunks)
+ap.add_argument("--skip", action='store_true',
+                help="Skip input files that don't exist without flagging an error")
 
 ns = ap.parse_args()
 
@@ -42,7 +44,11 @@ if ns.targtype == 'gfas':
 extname = tt.upper()
 
 # ADM convert passed csv strings to lists.
-fns = [ fn for fn in ns.infiles.split(';') ]
+if ns.skip:
+    fns = [ fn for fn in ns.infiles.split(';') if
+            os.path.isfile(os.path.expandvars(fn))]
+else:
+    fns = [ fn for fn in ns.infiles.split(';') ]
 
 # ADM read the header from the first file.
 fx = fitsio.FITS(fns[0])
--- a/bin/gather_targets
+++ b/bin/gather_targets
@@ -2,6 +2,7 @@
 
 from __future__ import print_function, division
 
+import os
 import fitsio
 from fitsio import FITS
 import numpy as np
@@ -15,6 +16,9 @@ start = time()
 from desiutil.log import get_logger
 log = get_logger()
 
+# ADM the default number of chunks to write in to save memory.
+nchunks = 10
+
 from argparse import ArgumentParser
 ap = ArgumentParser(description='Concatenate multiple FITS files into one large file. Retains the header from the FIRST listed input file')
 ap.add_argument("infiles", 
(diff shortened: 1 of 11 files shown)


geordie666:
Before I merge this PR I'm going to run an end-to-end check that the randoms write and are randomized correctly. So, I'll likely merge this tomorrow night.

sbailey:
This sounds great.  Could you point us to some target files generated with this version of the code that have all units?  Thanks.

sbailey:
Unit tests fail on "nanomaggy", which I thought was spelled "nanomaggie" (per [SDSS glossary](https://www.sdss.org/dr12/help/glossary/)), and is a case where I'm willing to grant an exception to normal IAU/astropy rules on valid units.

geordie666:
Amusingly, the unit test failure is because we are still testing against `Python 3.6` but our standard environment at NERSC is now `Python 3.8`. Sourcing `desiconda 20.4` triggers this error:
```
import astropy.units as u
u.Unit('nanomaggy')
ValueError: 'nanomaggy' did not parse as unit: At col 0, nanomaggy is not a valid unit.
```
But sourcing `desiconda 20.7` produces:
```
import astropy.units as u                                                                                                                                                                          
u.Unit('nanomaggy')                                                                                                                                                                                
Out[]: Unit("nmgy")
```
(nanomaggi**e** triggers a `ValueError` in either case!)

I'll add a workaround to the unit test, for now, although its probably time to update the `.travis.yml` file.

There's an example target file with units at:
```
/global/cscratch1/sd/adamyers/PR638/dr8/0.42.0.dev4183/targets/main/resolve/dark/targets-dark-hp-1.fits 
```

sbailey:
I think the actual relevant difference is astropy 2.x vs. 4.x rather than python 3.6 vs. 3.8.  Please update .travis.yml to use astropy 4.x and then workaround unit tests as needed from there.  OK to drop astropy 2.x support now.

It does appear that SDSS is inconsistent in "nanomaggy" (e.g. [here](https://www.sdss.org/dr12/algorithms/magnitudes/)) vs. "nanomaggie" (e.g. [here](https://www.sdss.org/dr12/help/glossary/)), though I also note that the Legacy Surveys use "nanomaggie" in the docs.  I have not tried to dig into the history further for the original usage (it's not in Oke & Gunn 1982...)

geordie666:
I'd prefer to merge this and update the `.travis.yml` file in a separate, dedicated PR. The unit tests now pass, thanks to an explicit check on `astropy` version (which will allow us to maintain backward compatibility for `astropy 2.x` until we decide to completely deprecate our `Python 3.6/astropy 2.x`conda environment).

sbailey:
OK to merge since you have it working, though I do consider the python 3.6 / astropy 2.x environment to be deprecated for any new versions of our code.  i.e. the previous releases are still kept around for continuity, but we don't need new versions of desitarget+ to work with the old environment.  e.g. for testing a new version of desitarget, please use software release 20.8 or current and then swap out to your clone of desitarget.

geordie666:
I added an issue to ensure that `.travis.yml` gets updated. For absolute clarity: all `desitarget` unit tests currently pass using software release `20.8` at NERSC. It's just the Travis environment that has fallen behind.

Issue #639, opened 2020-09-10
Update .travis.yml to better resemble desiconda release 20.8


geordie666:
Unit tests still use a `Python 3.6` environment that approximates `desiconda 20.4`. Update Travis tests to a `Python 3.8` environment that better resembles `20.8`. See, e.g., https://github.com/desihub/desitarget/pull/638#issuecomment-689902765 for additional context.

geordie666:
Addressed in #640.

PR #640, opened 2020-09-16
Update Travis for the Python 3.8/Astropy 4.x combination

geordie666:
This PR updates the online Travis tests to better resemble the `desiconda 20.8` release, which addresses #639.

Basically, we switch the `.travis.yml` file to install `Python 3.8`, `Numpy 1.19.1` and `Astropy 4.0.1` for tests. This also facilitates a few more updates, such as using `healpy 1.14.0` and `pip 20.2.2`.

Although the key tests are updated for the `Python 3.8`/`Astropy 4.0.1` combination, the "stable" test (which is an allowed failure) and the sphinx build still use `Python 3.6`. It didn't seem straightforward to update these tests to use `Python 3.8` (yet), and it isn't as critical that these tests align exactly with the environment at NERSC.

This PR also brought along a useful script that can generate the QSO Random Forest probabilities for a target (or sweeps) file for post-facto analysis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,8 +89,8 @@ env:
     matrix:
         #  - PYTHON_VERSION=3.5 NUMPY_VERSION=1.15.4 SETUP_CMD='egg_info'
         #  - PYTHON_VERSION=3.5 NUMPY_VERSION=1.15.4 SETUP_CMD='bdist_egg'
-        - PYTHON_VERSION=3.6 SETUP_CMD='egg_info'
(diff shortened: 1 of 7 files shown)


sbailey (approved):
Looks good, thanks.  For the record, regarding some tests on py3.6 and others on py3.8, I view it as a feature rather than a problem when the Travis tests have somewhat different versions of numpy etc. than NERSC so we get additional coverage since people's laptops may also have significantly newer or older versions than what is current at NERSC.  Good to update tests to fitsio/1.x and astropy/4.x though since we no longer support fitsio/0.x and astropy/2.x.

Issue #20, opened 2015-12-11
improve select_targets traceability


sbailey:
Improve the target selection traceability by tracking the complete list of input files used and their checksums.

The output of select_targets includes the desitarget version, the git revision (if running from a git checkout for testing; otherwise "unknown", though perhaps that should be the tag), and the input directory of files used, e.g.

```
DEPNAM00= 'desitarget'                                                          
DEPVER00= '0.2.0   '           / desitarget.__version__                         
DEPNAM01= 'desitarget-git'                                                      
DEPVER01= 'unknown '           / git revision                                   
DEPNAM02= 'tractor-files'                                                       
DEPVER02= '/project/projectdirs/desiproc/dr2/sweep/'                            
```

but it doesn't actually trace what files it found under /project/projectdirs/desiproc/dr2/sweep/ or what their checksums were (to be able to check whether someone mucked with the files later, e.g. since this isn't the real final locked down publicly released DR2 directory).

Add another bintable HDU that tracks the actual files used as input and their checksums.  In the case of tractor files as inputs instead of sweeps this could be a long list so it requires a table and not just DEPNAM/DEPVER header keywords.  Don't include the common path prefix that is already kept as DEPVER02.

geordie666:
Addressed in #641.

PR #641, opened 2020-09-23
Add table of input sweeps files and their checksums to the output target files

geordie666:
This PR adds a table of input sweeps files and their associated `SHA-256` checksums as a second extension (named `"INFILES"`) to files created by `select_targets`. Features include:

- `select_targets` takes a flag `-noc` (or `--nochecksum`) to turn off this new behavior, if desired.
- Only sweeps files actually considered by `desitarget.cuts.select_targets()` are included in the second extension table, to speed the calculation of checksums when the creation of target files is limited to certain HEALPixels or RA/Dec ranges.

This addresses #20, which has been dangling for a while (!) although I decided to write the full file paths (in the words of #20, I decided to "include the common path prefix that is already kept as `DEPVER02`") for convenience.


weaverba137:
Is the intention to compute the checksums for files other than the DR9 sweep files? Those will already have checksums as part of the release process. For that matter, I don't see the functionality that takes advantage of the pre-computed checksums.

geordie666:
The intention is to address issue #20. @sbailey can comment on whether issue #20 should just have been closed rather than left dangling. I'm certainly content to simply close issue #20 and close this PR without merging it.

I'm not sure I understand your point about the pre-computed checksums? My reading of issue #20 is that it implies that `desitarget` should compute the checksums itself for cases where input sweep files are not part of an official release (for instance, as we speak, I'm running targets for `dr9"m"` again).

weaverba137:
OK, after reading #20, that may pre-date even the *concept* that sweep files for regular releases would have pre-computed checksums. So `desitarget` *does* use pre-computed checksums when the sweep files *are* part of an official release? Or do we not actually care about the provenance when targeting on an official release? @sbailey could you provide some context here?

Another way of thinking about it: presumably you always want to have the same data model, so presumably you will always include the INFILES HDU. What is the advantage to recomputing the checksums when they are available in a file?

geordie666:
Thanks, @weaverba137, I think those are valid points. To stress, again, I have absolutely no problem with closing this PR without merging my work if we decide that is the best way forward. I tackled #20 over the last couple of days as I knew I would be too busy in the first part of this week to handle anything more substantive!

sbailey:
**Short version**: I'm fine with this PR, including re-calculating the checksums, as long as
  1. the flavor of checksum calculated here is supposed to match the flavor of checksum pre-calculated as part of a release
  2. recalculating the checksums doesn't significantly slow down target selection

**Further thoughts**: 

The main spirit of #20 was to provide full provenance traceability of the inputs used to generate the targeting outputs, so regardless of checksums this PR seems to improve that situation by including the list of specific files and not just a directory under which some unspecified files were found.

Even if sweeps have pre-calculated checksums, re-calculating them may be a feature not a bug, since it is getting the *checksum of the file that was actually used*.  Belt-and-suspenders would compare that to the pre-calculated checksum (if it exists) and complain if they disagree so that we don't blindly target select on a corrupted file.

Issue #20 was motivated by target selection runs on non-final data releases.  Hopefully that won't occur for final target selection that gets used on-sky, but is seems like it could still happen for R&D runs of target selection on a nearly-final-but-not-quite DR9 (or DR10...) where we want provenance of what was used even if we won't observe with those targets.  Is this is real possibility anymore?

Even if we only use fully released final sweeps forevermore, does this filepath+checksum concept still apply for secondary target lists that could be evolving, or is that a separate issue from what is solved here?

geordie666:
To answer some of Stephen's questions:

- This slows down the target selection code by about 10 minutes per node for a typical use-case of generating targets over `nside=2` HEALPixels at one HEALPixel per node.
- Anything regarding secondary targets is a separate issue. The code that generates a secondary target list from files in the `$SCND_DIR` is distinct from `select_targets`.

I'll await thoughts from @weaverba137 before merging.

weaverba137:
OK, I agree with @sbailey that if we're going to absorb the hit of computing checksums anyway, we might as well compare them to pre-existing ones rather than taking them as given.  The comparison should be much faster than the computation, so it wouldn't add that much more time.

geordie666:
@weaverba137: when you say you agree with @sbailey do you mean that you're also (to quote @sbailey's short version): _fine with this PR_ and I should merge it?

I guess one question is whether the flavor of checksums I'm calculating (SHA-256) is the correct choice. Although, it's an easy fix to one line of code if we change the checksum flavor.

weaverba137:
I'm OK with calculating the checksums *and* comparing them to pre-existing checksums to ensure the listed files are uncorrupted. SHA256 is the right choice.

geordie666:
I don't _have_ "pre-existing" checksums for experimental runs (DR9x, perhaps DR10 etc.) as the imaging is evolving. That, I think, was the spirit of #20 (calculating the checksums on the fly because they don't exist for experimental runs).

So, I think I need clarification on what "pre-existing" checksums you think desitarget should be comparing to? Or, do you mean that downstream other aspects of the workflow should check against checksums generated by desitarget?

Or, do you mean that in cases where we have an "official" release, some code in the end-to-end integration checks that any official checksums for final releases agree with the checksums that have been written into targeting files by desitarget?

weaverba137:
OK, I must be missing something obvious.  Let me try again to make myself clear.

* All official releases have checksum files.  For example, `cosmo/data/legacysurvey/dr8/north/sweep/8.0/legacysurvey_dr8_north_sweep_8.0.sha256sum`.
* If a targeting file is loaded compute its checksum, using SHA256, regardless of origin.
* If the file *happens* to be from an official release, also read in the file, *e.g.* `cosmo/data/legacysurvey/dr8/north/sweep/8.0/legacysurvey_dr8_north_sweep_8.0.sha256sum` and make sure the computed checksum matches the official checksum.
* If they match, great!  If they don't match, fail, or issue a strong warning.

That's what I understood from @sbailey's comment:
> Even if sweeps have pre-calculated checksums, re-calculating them may be a feature not a bug, since it is getting the checksum of the file that was actually used. Belt-and-suspenders would compare that to the pre-calculated checksum (if it exists) and complain if they disagree so that we don't blindly target select on a corrupted file.

If that is still not clear, maybe we should chat on Slack.

geordie666:
No need to chat. It's clear with your clarification. I was missing the `if` statement (namely, only check `if` there is an existing `.sha256sum` file).

I'll add this extra check when I have time to get back to this PR.

weaverba137:
Sounds good, thanx!

geordie666:
@weaverba137: I've added the code that automatically checks for existing shasum files and raises an exception if there is a mismatch in the `SHA256` checksum.

I'll probably merge this PR in about a day, as I'm going to need a new tag of the code to run some new commissioning files. But, certainly, let me know if you have further comments.

py/desitarget/io.py
```diff
@@ -1772,6 +1789,106 @@ def load_pixweight_recarray(inmapfile, nside, pixmap=None):
...
+    # ADM in case a single string was passed.
+    infiles = np.atleast_1d(infiles)
+
+    # ADM we'll first populate a dictionary with the checksums.
+    shadict = {}
+    nf = len(infiles)
+    # ADM if verbose is True, write out info for 20 blocks of files.
+    block = nf // 20 if nf // 20 else 1
+    for ifn, infile in enumerate(infiles):
+        p = Popen(['sha256sum', infile], stdout=PIPE, stderr=PIPE)
```
weaverba137, line 1824: My only concern is that you are using the external checksum command instead of Python's built-in `hashlib`, https://docs.python.org/3/library/hashlib.html.  Is there a performance reason for that?  This also assumes that `sha256sum` is always in `PATH`, which might cause problems testing on MacOS, where `sha256sum` isn't installed by default.

weaverba137 (commented):
See comment on `hashlib`.

py/desitarget/io.py
geordie666: I had no particular reason for using the external command, other than that it hadn't occurred to me that `sha256sum` might not be available on Mac OS X. Good point.

I'll happily switch, although this PR will have to go on the back-burner until after I've finished targeting for recommissioning.

geordie666: I added a function called `get_sha256sum()` that uses Python's built-in `hashlib` to generate the checksums instead of relying on the OS.

geordie666:
@weaverba137: I've added code to use Python's built-in `hashlib` to generate checksums instead of relying on the OS. Let me know if you think this is ready to merge, now (assuming tests pass).

weaverba137:
This looks good to me.  Merge when ready.

PR #642, opened 2020-09-24
Ledgers example notebook 

michaelJwilson:
Jupyter notebook for using and abusing ledgers.

### DIFF
--- a/doc/nb/Ledgers.ipynb
+++ b/doc/nb/Ledgers.ipynb
@@ -2,10 +2,13 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 32,
+   "execution_count": 22,
    "metadata": {},
    "outputs": [],
    "source": [
+    "import glob\n",
+    "import numpy as np\n",
+    "\n",
     "from astropy.table  import Table\n",
     "\n",
     "from desitarget.io  import read_mtl_ledger\n",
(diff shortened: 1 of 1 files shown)


michaelJwilson:
Potential questions (unless I realise the answer):

--  How do we know the update directory list?  Is there only one such directory?

geordie666:
Thanks for turning my terse line-by-line instructions into a far more informative notebook, @michaelJwilson.

In answer to your question about whether there is only one directory for the ledgers: Each "flavor" of program in the datamodel at, e.g.: 

https://desidatamodel.readthedocs.io/en/latest/DESI_TARGET/TARG_DIR/DR/VERSION/TYPE/PHASE/RESOLVE/OBSCON

would have a single corresponding ledger. So, for example, one ledger for the `OBSCON=bright` program during `PHASE=SV1` and one ledger for the `OBSCON=dark` program in, e.g., `PHASE=main`.

I haven't written up the full data model for MTL ledgers and posted it to the `desidatamodel`, yet, but there would be a `TYPE=mtl` in the above directory tree.

The structure for each of these directories is generated automatically by `desitarget.io.find_target_files()`:

https://github.com/desihub/desitarget/blob/dafb7c4e1e06adc40d493f3f7e6c610ee31eb07d/py/desitarget/io.py#L1989-L1992

where the syntax is slightly different to that of the data model (`PHASE <-> survey` and `TYPE<->flavor`).

michaelJwilson:
No problem.  It's not quite what I was expecting so I'll hold off from further questions / a pull request until I acclimatize.   Thanks!

michaelJwilson:
I hope to close this out today.  Is it worth moving the example ledger to cfs/desi/target/.  Anything you think should be added?

geordie666:
It's fine to close this out. I think that `/cfs/` should be reserved for final work, not examples, so I'd propose leaving the example ledger in `$SCRATCH` for now, and then updating the notebook once we have an _actual_ ledger in `/cfs/` rather than just an example.

michaelJwilson:
ok, no worries.  I'm curious about TARGET_STATE.  I'm not going to pester you about data model questions until you write the docs, but the ELG, LRG, QSO inclusion suggests a tie to DESI_TARGET.  In which case, SV1_TARGET_STATE to mirror SV1_DESI_TARGET?

geordie666:
Specifically, I think we chose things like `SV1_DESI_TARGET` to indicate which bit-mask to look up in the event that you open a file and aren't 100% sure which bit-mask you were working with. 

I'm not sure we also need it for `TARGET_STATE`. Although it's an excellent point, as technically the `MTL` behavior is linked to the survey "flavor". Let me think about it. Such a change leads to more complex machinery, but it is a good idea. I'll open an issue to remind myself

michaelJwilson:
OK.  I'd also have expected RELEASE in the slimmed mtl, at least to facilitate an assert on inflate_ledger.  I think this will happily mismatch targetids across releases as is?

Finally, I'd also have expected NUMOBS >= 1 for the files under egupdates, which doesn't seem to be the case.  Maybe this is less realworld than that.

In any case, switching this from a draft.  Happy to add more, or merge if happy.

geordie666:
Note that `RELEASE` is embedded in `TARGETID` and can be retrieved using `desitarget.targets.decode_targetid()`:

https://github.com/desihub/desitarget/blob/f355f142ae8085776f7c9c21b0a7565284e32372/py/desitarget/targets.py#L133

I'll merge this when tests pass (which could be a while, as Travis has been slow lately).

michaelJwilson:
ok, thanks.  If I had known that, I'd forgotten! (Travis looks to be done).

PR #649, opened 2020-11-12
Clean-up targets and randoms for the internal DR9 release.

geordie666:
Includes:
- A function `geomask.imaging_mask()` that allows easier parsing of maskbits cuts by using a list of strings (such as `["BRIGHT", "CLUSTER"]` etc.
- A default set of cuts on maskbits as a kwarg in the new `geomask.imaging_mask()` function.
- A new executable `alt_split_randoms` that is slower than `split_randoms` but saves memory, so is useful for large files.
- Increased flexibility for adding MTL columns to randoms, which saves memory when making large random catalogs:
    * MTL columns can still be added directly when running the randoms using `select_randoms` (but now must be requested with an `--add_mtl` command-line flag).
    * MTL columns can now _alternatively_ be added when splitting a larger random catalog (again by adding an `--add_mtl` command-line flag to `split_randoms` or `alt_split_randoms`).

### DIFF
--- a/bin/alt_split_randoms
+++ b/bin/alt_split_randoms
@@ -1,0 +1,61 @@
+#!/usr/bin/env python
+
+import os, sys
+import numpy as np
+import fitsio
+from time import time
+start = time()
+import fitsio
+
+from desiutil.log import get_logger
+log = get_logger()
+
+from argparse import ArgumentParser
+ap = ArgumentParser(description='Split a random catalog into N smaller catalogs. Shuffle the random catalog first to ensure randomness.')
+ap.add_argument("randomcat",
+                help='A random catalog (e.g /project/projectdirs/desi/target/catalogs/randoms-dr4-0.20.0.fits). For an input catalog /X/X.fits N smaller catalogs will be written to /X/X-[0:N-1].fits')
+ap.add_argument("-n", "--nchunks", type=int,
+                help='Number of smaller catalog to split the random catalog into. Defaults to [10].',
+                default="10")
+
+ns = ap.parse_args()
+
+if not os.path.exists(ns.randomcat):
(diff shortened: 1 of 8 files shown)


geordie666:
@djschlegel and @moustakas: Could you take a look at the `maskbits` masking cuts and check that I have the logic that you requested correct for the DR9 defaults?

The overall default, which is applied to the ELGs, LRGs and QSOs in both SV and the Main Survey is here:

https://github.com/desihub/desitarget/blob/d340aa67ae1d551b885d3ba78af426a2eeda136a/py/desitarget/geomask.py#L35

and its pretty easy, now, to just read the default from the new `imaging_mask()` function (it's `bitnamelist`). 

The logic applied to the BGS targets is in the following places:

https://github.com/desihub/desitarget/blob/d340aa67ae1d551b885d3ba78af426a2eeda136a/py/desitarget/cuts.py#L1039
https://github.com/desihub/desitarget/blob/d340aa67ae1d551b885d3ba78af426a2eeda136a/py/desitarget/cuts.py#L1131
https://github.com/desihub/desitarget/blob/d340aa67ae1d551b885d3ba78af426a2eeda136a/py/desitarget/sv1/sv1_cuts.py#L1033

I have removed any cuts on `ALLMASK_G/ALLMASK_R/ALLMASK_Z` completely.

Once you sign off on these changes, I can tag a new version of `desitarget` and run the targets and randoms for the "internal" DR9 release. Assuming that Travis isn't very slow with checks during the tagging process, I may well then have time to make target files by late tomorrow afternoon.

I didn't quite have time to also fix the bug in the skies, but I'll leave that for later work.

geordie666:
@djschlegel confirms that the logic looks correct. So, I'm merging this. I'll tag through the morning and hopefully produce targets this afternoon.

Issue #647, opened 2020-11-10
Duplicate TARGETID in skysupp


forero:
Using `/global/cfs/projectdirs/desi/target/catalogs/gaiadr2/0.43.0/skies-supp/` I found the duplicate `TARGETID` 2882304691935379461. That duplicate TARGETID made `fba_run` crash on cmx tests I was running for `TILEID` 5009.

forero:
Anand in an email:

so I gave a check, I find that:
- there are no duplicates in skies-hp-*.fits
- there are no duplicates between skies-hp-*.fits and skies-supp-hp-*.fits
perfect.

but there are some duplicates inside the skies-supp-hp-*.fits:
- skies-supp-hp-0.fits has 1291 duplicates
- there are some duplicates between the skies-supp-hp-*.fits files (if I
got it right, overall, 4594 out of 354,064,078, with counting the 1291
duplicates from skies-supp-hp-0.fits)

to my understanding, that shouldn t happen, right? or is it expected?

thanks!

anand

ps: I m checking:
/global/cfs/projectdirs/desi/target/catalogs/dr8/0.43.0/skies/skies-hp-*.fits
/global/cfs/projectdirs/desi/target/catalogs/gaiadr2/0.43.0/skies-supp/skies-supp-hp-*.fits

geordie666:
Thanks for the digging @forero (via Anand!). That's definitely a bug. I'll try to look at it later today.

geordie666:
Addressed in #653.

PR #653, opened 2020-11-21
Prepare for making commissioning targets and skies with DR9

geordie666:
This PR cleans up the targets and skies with a view to making files for commissioning using DR9 imaging. It also continues a refactor of the randoms that allows large random catalogs to be produced without memory issues. It includes:
- Use HEALPixels instead of `BRICKID`s for "supp" skies, which avoids "supp" skies having duplicated `TARGETID`s in regions where bricks span multiple pixels (this addresses #647).
- Update the faint-end cut for `STD_DITHER_GAIA` cmx targets near the Galaxy to G < 19.
- Allow `gather_targets` to restrict to just a subset of columns when concatenating FITS files.
- Ignore the new "light-curve" and "extra" flavors of sweeps when searching for input sweep files.
- Smarter processing of the randoms when writing and executing the "bundled" slurm file.
  * The new process is to break the HEALPixel-split files into `N` smaller files first and _then_ to combine across the different pixels to make `N` random catalogs.
  * This circumvents the need to make a large random catalog and then split it into smaller random catalogs, which saves a factor of `N` in memory.

### DIFF
--- a/bin/alt_split_randoms
+++ b/bin/alt_split_randoms
@@ -21,12 +21,18 @@ ap.add_argument("-n", "--nchunks", type=int,
                 default="10")
 ap.add_argument("--addmtl", action='store_true',
                 help="If passed, then add the columns needed for MTL to the random catalogs after they are split.")
+ap.add_argument("--skip", action='store_true',
+                help="Check if the input random catalog exists. If it doesn't, do absolutely nothing (useful for scripting across all possible HEALPixels).")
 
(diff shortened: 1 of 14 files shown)


geordie666:
There is some urgency to merge this PR so I can make a new tag and run DR9 targets for commissioning. So, I'll leave this PR open until tomorrow morning and merge it promptly then.

PR #657, opened 2020-11-24
mock_select_target reads Buzzard mocks

zhouconghao:


zchvsre:
It seems the number density is definitely not right with this. Fixing it now.

zchvsre:
`read_mock: Target density = 500.00 targets/deg2 (downsampling factor = 0.009).` 

It seems the downsampling factor is too small.

zchvsre:
I've tested selecting BGSs, LRGs, and ELGs on pixels with the desired density. It seems it works fine.

zchvsre:
Currently, the code read buzzard mocks and output a list of LRG targets. However, the density is higher than desired.

PR #659, opened 2020-12-04
Bgs sv update

qmxp55:
Two main changes for BGS SV selection: 
(1) We've removed the FRACS* cuts. but the LOWQ superset does include rejections by the FRACS*. 
(2) The FIBMAG superset has been shrunk to the limit of 20.5 instead 21.0 in the r-band magnitude.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -16,7 +16,7 @@ desitarget Change Log
 .. _`issue #20`: https://github.com/desihub/desitarget/issues/20
 .. _`PR #641`: https://github.com/desihub/desitarget/pull/641
 .. _`PR #655`: https://github.com/desihub/desitarget/pull/655
-.. _`PR #659`: https://github.com/desihub/desitarget/pull/
+.. _`PR #659`: https://github.com/desihub/desitarget/pull/659
 
 0.45.1 (2020-11-22)
 -------------------
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,6 +5,10 @@ desitarget Change Log
 0.45.2 (unreleased)
 -------------------
 
+* Two main changes for BGS SV selection for DR9 [`PR #659`_]:
+    * We've removed the FRACS* cuts. but the LOWQ superset does include rejections by the FRACS*.
+    * The FIBMAG superset has been shrunk to the limit of 20.5 instead 21.0 in the r-band magnitude.
+
 * Extension of mag limit to 22.3 for RF selection [`PR #655`_].
 * Add input sweep files and their checksums to target files [`PR #641`_].
     * Addresses `issue #20`_.
@@ -12,6 +16,7 @@ desitarget Change Log
 .. _`issue #20`: https://github.com/desihub/desitarget/issues/20
 .. _`PR #641`: https://github.com/desihub/desitarget/pull/641
 .. _`PR #655`: https://github.com/desihub/desitarget/pull/655
+.. _`PR #659`: https://github.com/desihub/desitarget/pull/
 
 0.45.1 (2020-11-22)
 -------------------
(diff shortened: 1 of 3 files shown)


qmxp55:
wiki has been updated https://desi.lbl.gov/trac/wiki/TargetSelectionWG/SurveyValidation

doc/changes.rst
```diff
@@ -5,13 +5,18 @@ desitarget Change Log
...
+    * The FIBMAG superset has been shrunk to the limit of 20.5 instead 21.0 in the r-band magnitude.
+
 * Extension of mag limit to 22.3 for RF selection [`PR #655`_].
 * Add input sweep files and their checksums to target files [`PR #641`_].
     * Addresses `issue #20`_.
     
 .. _`issue #20`: https://github.com/desihub/desitarget/issues/20
 .. _`PR #641`: https://github.com/desihub/desitarget/pull/641
 .. _`PR #655`: https://github.com/desihub/desitarget/pull/655
+.. _`PR #659`: https://github.com/desihub/desitarget/pull/
```
geordie666, line 19: You are missing `659` at the end of the url, here.

py/desitarget/cuts.py
```diff
@@ -1036,7 +1036,7 @@ def notinBGS_mask(gnobs=None, rnobs=None, znobs=None, primary=None,
     bgs &= (gfluxivar > 0) & (rfluxivar > 0) & (zfluxivar > 0)
 
     # ADM geometric masking cuts from the Legacy Surveys.
-    bgs &= imaging_mask(maskbits, ["BRIGHT", "CLUSTER"])
+    bgs &= imaging_mask(maskbits)
```
geordie666, line 1039: Are you sure that you want to default to the standard set of cuts on `MASKBITS`, here? Those are `["BRIGHT", "CLUSTER", "GALAXY"]`. From conversations I'd had with @moustakas I thought that you wouldn't want to mask on `GALAXY` so that you could investigate the genuine galaxies within the SGA masks. My understanding could certainly be outdated, though!

geordie666 (changes requested):
@qmxp55: I have a couple of minor comments. Once you've addressed those, let me know if this is final and I can merge it (once tests pass).

py/desitarget/cuts.py
moustakas: Good catch @geordie666, thank you. We should *definitely* not be applying the `GALAXY` mask for BGS selection (definitely not for SV and probably not for the main survey, either). 

Another selection tidbit is that `CLUSTER` also throws out the Fornax and Sculptor local-group dwarfs, but these will probably never be BGS targets so I think we're OK.

doc/changes.rst
qmxp55: Ups! I'll fix it right away.

py/desitarget/cuts.py
qmxp55: Yes, that's the idea, and for SV we don't mask around large galaxies. However, this cuts are for nominal and we have not decided yet the final DR9 selection. Most probably that we will end up not masking those, but we can come with other changes for dr9 selection so with this change I'm just going back to dr8 selection.

geordie666: Ah, I see, this change is just for the Main Survey selection. I guess we just have to stay on top of the issue of what to do about `MASKBITS` for BGS for the Main Survey.

doc/changes.rst
qmxp55: Cori is down, I'll fix this once its back.

geordie666:
@qmxp55: I'll merge this in a couple of hours and fix the `changes.rst` file myself as part of #658. Let me know if you _don't_ want me to do so, though.

geordie666 (approved):

qmxp55:
Thanks, @geordie666 I've just fixed this.

PR #660, opened 2020-12-08
re-tune LRG cuts for DR9 and modify LRG SV bits

rongpu:

PR #661, opened 2020-12-09
Finalize retuning of LRG cuts for DR9 and modify LRG SV bits

geordie666:
This PR supersedes #660. It includes code to:

- Retune the LRG cuts for DR9 and update the LRG SV target bits.
- Switch to only using the default `BRIGHT`, `GALAXY` and `CLUSTER` masks in `MASKBITS` (i.e. ignoring `ALLMASK` and `MEDIUM`).
- Increase the SV faint limits from `z < 20.5` to `z < 21.0` and from `zfiber < 21.9` to `zfiber < 22.0`.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,6 +5,12 @@ desitarget Change Log
 0.45.2 (unreleased)
 -------------------
 
+* Re-tune LRG cuts for DR9 and updating the LRG SV target bits [`PR #660`_].
+  Other changes include:
+    * Only use the default BRIGHT, GALAXY and CLUSTER masks (ignoring ALLMASK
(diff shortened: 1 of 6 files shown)


geordie666:
@rongpu: I incorporated your work from PR #660 into this branch. A bunch of function calls needed to be modified to account for your new inputs to the `isLRG()` function and I figured it would be most expedient to make those changes myself. I believe that unit tests should now pass on this branch. But, could you quickly check that I didn't actually change any of your intended cuts?

If you think this looks good, I'll merge it once unit tests pass.

rongpu:
@geordie666 I checked these changes and they look good. Thanks!

PR #669, opened 2021-01-05
First run of SV secondaries with real target files.

geordie666:
This PR updates the secondary code based on a first full-scale run of secondary targets for SV. It includes:
- Some general optimization, updating and debugging of the secondary code.
- Matching to the input sweeps files, not just to the actual targets, to find and populate primary `TARGETIDs`.
- Add the initial full bitmask for the `sv1` secondary target files.
- Updates to the data model for the secondary targets to better reflect the data model of the primary targets.

### DIFF
--- a/bin/select_secondary
+++ b/bin/select_secondary
@@ -43,7 +43,7 @@ hdr = fitsio.read_header(fns[0], 'SCND_TARG')
 surv = hdr["SURVEY"].rstrip()
 
 # ADM find the SCND_DIR environment variable, if it wasn't passed.
-scxdir = _get_scxdir(ns.scnddir)
+scxdir = _get_scxdir(ns.scnddir, survey=surv)
 # ADM and augment the scxdir if this is an SV set of primary files.
 if surv != 'main':
     scxdir = os.path.join(scxdir, surv)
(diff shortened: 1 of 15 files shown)


geordie666:
This PR now also adds outside-of-the-footprint, Gaia-only standard stars to the MWS masks for SV and the Main Survey. The new classes are:
- `GAIA_STD_FAINT`
- `GAIA_STD_BRIGHT`
- `GAIA_STD_WD`

the selections for these bits use a combination of existing code to implement:

- For `GAIA_STD_WD`:
  - The Main Survey `STD_WD` class for `GAIA_STD_WD`, just without matching to the Legacy Surveys.

- For `GAIA_STD_FAINT` and `GAIA_STD_BRIGHT`:
  - Our standard point source cut on Gaia Astrometric Excess Noise.
  - Our Gaia quality cuts for standards.
  - An isolation criterion preventing any standard from being closer than 3.5" to another Gaia source that is within 5 mags of brightness, borrowed from the `STD_DITHER_GAIA` cmx class.
  - Color cuts supplied by Arjun Dey that mimic our inside-of-the-footprint selections for `STD_FAINT` and `STD_BRIGHT` (which rely on Legacy Surveys fluxes).

PR #672, opened 2021-01-22
Update MWS nearby and BHB cuts for main survey

apcooper:
- Adds a bright limit to the main survey selection for MWS_NEARBY, to match what is written on the wiki (and the SV selection)
- Adds MWS_BHB as a main survey target class; maybe better to put this in now, although in principle the parameters may change (or the class *may* be dropped altogether, although unlikely) following SV.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,12 @@ desitarget Change Log
 0.49.1 (unreleased)
 -------------------
 
-No changes yet.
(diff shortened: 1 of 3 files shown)


apcooper:
I'll update the main TS wiki page with the BHB cuts if this PR is ok.

geordie666 (approved):
Looks good.

PR #673, opened 2021-01-29
Improve functionality for Gaia-only targets.

geordie666:
This PR includes a range of improved functionality for Gaia-only targets:
- Deredden the Gaia colors when selecting Gaia-only standards.
    * Include a general function to de-extinct Gaia magnitudes.
    * Add `E(B-V)` as an `EBV` column in the supplemental targets for downstream spectrophotometric calibrations.
- Add a `flavor` property to the bitmask for the secondary targets.
     * This `flavor` can be `"SPARE"`, `"DEDICATED"`, `"SSV"`, `"QSO"` or `"TOO"`.
     * Also add a unit test to check the allowed values of `flavor`.
- Update the priorities for "dedicated" secondary targets.
     * This is facilitated by the fact that such targets can be distinguished from "spare" targets using `flavor`.
- Include `PHOTSYS` of `"G"` for targets that are selected using only Gaia.
- Fix some minor bugs in directory structures and columns in `pixweight` files that arose due to changes in the data model.
- Add code to scrape the Gaia EDR3 flat (csv) files from the web.
    * Also code to reformat them as FITS files, and by HEALPixel.


geordie666:
I need to run new targets to go back on-sky tomorrow. So, I'll merge this and make a new tag after tests pass (likely early tomorrow morning).

PR #675, opened 2021-02-01
Migrate to GitHub Actions

weaverba137:
This PR migrates CI testing to GitHub Actions.  As a by-product, many documentation warnings were cleaned up.

@geordie666, the structure of desitarget looks a bit "shaggy" to me, for lack of a better word.  Are you fully tracking the development of the package?

PR #676, opened 2021-02-11
First Machinery for Targets of Opportunity.

geordie666:
This PR adds the initial formalism for Targets of Opportunity (ToO). It includes:
- A New ToO module.
- Functionality for writing an initial (example) ToO `.ecsv` ledger (with an associated command-line script).
    * This ledger will be monolithic. New ToOs will just be appended to the end of it.
- Functionality to convert the ToO ledger to a targets file with the necessary columns to run `fiberassign` (with an associated command-line script).
- The process of converting the ToO ledger, which the time domain team will be allowed to directly edit, to the targets file includes:
    * Checks on the general format and fidelity of the ledger.
    * Checks that observations in the ledger do not exceed any fiber allocation that the collaboration might decide to bestow.

An example initial ledger is available at NERSC in:
```
/global/cscratch1/sd/adamyers/ToO/dr9/0.50.0.dev4501/ToO/ToO-input.ecsv
```
Hopefully many of the columns in that file are self-explanatory, but here are some notes:
- The time domain team will provide a range of MJDs (`MJD_BEGIN`, `MJD_END`) over which a ToO observation has utility. The fiberassign code will need to read those MJDs (somehow) when scheduling observations.
- Each ToO will be associated with a type of observation (`TOO_TYPE`) which can be either `FIBER` or `TILE` to allow reallocating individual fibers, or hijacking targets across all (or most) of a tile.
- Every possible time domain observation would need to be vetted by someone who would add their initials in the `"CHECKER"` column.
    * At a minimum, this check would involve looking at the location in the [viewer](https://www.legacysurvey.org/viewer) to check that there were no bright stars near the ToO.
    * If `TOO_TYPE==FIBER` observations ever adversely affected the primary, Key Project targets, then, presumably, the ToO program would be suspended---possibly, indefinitely, or maybe just until the `CHECKER` had explained the mistake.


weaverba137:
Yes, as you noticed the configuration of GitHub Actions has nothing to do with the values that Coveralls uses to determine whether a change in coverage is acceptable. Please make sure that any changes made to the GitHub Actions configuration file are removed.

I also do not recommend changing the Coveralls configuration and the values it uses to determine acceptable coverage.

**You are free to ignore the results of any coverage test and merge anyway.** A drop in coverage does not disable the Merge button.

For a relatively large package like desitarget, 1% drop in coverage is a lot of lines. It is important that everyone be aware of that even if you choose to merge anyway for the sake of expediency.

Although the coverage is measured on the branch, that value becomes the coverage of master after the merge.

geordie666:
Ah, thanks @weaverba137, that all makes sense and is useful information. I have reverted all my changes to the GitHub Actions configuration and will stop experimenting with coverage checks! Thanks again for migrating `desitarget` to GitHub actions in #675.

PR #679, opened 2021-03-10
Add the SV2 infrastructure in preparation for the ".1% Survey"

geordie666:
This PR is mostly intended to prepare the SV2 infrastructure for the so-called ".1% Survey". It also incorporates a few other recent decisions. This PR:

- Adds an `sv2_cuts.py` module and an sv2 bitmask yaml file.
    * The sv2 cuts can be run by, e.g., passing `--iteration 2` to the `select_sv_targets` script.
- Renames `POOR` observing conditions to be `BACKUP` conditions throughout the `desitarget` code.
- Retains `BACKUP` targets that are duplicates of `DARK` or `BRIGHT` targets. 
    * Previously, we'd been removing Gaia-only `BACKUP` sources that were duplicates of Legacy Surveys targets. 
    * Such duplicated targets have different `TARGETID`s (one derived from Gaia and one from the Legacy Surveys). 
    * It was decided it was preferable to always retain these as Gaia-only `BACKUP` targets, even when they were also a primary, Key Project target derived from the Legacy Surveys.
   * If needed, such duplicates can be identified and removed based on the `desitarget` `REF_ID` (which corresponds to `SOURCE_ID` in Gaia).

### DIFF
--- a/bin/select_sv_targets
+++ b/bin/select_sv_targets
@@ -77,11 +77,16 @@ ap.add_argument("-noc", "--nochecksum", action='store_true',
                 help='Do NOT add the list of input files and their checksums to the output target file as the second ("INFILES") extension')
 
 ns = ap.parse_args()
+
+# ADM never run secondary targets for sv2.
+if ns.iteration == '2':
+    ns.nosecondary = True
+
 # ADM build the list of command line arguments as
 # ADM bundlefiles potentially needs to know about them.
 extra = " --numproc {}".format(ns.numproc)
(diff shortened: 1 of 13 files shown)


geordie666:
@Cyeche and @echaussidon: Once I merge this PR (likely in the morning, so in about 12 hours) the necessary code for SV2 will be in place. You can add a retrained Random Forest for the ".1% Survey" by changing the usual `isQSO_randomforest()` function, but now in the `sv2_cuts.py` module.

PR #680, opened 2021-03-12
NEW QSO target selection

echaussidon:
Hi @geordie666, 

I add the new qso target selection in the sv2 code !

I check if it works copying isQSO_randomforest in the main selection (now it is removed of course)

I don't run the sv2 selection since I don't know the command.

Edmond


echaussidon:
(just for notification : @Cyeche)

echaussidon:
The method was presented to the group during the TS meeting (10/03/21) [slides on DocDB](https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=6112)

geordie666:
@echaussidon: For future reference, the command to run the `sv2` targets is the same as the command to run the `sv1` targets but with `--iteration 2` (and the command to run the `sv3` targets will be the same but with `--iteration 3`). So, for example:

```
LS=/global/cfs/cdirs/cosmo/data/legacysurvey/dr9
select_sv_targets $LS/south/sweep/9.0 $CSCRATCH -s2 $LS/north/sweep/9.0 --nside 8 --healpixels 508 --numproc 8 --iteration 2
```

I've tried this on your branch, and it runs to completion, so I think this looks good. 

But, before I merge this, I have two requests for changes related to documentation and style:

(1) It looks like you've added some new modules in `desitarget/py/desitarget/train`. Could you please include those new modules in the API at `desitarget/doc/api.rst` (the syntax to do this should be obvious from the other `desitarget.train` modules in the `api.rst` file).

(2) You've added a lot of code that doesn't conform to style. See anything that doesn't include the word "mocks" under "Test the style; failures are allowed" here:

https://github.com/desihub/desitarget/pull/680/checks?check_run_id=2094650704

Please try to clean up that code-style much as possible.

echaussidon:
Sorry Adams, I didn't check that before !

I clean the code, i think every thing is good now !

geordie666:
Thanks, @echaussidon, this looks great. I really appreciate all the work you're willing to do to make this easier on my side. I'll merge this in an hour or so.

PR #682, opened 2021-03-13
New secondary bits for the COSMOS dedicated programs and for unusual point sources.

geordie666:
This PR:

- updates the SV1 secondary targeting bits with the latest (final) versions for the dedicated programs in COSMOS.  
- deprecates some earlier bits for those COSMOS programs which correspond to defunct selections.
- adds the bits for the unusual point sources secondary program (`PSF_OUT_BRIGHT`, `PSF_OUT_DARK`).

There is no new code, other than updating bit-masks, so I'll merge as soon as tests pass to make a new tag for the new secondary targets.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,6 +5,7 @@ desitarget Change Log
 0.51.1 (unreleased)
 -------------------
 
+* New secondary bits for COSMOS/unusual point sources [`PR #682`_].
 * Add formalism to make ledger for BACKUP targets [`PR #681`_].
 * New QSO target selection in SV2 [`PR #680`_] for validation:
     * RF file (dr9_final) trained with significance > 10 and SV QSOs.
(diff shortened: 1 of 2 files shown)


sbailey:
@geordie666 can you confirm that the deprecated/replaced bits were never used for tiles that were observed?

geordie666:
I can't. Unfortunately, because of the way secondary targeting works, if people request a similar targeting set multiple times there is no way to determine an unambiguous SCND_ORDER for those target classes. So, a new bit has to be cut and the previous bit _has_ to be removed. The deprecated bits should all be for DEDICATED programs, though. So, people who are interested in DEDICATED secondary programs just need to remember to use the mask that corresponds to the version of the code used to process their targets. This is unfortunate, but I hadn't envisioned that we'd be updating secondary targets in official versions of the code 2, 3 even 4 times before observing DEDICATED tiles.

PR #686, opened 2021-03-18
Install sv2 data files.

tskisner:
The sv2 data files were not being installed with setup.py.  This adds that data directory.


geordie666:
Ah, I just forgot to do this install. My bad. Thanks for catching this, and feel free to self-merge once tests pass.

PR #688, opened 2021-03-19
Don't pass the Data Release when constructing MTL filenames

geordie666:
As part of #684, we changed the data model such that the directory path to MTL files no longer includes the Data Release. This broke the MTL-specific read routines, such as `desitarget.io.read_targets_in_tiles(..., mtl=True)` which were still expecting to need to pass around Data Release names.

This is a very minor PR to fix that bug.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,11 +5,13 @@ desitarget Change Log
 0.53.1 (unreleased)
 -------------------
 
+* Don't pass the DR when constructing MTL filenames [`PR #688`_].
 * Don't insist that ``ZTILEID`` has to be in the ``zcat`` [`PR #687`_].
 * Install the SV2-related data files [`PR #686`_].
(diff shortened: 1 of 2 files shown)


geordie666:
@araichoor: Thanks for finding this bug when reading MTL ledgers and reporting it to me. I think this PR will fix your issues. If this _does_ work, then please feel free to merge this PR (when tests pass) if it's a blocking factor for your work.

Alternatively, I can merge this PR myself when I get out of meetings in an hour or two.

Also: Let me know if you need me to make a new tag of `desitarget` in the short-term so that you can make "official" use of this update.

araichoor:
I confirm it runs fine now; I let you do the merging.
about a new tag: right, that may be a cleaner way to proceed (but won t change the files); if that s not too much hassle for you, yes.

PR #692, opened 2021-03-26
TS QSO : Update r_min from 17.5 to 16.5 for SV2

echaussidon:
Hi @geordie666, 

As I shown during the TS meeting wednesday, we decrease the minimum value of r required inside isQSO_randomforest from 17.5 to 16.5. 

Could you confirm that this change in SV2 will be imported in SV3 ? 

Cheers, 
Edmond

### DIFF
--- a/py/desitarget/sv2/sv2_cuts.py
+++ b/py/desitarget/sv2/sv2_cuts.py
@@ -1475,8 +1475,7 @@ def isQSO_randomforest(gflux=None, rflux=None, zflux=None, maskbits=None,
 
     # Preselection to speed up the process
     rMax = 23.0   # r < 23.0
-    print("[INFO] WE cut r from 17.5 to 23.0")
-    rMin = 17.5   # r > 17.5
+    rMin = 16.5   # r > 16.5
     preSelection = (r < rMax) & (r > rMin) & photOK & primary
 
     # ADM targets have to be observed in every band.
--- a/py/desitarget/sv2/sv2_cuts.py
+++ b/py/desitarget/sv2/sv2_cuts.py
@@ -1475,6 +1475,7 @@ def isQSO_randomforest(gflux=None, rflux=None, zflux=None, maskbits=None,
 
     # Preselection to speed up the process
     rMax = 23.0   # r < 23.0
+    print("[INFO] WE cut r from 17.5 to 23.0")
     rMin = 17.5   # r > 17.5
     preSelection = (r < rMax) & (r > rMin) & photOK & primary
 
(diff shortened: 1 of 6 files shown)


echaussidon:
@Cyeche (for notification)

echaussidon:
@geordie666, I don't understand the problem with the Test coverage..

geordie666:
@echaussidon, @Cyeche: There should be no new changes to SV2 at this stage. Please remove your change from `sv2_cuts.py` and put it in `sv3_cuts.py` instead. As SV3 has already been created, any changes to SV2 will _not_ propagate to SV3 and will simply be lost.

Note that r of about 16.5 seems quite bright, to me. @julienguy has suggested that the dynamic range of DESI observations is something like 5 magnitudes. With this change, your target class will span 6.5 magnitudes in r. If there isn't much to be gained in the range 16.5 < r < 17, I'd consider limiting to r~17 (or a little fainter) instead. Although, I appreciate that the number of targets in the range 16.5 < r < 17 will be low, so perhaps it's not a major issue.

The unit tests fail (I think) because you're running on your own fork rather than a branch of `desitarget`. The coverage tests pass around a special token that isn't available outside of the main `desihub` repositories. There isn't an easy fix to this, unfortunately. But, if the _only_ test that fails is `Test coverage` then don't be concerned!

echaussidon:
@geordie666 : I change for sv3 (it was not fully clear for me!)

I decrease the r_min value to collect bright quasars that BGS program cannot collect (since they are PSF) and I shown that the density is not impacted by this additional targets.

py/desitarget/sv2/sv2_cuts.py
```diff
@@ -1475,7 +1475,7 @@ def isQSO_randomforest(gflux=None, rflux=None, zflux=None, maskbits=None,
 
     # Preselection to speed up the process
     rMax = 23.0   # r < 23.0
-    rMin = 16.5   # r > 16.5
+    rMin = 15.5   # r > 17.5
```
geordie666, line 1478: Can you restore this to `rMin = 16.5`? We likely won't run any new SV2 target files, so we shouldn't change anything in `sv2_cuts.py`, if possible.

geordie666 (changes requested):
@echaussidon: One more request. Can you restore the original cuts in SV2, as we're "done" with SV2 and it's probably best not to update it further, now (see the comment, above).

Also, can you make sure to update the [SV3 wiki](https://desi.lbl.gov/trac/wiki/TargetSelectionWG/SV3?action=diff&version=5&old_version=3)? Your current cuts in the code are for `rMin = 17.0` but the cuts on the wiki still state `r > 16.5`

echaussidon:
@geordie666: Sorry a tipo mistake for SV2. 

I check the cut on the code is 16.5 which is what is in the wiki !

geordie666:
OK, makes sense now, thanks! I'll merge once the necessary tests pass.

PR #693, opened 2021-03-26
Update MTL loop for new data model

geordie666:
This PR updates the MTL "loop" to adhere to the new directory structure for the spectroscopic pipeline. It includes:
- Functionality to make initial MTL ledgers for secondary targets.
- Use the new `ZTILE` file to look-up redshift "done-ness" (`zdone`) instead of relying on a "done" directory.
- Also derive the survey (`main/sv2`/etc.) and program/obscon (`DARK/BRIGHT/BACKUP`) from the new `ZTILE` file.
- Switch to reading the `zbest` files in the _cumulative_ directory.
- Remove negative `TARGETID`s before matching to ensure one-to-one matches.
- Add the `YYYYMMDD` date on which redshifts were extracted to the `mtl-done-tiles.ecsv` file.
- Add the root directory for looking up the redshift information (`ZCAT_DIR`) to the `etc/desitarget.module` file.

### DIFF
--- a/bin/run_mtl_loop
+++ b/bin/run_mtl_loop
@@ -24,10 +24,6 @@ ap.add_argument('--mtldir',
                 help="Full path to the directory that hosts the MTL ledgers.    \
                 Default is to use the $ZCAT_DIR environment variable",
                 default=None)
-ap.add_argument('--tilefn',
-                help="Full path to the name of the tile file.                   \
-                Default is to use the $TILE_FN environment variable",
-                default=None)
 ap.add_argument("-n", "--numobsfromzcat", action='store_true',
                 help="If passed, the zcat includes a meaningfule NUMOBS.        \
                 Otherwise, NUMOBS is looked up from the ledger itself")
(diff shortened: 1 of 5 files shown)


geordie666:
@schlafly: These are the updates required to run the full MTL loop with the new data model. I'd like to leave this PR open for about an hour, and then, once I've merged it, I'll need to tag a new version of `desitarget`.

After that, I can run updated `BRIGHT` files for `sv2` and commit them to svn if we think we'll want to observe such tiles tonight. I have a couple of meetings this afternoon, so that all might be finished at, say 2PM Pacific?

schlafly:
That sounds like a reasonable schedule for enabling observations to be made tonight.  If you're also asking for comments on this PR, I'd appreciate a pointer to the area you'd like me to focus on.  Thanks!

geordie666:
No need for comments, I think. I've tested everything end-to-end carefully (and already processed all of the ledgers in test mode). I was just checking I wasn't blocking anyone by holding off until ~2PM.

PR #695, opened 2021-03-27
Bgs1percent

qmxp55:
BGS selection for sv3/1% survey. Main changes include:

- BGS faint goes as deep as r-tot mag of 20.3 and includes a fibermag-colour based selection.
- BGS bright holds same r-tot mag limits (< 19.5).
- We've revived the WISE bit TARGTYPE that selects AGN's (~3 /deg^2).
- WE've increased the r-tot mag faint limits to around 0.025 mags for BASS/MzLS only. It goes to r-tot < 19.5+0.025 for the BRIGHT sample and to 19.5+0.025 < r-tot < 20.3+0.025 for the FAINT sample.
- HIP was changed to promote 20% targets from the FAINT sample instead of 10%.

I've tested the implementations with my local copy of BGS for one SWEEP brick in the north and in the south and they match perfectly.

I'll update the wiki later in the weekend.


geordie666:
@qmxp55: I'll review this in full early tomorrow morning. But, two quick comments for now:

- Did you apply a bright-end limit? In particular, I'm worried that `BGS_BRIGHT` might still select very bright targets.
- This line: `hip = np.random.choice(w, nbgsf//20, replace=False)` seems like it should promote _5%_ of targets to being high priority, not your intended 20%?

Apologies if I've missed something obvious related to these points. As I said, this is just some super-quick feedback in case your working more on this today, and I'll provide a more careful review tomorrow.

qmxp55:
@geordie666: Thanks for the quick review, it was helpful. I've added the bright limit at r-tot mag of 12 and corrected the HIP to `nbgsf//5`.

py/desitarget/sv3/sv3_cuts.py
```diff
@@ -1252,65 +1262,107 @@ def isBGS_colors(rfiberflux=None, gflux=None, rflux=None, zflux=None, w1flux=Non
...
+    rfibcol =(rfib < 20.75) | ((rfib < 21.5) & (schlegel_color > 0.))
+
+    #BASS r-mag offset with DECaLS
+    offset = 0.025
+    
     if targtype == 'bright':
-        bgs &= rflux > 10**((22.5-19.5)/2.5)
+        if south:
+            bgs &= rflux > 10**((22.5-19.5)/2.5)
+            bgs &= rflux <= 10**((22.5-12.0)/2.5)
```
geordie666, line 1290: I'd think we want bright-end cuts much fainter than `12.0`. People have generally been suggesting more like `> 16` (for extended sources) and using `FIBERTOTFLUX` instead of `FLUX`. I'm not sure what's appropriate for high-surface-brightness galaxies like the `BGS_BRIGHT` sample in `BRIGHT` time, though. @araichoor and @deisenstein might want to comment.

py/desitarget/sv3/sv3_cuts.py
```diff
@@ -1252,65 +1262,107 @@ def isBGS_colors(rfiberflux=None, gflux=None, rflux=None, zflux=None, w1flux=Non
...
+        else:
+            bgs &= rflux > 10**((22.5-(19.5+offset))/2.5)
+            bgs &= rflux <= 10**((22.5-12.0)/2.5)
```
geordie666, line 1293: Again, I'd think we want a bright-end limit of more like `16.0` than `12.0`.

py/desitarget/sv3/sv3_cuts.py
```diff
@@ -1237,6 +1232,21 @@ def isBGS_colors(rfiberflux=None, gflux=None, rflux=None, zflux=None, w1flux=Non
         primary = np.ones_like(rflux, dtype='?')
     bgs = primary.copy()
     fmc = np.zeros_like(rflux, dtype='?')
+
```
geordie666, line 1235: In the previous version, we kept the `isBGS_lslga()` function (which should now be called `isBGS_sga()`, I guess) separate rather than embedding that part of the selection in `isBGS_colors()`. 

This was a deliberate choice because the mock simulations don't include a concept of `refcat`. Please return the code to its original structure, with a separate `isBGS_sga()` function.

py/desitarget/sv3/sv3_cuts.py
```diff
@@ -1237,6 +1232,21 @@ def isBGS_colors(rfiberflux=None, gflux=None, rflux=None, zflux=None, w1flux=Non
         primary = np.ones_like(rflux, dtype='?')
     bgs = primary.copy()
     fmc = np.zeros_like(rflux, dtype='?')
+    
+    # the LSLGA galaxies.
```
geordie666, line 1236: While we're updating `isBGS_lslga()`->`isBGS_sga()`, this comment should be updated to "SGA" instead of "LSLGA".

py/desitarget/sv3/sv3_cuts.py
```diff
@@ -1252,65 +1262,107 @@ def isBGS_colors(rfiberflux=None, gflux=None, rflux=None, zflux=None, w1flux=Non
...
     r = 22.5 - 2.5*np.log10(rflux.clip(1e-16))
     z = 22.5 - 2.5*np.log10(zflux.clip(1e-16))
+    w1 = 22.5 - 2.5*np.log10(w1flux.clip(1e-16))
     rfib = 22.5 - 2.5*np.log10(rfiberflux.clip(1e-16))
 
     # Fibre Magnitude Cut (FMC) -- This is a low surface brightness cut
     # with the aim of increase the redshift success rate.
     fmc |= ((rfib < (2.9 + 1.2 + 1.0) + r) & (r < 17.8))
     fmc |= ((rfib < 22.9) & (r < 20.0) & (r > 17.8))
+    # Remove next line?
```
geordie666, line 1272: Do you need this comment? If you do, maybe change it to something like "Remove next line for the Main Survey?".

py/desitarget/sv3/sv3_cuts.py
```diff
@@ -1252,65 +1262,107 @@ def isBGS_colors(rfiberflux=None, gflux=None, rflux=None, zflux=None, w1flux=Non
...
+            bgs &= rflux > 10**((22.5-(20.3+offset))/2.5)
+            bgs &= rflux <= 10**((22.5-(19.5+offset))/2.5)
+            bgs &= (rfibcol)
 
     return bgs
 
 
-def isBGS_lslga(gflux=None, rflux=None, zflux=None, w1flux=None, refcat=None,
-                maskbits=None, south=True, targtype=None):
-    """Module to recover the LSLGA objects in all BGS target selection classes
```
geordie666, line 1279: LSLGA -> SGA

geordie666 (changes requested):
Thanks @qmxp55. In general this looks great. I have a few requested changes. 

From the perspective of code structure, the most important request is to reestablish the separate `isBGS_lslga()` function (which would now be called `isBGS_sga()`). It's important for this to be separate as the mock/simulation code expects code that requires `refcat` to _not_ be included in `isBGS_colors()`.

py/desitarget/sv3/sv3_cuts.py
qmxp55: 16 seems to aggressive, by doing so we'll be removing ~14 /deg^2 of which 10 /deg^2 are SGA galaxies. What about a cut in r and rfibermag, `r > 12 & rfibmag > 15`, this will remove only 0.3 /deg^2 of which 0.06 /deg^2 are SGA.
image

py/desitarget/sv3/sv3_cuts.py
qmxp55: I see. I'll put it back.

qmxp55:
Right, I've added back the SGA module (previously LSLGA) and changed the bright limit to `r > 12 & rfibmag > 15`

deisenstein:
We definitely should be setting our bright limits in terms of fiberflux
(and really fibertotflux) rather than total flux.

In the spirit of the 1% survey, I'm ok with r_fibertot>15.  But I think
we'll need to check after the 1% whether we need to tighten this a little
more to limit fiber crosstalk.

Thanks!

py/desitarget/sv3/sv3_cuts.py
geordie666: Let's adopt that for now and then make a final decision next week. I think the issue is that the 10/deg^2 we'd remove at the bright-limit that are SGA galaxies could saturate anyway, so there's no realistic way to observe them.

I'm happy to merge with your chosen `r > 12 & rfibmag > 15` but you should ask on-list (specifically the operations mailing list) what is a realistic bright-end limit to adopt.

geordie666: Ah, I see @deisenstein made a comment as I was writing this. can you use rfibertotmag (the mag based on rfibertotflux) instead? So `r > 12 & rfibtotmag > 15`?

qmxp55: Hi @geordie666, is this `rfibtotmag` available in sv3? I cannot find it.

qmxp55: https://github.com/desihub/desitarget/blob/master/py/desitarget/sv3/sv3_cuts.py#L1749-L1761

geordie666: @qmxp55: The `rfibertotflux` column is available, but it's a little tricky to pass it across the different iterations of sv. For now, let's just use your cut of `r > 12 & rfibmag > 15`. I'm currently working on a different branch, and I'll add passing `rfibertotflux` around in _that_ branch.

Let me know when you've added the `r > 12 & rfibmag > 15` cut and I'll merge this PR.

geordie666: I just reviewed the code carefully again, and it looks like you already included the `r > 12 & rfibmag > 15` cut. Thanks!

geordie666:
@qmxp55: I think we resolved everything. Let me know whether you intend to make any additional changes. Otherwise, I'll merge this early tomorrow morning.

PR #696, opened 2021-03-27
Update the ELG selection for SV3

geordie666:
This PR updates the ELG selection for SV3 (i.e. the 1% survey). In particular, it adds two classes of ELGs, a low-priority class and a high-priority class.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,9 +5,14 @@ desitarget Change Log
 0.54.1 (unreleased)
 -------------------
 
-* Update the LRG selection for SV3 (1% Survey):
+* Update the ELG selection for SV3 (1% Survey) [`PR #696`_]:
+    * Add functionality for low (LOP) and high (HIP) priority ELGs.
+* Update the LRG selection for SV3 (1% Survey) [`PR #694`_]:
     * Add fainter (and higher redshift) LRG targets
-    
+
+.. _`PR #694`: https://github.com/desihub/desitarget/pull/694
+.. _`PR #696`: https://github.com/desihub/desitarget/pull/696
+
 0.54.0 (2021-03-26)
 -------------------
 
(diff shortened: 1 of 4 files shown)


geordie666:
@araichoor: Can you check my working, here? Also what relative initial (`"UNOBS"`) priorities would you want for the [high-priority ELGs](https://github.com/desihub/desitarget/blob/8a98f4caf30e0aeb8b572b3509edf596235339d0/py/desitarget/sv3/data/sv3_targetmask.yaml#L201) and the [low-priority ELGs](https://github.com/desihub/desitarget/blob/8a98f4caf30e0aeb8b572b3509edf596235339d0/py/desitarget/sv3/data/sv3_targetmask.yaml#L200)?

araichoor:
hi Adam,

I m out for the day (probably until like 5pm pacific): would it be too late?
I just had a look at the code on my phone, it looks fine.
About unobs: I m not sure how that works..

thanks,

anand

geordie666:
@araichoor: No problem, 5PM Pacific will be fine. `UNOBS` is just the initial priority assigned to a target. So, let me put it this way: If the main dark time classes currently have the following priorities:
```
ELG: 3000 
LRG: 3200
QSO: 3400
```
What should I assign as the priority for the low-priority ELGs? And, what should I assign as the priority for the high-priority ELGs? 

Is the goal just to differentiate the ELGs (so, then, say the low-priority ELGs could be `3000` and the high-priority ELGs could be `3010`)?

Or, is the goal to make the high-prioritiy ELGs a higher priority than other target classes (so, then, say the low-priority ELGs could be `3000` and the high-priority ELGs could be `3250`)?

araichoor:
ok I see; well we can start with priorities internal to elgs.
putting elgs at lrg or qso priority deserves wider collaboration discussion.

thanks,

anand

araichoor (approved):
looks good, thanks!

PR #698, opened 2021-03-29
Another update to the data model for the MTL loop 

geordie666:
This PR updates the expected directory structure for the MTL loop. 

Previously, the loop assumed that the spectroscopic pipeline would only maintain one night in the `cumulative` directory and would unlink previous nights. With this update, the loop instead looks up the (`YYYYMMDD`) night that corresponds to `LASTNIGHT` in the redshift `tiles.csv` file for a given tile and then uses the `zbest` files in the corresponding directory.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,6 +5,7 @@ desitarget Change Log
 0.54.1 (unreleased)
 -------------------
 
+* Another update to the data model for the MTL loop [`PR #698`_].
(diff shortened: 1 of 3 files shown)


geordie666:
This is a (relatively) minor update, and I've checked it works second-time-through-the-loop. I'll merge this in about an hour so I can make a new tag of `desitarget` and use it to update the `BRIGHT/sv2` ledgers in svn.

PR #700, opened 2021-03-30
Some bug fixes for SV3.

geordie666:
This PR fixes some bugs for SV3:

- The `BGS_WISE` bit had been turned off for a while. When we redefined the `BGS_WISE` selection to mean "AGN-like" targets, we didn't turn the `BGS_WISE` bit back on, so there were no `BGS_WISE` targets in the target files. `BGS_WISE` is turned on for `sv3` in this PR.
- The MTL decisions were still being driven by the `ELG` bit, meaning that `ELG_HIP` had zero effect on priorities after one run through MTL. The `make_mtl()` code will now understand the `ELG_HIP` bit for `sv3`.
  * To help fix this and to simplify bit-definitions, I've redefined `ELG` to mean "low-priority ELGs" and retained `ELG_HIP` to mean "high-priority ELGs". Previously I'd decided to use `ELG` to mean "either `ELG_LOP` or `ELG_HIP` (i.e. `ELG_LOP | ELG_HIP`). The `ELG_LOP` bit still exists, it just won't be turned on.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,10 +5,15 @@ desitarget Change Log
 0.55.1 (unreleased)
 -------------------
 
+* Some bug fixes for SV3 [`PR #700`_]. Includes:
+    * Turn on the ``BGS_WISE`` bit, which had been deprecated.
+    * Correct behavior for ``ELG_HIP`` when making MTLs.
+        * Also change ``ELG_LOP`` to just ``ELG`` for simplicity.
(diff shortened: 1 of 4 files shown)


geordie666:
@michaelJwilson and @araichoor: Thanks for finding these bugs. I'll run new targeting files soon and email the desi-targets list with the location of the new files.

PR #705, opened 2021-04-02
updates in BGS SV3

qmxp55:
Updates include:

1. Bright limit in r-mag > 12 & rfibertotmag > 15.
2. Magnitude offset in r-mag and rfibermag limits of 0.04 for BASS/MzLS only to get a target density closer to DECaLS.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,11 @@ desitarget Change Log
 0.56.1 (unreleased)
 -------------------
 
-* No changes yet.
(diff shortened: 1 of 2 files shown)


geordie666:
@qmxp55: This looks fine to me, and I've confirmed that the new code runs to completion at NERSC. I'm leaving this PR open for a while, though, as @djschlegel may want to scan through your new code. I'll merge it in about an hour with or without further review.

Please fix the (usual) conflicts in the `doc/changes.rst` file. Or, I can fix them before I merge.

PR #711, opened 2021-04-11
Fix new ZWARN bit-mask unit test.

geordie666:
This PR fixes the `ZWARN` unit test that was added in #710.


geordie666:
This is a very minor PR. I'll merge it in a couple of hours so that the nightly integration tests work again.

PR #714, opened 2021-04-20
Lyaretarget

eleanorlyke:
I updated the lyazcat script to incorporate QuasarNP, SQUEzE, and MgII Absorption with user options to turn off each of these as necessary (though MgII Absorption hasn't been completed yet).

### DIFF
--- a/py/desitarget/lyazcat.py
+++ b/py/desitarget/lyazcat.py
@@ -1,0 +1,184 @@
+#This is where the script for making the LyA retargeting ZCAT is going to go, separate from the mtl.py file.
+import os
+import numpy as np
+import healpy as hp
+import numpy.lib.recfunctions as rfn
+import sys
+from astropy.table import Table
+from astropy.io import ascii
+import fitsio
+from time import time
+from datetime import datetime
+from glob import glob
+
+from . import __version__ as dt_version
+from desitarget.targetmask import obsmask, obsconditions, zwarn_mask
+from desitarget.targets import calc_priority, calc_numobs_more
+from desitarget.targets import main_cmx_or_sv, switch_main_cmx_or_sv
+from desitarget.targets import set_obsconditions, decode_targetid
+from desitarget.geomask import match, match_to
+from desitarget.internal import sharedmem
+from desitarget import io
+
+# ADM set up the DESI default logger.
+from desiutil.log import get_logger
+log = get_logger()
+
+zcatdatamodel = np.array([], dtype=[
+    ('RA', '>f8'), ('DEC', '>f8'), ('TARGETID', '>i8'),
+    ('NUMOBS', '>i4'), ('Z', '>f8'), ('ZWARN', '>i8'), ('ZTILEID', '>i4'),
+    ('Z_COMB','>f8'),('Z_COMB_PROB','>f8'),('Z_QN','>f8'),('Z_QN_CONF','>f8'),
+    ('Z_SQ','>f8'),('Z_SQ_CONF','>f8')
+    ])
+
+def make_blank_zcat(zcatdir, tiles):
+    """Make the simplest possible zcat using SV1-era redrock outputs.
+    Parameters
+    ----------
+    zcatdir : :class:`str`
+        Full path to the "daily" directory that hosts redshift catalogs.
+    tiles : :class:`~numpy.array`
+        Numpy array of tiles to be processed. Must contain at least:
+        * TILEID - unique tile identifier.
+        * ZDATE - final night processed to complete the tile (YYYYMMDD).
+    Returns
+    -------
+    :class:`~astropy.table.Table`
+        A zcat in the official format (`zcatdatamodel`) compiled from
+        the `tiles` in `zcatdir`.
+    Notes
+    -----
+    - How the `zcat` is constructed could certainly change once we have
+      the final schema in place.
+    """
+    # ADM the root directory in the data model.
+    rootdir = os.path.join(zcatdir, "tiles", "cumulative")
+
+    # ADM for each tile, read in the spectroscopic and targeting info.
+    allzs = []
+    allfms = []
+    for tile in tiles:
+        # ADM build the correct directory structure.
+        tiledir = os.path.join(rootdir, str(tile["TILEID"]))
+        ymdir = os.path.join(tiledir, tile["ZDATE"])
+        # ADM and retrieve the redshifts.
+        zbestfns = glob(os.path.join(ymdir, "zbest*"))
+        for zbestfn in zbestfns:
+            zz = fitsio.read(zbestfn, "ZBEST")
+            allzs.append(zz)
+            # ADM only read in the first set of exposures.
+            fm = fitsio.read(zbestfn, "FIBERMAP", rows=np.arange(len(zz)))
+            allfms.append(fm)
+            # ADM check the correct TILEID was written in the fibermap.
+            if set(fm["TILEID"]) != set([tile["TILEID"]]):
+                msg = "Directory and fibermap don't match for tile".format(tile)
+                log.critical(msg)
+                raise ValueError(msg)
+    zs = np.concatenate(allzs)
+    fms = np.concatenate(allfms)
+
+    # ADM remove -ve TARGETIDs which should correspond to bad fibers.
+    zs = zs[zs["TARGETID"] >= 0]
+    fms = fms[fms["TARGETID"] >= 0]
+
+    # ADM check the TARGETIDs are unique. If they aren't the likely
+    # ADM explanation is that overlapping tiles (which could include
+    # ADM duplicate targets) are being processed.
+    if len(zs) != len(set(zs["TARGETID"])):
+        msg = "a target is duplicated!!! You are likely trying to process "
+        msg += "overlapping tiles when one of these tiles should already have "
+        msg += "been processed and locked in mtl-done-tiles.ecsv"
+        log.critical(msg)
+        raise ValueError(msg)
+
+    # ADM currently, the spectroscopic files aren't coadds, so aren't
+    # ADM unique. We therefore need to look up (any) coordinates for
+    # ADM each z in the fibermap.
+    zid = match_to(fms["TARGETID"], zs["TARGETID"])
+
+    # ADM write out the zcat as a file with the correct data model.
+    zcat = Table(np.zeros(len(zs), dtype=zcatdatamodel.dtype))
+
+    zcat["RA"] = fms[zid]["TARGET_RA"]
+    zcat["DEC"] = fms[zid]["TARGET_DEC"]
+    zcat["ZTILEID"] = fms[zid]["TILEID"]
+    zcat["NUMOBS"] = zs["NUMTILE"]
+    for col in set(zcat.dtype.names) - set(['RA', 'DEC', 'NUMOBS', 'ZTILEID']):
+        zcat[col] = zs[col]
+
+    return zcat
+
+def add_qn_data(zcat):
+    #This is where the QuasarNET stuff will go.
+    print('Adding QuasarNET data is not yet supported.')
+    zcat['Z_QN'][:] = -1
+    zcat['Z_QN_CONF'][:] = -1
+    
+    return zcat
+
+def add_sq_data(zcat):
+    #This is where the SQUEzE stuff will go.
+    print('Adding SQUEzE data is not yet supported.')
+    zcat['Z_SQ'][:] = -1
+    zcat['Z_SQ_CONF'][:] = -1
+
+    return zcat
+
+def zcomb_selector(zcat,proc_flag=False)
+    #This is where the final Z_COMB and Z_COMB_CONF will be selected and 
+    # the columns are populated.
+
+    #proc_flag is in case I want to add a procedure later that weighs
+    #model fits like eBOSS, or doing more complex comparisons between 
+    #QuasarNET and SQUEzE
+
+    zcat['Z_COMB'][:] = zcat['Z']
+    zcat['Z_COMB_PROB'][:] = 0.95
+
+    return(zcat)
+
+def zcat_writer(outputdir,zcat,qn_check,sq_check):
+    file_dtag = time.strftime('%Y%m%dT%H%M%S')
+    outputname = os.path.join(outputdir, 'zbest_lya_', file_dtag, '.fits')
+
+    prim_hdu = fits.PrimaryHDU()
+    prim_hdu.header['QN_ADDED'] = str(qn_check)
+    prim_hdu.header['SQ_ADDED'] = str(sq_check)
+
+    data_hdu = fits.BinTableHDU.from_columns(zcat,name='ZCATALOG')
+
+    data_out = fits.HDUList([prim_hdu,data_hdu])
+    data_out.writeto(outputname)
+
+    return outputname    
+
+
+if __name__=='__main__':
+    zcat_dir = sys.argv[1]
+    tile_list = sys.argv[2]
+    qn_check = False
+    sq_check = False
+
+    rr_zcat = make_empty_zcat(zcat_dir,tile_list)
+    if qn_check and sq_check:
+        qn_rr_zcat = add_qn_data(rr_zcat)
+        sq_qn_rr_zcat = add_sq_data(qn_rr_zcat)
+        fzcat = zcomb_selector(sq_qn_rr_zcat)
+
+    elif qn_check and not sq_check:
+        qn_rr_zcat = add_qn_data(rr_zcat)
+        fzcat = zcomb_selector(qn_rr_zcat)
+
+    elif sq_check and not qn_check:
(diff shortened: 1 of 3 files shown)


geordie666:
Thanks for pushing this. I'll try it out and make comments (although I won't hijack the branch and work on it myself until you're ready to merge).

py/desitarget/lyazcat.py
```diff
@@ -0,0 +1,441 @@
...
+        
+    Output
+    ------
+    :class:'str'
+        A string with the date and time in ISO 8061 standard followed
+        by the istring.
+    """
+    t0 = time.time()
+    t_start = time.strftime('%Y-%m-%d | %H:%M:%S')
+    print('\n{}: {}'.format(istring,t_start))
```
geordie666, line 67: Replace any `print` statements with `log.info` to use the DESI logger.

py/desitarget/lyazcat.py
```diff
@@ -0,0 +1,441 @@
...
+    :class:'~numpy.array'
+        The zcat array with QuasarNP data included in the columns:
+        * Z_QN
+        * Z_QN_CONF
+        * IS_QSO_QN
+    """
+    tmark('    Adding QuasarNP data')    
+    lines = ['LYA','CIV(1548)','CIII(1909)', 'MgII(2796)','Hbeta','Halpha']
+    lines_bal = ['CIV(1548)']
+    model = load_model('/global/cfs/cdirs/desi/science/lya/qn_models/boss_dr12/qn_train_coadd_indtrain_0_0_boss10.h5')
```
geordie666, line 121: Maybe have this load at the top of the code, so it's only run once at import and is available throughout the module as a global. So, e.g., call it `qn_model` and load it after your initial import statements.

py/desitarget/lyazcat.py
```diff
@@ -0,0 +1,441 @@
...
+    Notes
+    -----
+    - There will be many here.
+    """
+    # EBL Load the SQUEzE Model file first. This is a very large file,
+    # so if multiple petals are to be processed, we only want to load
+    # it into memory once.
+    if isinstance(squeze_model, str) and sq_flag:
+        tmark('    Loading SQUEzE Model file')
+        squeze_model = Model.from_json(load_json(squeze_model))
```
geordie666, line 355: Maybe have this load at the top of the code, so it's only run once at import and is available throughout the module as a global. So, call it `squeze_model` and load it after your initial import statements.

py/desitarget/lyazcat.py
```diff
@@ -0,0 +1,441 @@
...
+        fzcat = zcomb_selector(zcat)
+    
+        full_outputname = zcat_writer(outputdir,fzcat,outputname, qn_flag,
+                                      sq_flag, abs_flag)
+    
+        tmark('--{} written out correctly.'.format(full_outputname))
+        print('===================================')
+
+    
+if __name__=='__main__':
```
geordie666, line 390: The formalism in `desitarget` is to have separate scripts in `bin` instead of using `if __name__=='__main__'` in the module itself. See the many examples in [the bin directory](https://github.com/desihub/desitarget/tree/master/bin).

py/desitarget/lyazcat.py
```diff
@@ -0,0 +1,441 @@
...
+    Returns
+    -------
+    :class:'~numpy.array'
+        The zcat array with QuasarNP data included in the columns:
+        * Z_QN
+        * Z_QN_CONF
+        * IS_QSO_QN
+    """
+    tmark('    Adding QuasarNP data')    
+    lines = ['LYA','CIV(1548)','CIII(1909)', 'MgII(2796)','Hbeta','Halpha']
```
geordie666, line 119: Code style. Need a white-space after every comma in a list.

py/desitarget/lyazcat.py
```diff
@@ -0,0 +1,441 @@
...
+zcatdatamodel_formats = np.array(['>int64','>float64','>int64','>float64',
+                                  '>float64','>float64','>float64','>float64',
+                                  '>int16','>float64','>float64','>float64',
+                                  '>float64'])
+"""
+zcatdatamodel_formats = np.array(['>i8','>f8','>i8','>f8',
+                                  '>f8','>f8','>f8','>f8',
+                                  '>i2','>f8','>f8','>f8',
+                                  '>f8'])
+zcols_copy = np.array(['TARGETID','Z','ZWARN','DELTACHI2'])
```
geordie666, line 46: Code style. Need a white-space after every comma in a list.

py/desitarget/lyazcat.py
```diff
@@ -0,0 +1,441 @@
...
+n1_cols = np.array(['Z_QN','Z_QN_CONF','IS_QSO_QN',
+                    'Z_SQ','Z_SQ_CONF','Z_ABS','Z_ABS_CONF'])
```
geordie666, line 48: Code style. Need a white-space after every comma in a list.

py/desitarget/lyazcat.py
```diff
@@ -0,0 +1,441 @@
...
+from squeze.common_functions import save_json, load_json
+from squeze.candidates import Candidates
+from squeze.desi_spectrum import DesiSpectrum
+from squeze.spectra import Spectra
+
+# ADM set up the DESI default logger.
+from desiutil.log import get_logger
+log = get_logger()
+
+zcatdatamodel_names = np.array(['TARGETID','Z','ZWARN','DELTACHI2','Z_COMB',
```
geordie666, line 33: Code style. Need a white-space after every comma in a list.

py/desitarget/lyazcat.py
```diff
@@ -0,0 +1,441 @@
...
+                                'Z_COMB_PROB','Z_QN','Z_QN_CONF','IS_QSO_QN',
+                                'Z_SQ','Z_SQ_CONF','Z_ABS','Z_ABS_CONF'])
+"""
+zcatdatamodel_formats = np.array(['>int64','>float64','>int64','>float64',
```
geordie666, line 37: Code style. Need a white-space after every comma in a list.

py/desitarget/lyazcat.py
```diff
@@ -0,0 +1,441 @@
...
+                                  '>float64','>float64','>float64','>float64',
+                                  '>int16','>float64','>float64','>float64',
+                                  '>float64'])
+"""
+zcatdatamodel_formats = np.array(['>i8','>f8','>i8','>f8',
```
geordie666, line 42: Code style. Need a white-space after every comma in a list.

py/desitarget/lyazcat.py
```diff
@@ -0,0 +1,441 @@
...
+    -------
+    :class:'~numpy.array'
+        The zcat array with SQUEzE data included in the columns:
+        * Z_SQ
+        * Z_SQ_CONF
+    """
+    tmark('    Adding SQUEzE data')
+    mdata = ["TARGETID"]
+    single_exposure = False
+    sq_cols_keep = ["PROB","Z_TRY","TARGETID"]
```
geordie666, line 162: Code style. Need a white-space after every comma in a list.

py/desitarget/lyazcat.py
```diff
@@ -0,0 +1,441 @@
...
+        tmark('--{} written out correctly.'.format(full_outputname))
+        print('===================================')
+
+    
+if __name__=='__main__':
+    # TODO:  Need to pass it TILEID, NIGHTID, PETAL_NUM
+    # TODO:  Check data precision and match from redrock (greater than/less than are important for Big-endian, little-endian)
+    # TODO:  Check data precision and match from QN and SQ
+    # TODO:  FIX MY DOC STRINGS AND COMMENTS AND TERMINAL FEEDBACK
+    # TODO:  Figure out how I screwed up PEP8 somewhere.
```
geordie666, line 395: Don't worry too much about code style. I can always fix that later, and it's an allowed failure. although your main transgression appears to be just that you haven't include a white-space after every comma in a list. I've identified those where I found them, in case it's gnawing at you.

geordie666 (commented):
I've added some initial comments just scanning through the code. I may have more comments when I've actually tried to run the code, though.

In general, this is fantastic work. You've hewed closely to the style and formalism used elsewhere in `desitarget`, which I really appreciate.

py/desitarget/lyazcat.py
eleanorlyke: It would be better to leave this here, I think. If it loads at import, every time the script is called it will load whether the user wants to use SQUEzE or not. If the user chooses not to use SQUEzE that is a ~6-8 min. time sink for something unused (and a few GB of memory it's carrying around per instance).

py/desitarget/lyazcat.py
eleanorlyke: Fixed for the next push.

py/desitarget/lyazcat.py
eleanorlyke: Fixed for the next push.

py/desitarget/lyazcat.py
eleanorlyke: Fixed for the next push.

py/desitarget/lyazcat.py
eleanorlyke: Fixed for the next push.

py/desitarget/lyazcat.py
eleanorlyke: Fixed for the next push.

py/desitarget/lyazcat.py
geordie666: True, let's leave this load where you have it for now. Depending on whether we definitely choose to use SQUEzE, and how we parelellize, we may want to choose, later on, to load this elsewhere.

py/desitarget/lyazcat.py
eleanorlyke: Fixed for the next push.

py/desitarget/lyazcat.py
eleanorlyke: Fixed for the next push.

py/desitarget/lyazcat.py
eleanorlyke: I forgot the TODO list was still in there. These are mostly done. I've fixed the whitespace/comma issue and standardized the single-quote/double-quote usage throughout.

py/desitarget/lyazcat.py
eleanorlyke: I've set this up to run the same way SQUEzE does, by loading early, for now. We can find a better place to load this once we start figuring out parallelization (and can make the same changes we do to SQUEzE).

py/desitarget/lyazcat.py
eleanorlyke: I'm using my version, for now, because it prints out date and time codes so I can see how long each step takes. Once the script is nailed down I'll remove this function (and all calls for it) and convert to the DESI logger.

geordie666: If you just replace `print` with `log.info`, your bespoke function should still work. So, there's no need to remove the function and the calls to it. Just switch over to logging instead of printing.

py/desitarget/lyazcat.py
eleanorlyke: I haven't done it that way before, so I'll look at that tonight to make the changeover when I add the MgII absorption functionality.

py/desitarget/lyazcat.py
eleanorlyke: I'll check this tonight when I make the MgII absorption additions and the logger changeover.

py/desitarget/lyazcat.py
```diff
@@ -0,0 +1,463 @@
...
+from desitarget.geomask import match, match_to
+from desitarget.internal import sharedmem
+from desitarget import io
+from desispec.io import read_spectra
+
+from quasarnp.io import load_model
+from quasarnp.io import load_desi_coadd
+from quasarnp.utils import process_preds
+from squeze.model import Model
+from squeze.common_functions import save_json, load_json
```
geordie666, line 24: I don't think you actually use `save_json` anywhere in the code, so there shouldn't be any need to import it?

py/desitarget/lyazcat.py
```diff
@@ -0,0 +1,463 @@
...
+    absorption redshifts and confidence values.
+    
+    Notes
+    -----
+    - There will be many here.
+    """
+    # EBL Load the SQUEzE Model file first. This is a very large file,
+    # so if multiple petals are to be processed, we only want to load
+    # it into memory once.
+    if isintance(qnp_model, str) and qn_flag:
```
geordie666, line 362: This should be `isinstance`.

py/desitarget/lyazcat.py
```diff
@@ -0,0 +1,463 @@
...
+
+    return full_outputname
+
+
+def create_zcat(tile, night, petal_num,
+                zcatdir='/global/cfs/cdirs/desi/spectro/redux/daily', 
+                outputdir='/global/cfs/cdirs/desi/spectro/redux/daily',
+                qn_flag=False, sq_flag=False, abs_flag=False,
+                qnp_model='/global/cfs/cdirs/desi/science/lya/qn_models/boss_dr12/qn_train_coadd_indtrain_0_0_boss10.h5',
+                squeze_model='/global/homes/e/ebie/local/SQUEzE/data/BOSS_train_64plates_model.json',
```
geordie666, line 310: Use paths that can be shared by anyone at NERSC, not in your local home directory. I recommend making new spaces in `/global/cfs/cdirs/desi/target/catalogs/lya/qn_models` and `/global/cfs/cdirs/desi/target/catalogs/lya/sq_models`.

py/desitarget/lyazcat.py
```diff
@@ -0,0 +1,463 @@
...
+    # it into memory once.
+    if isintance(qnp_model, str) and qn_flag:
+        tmark('    Loading QuasarNP Model file')
+        qnp_lines = ['LYA', 'CIV(1548)', 'CIII(1909)', 'MgII(2796)', 'Hbeta', 'Halpha']
+        qnp_lines_bal = ['CIV(1548)']
+        qnp_model = load_model('/global/cfs/cdirs/desi/science/lya/qn_models/boss_dr12/qn_train_coadd_indtrain_0_0_boss10.h5')
+        tmark('      QNP model file loaded')
+    if isinstance(squeze_model, str) and sq_flag:
+        tmark('    Loading SQUEzE Model file')
+        squeze_model = Model.from_json(load_json(squeze_model))
```
geordie666, line 370: Let's create a separate `load_sq_model()` function instead of this complicated construction. In fact let's:

- Have separate functions for `load_sq_model(filename=X)` and `load_qn_model(filename=Y)`, where `X` and `Y` default to the standard locations at NERSC.
- Write `create_zcat` to always take the loaded models as inputs rather than either a string or a loaded model.

Loading the SQUEzE model is likely to be a critical bottleneck, so let's have single, transparent functions to load the models and not embed the model-loading in the `create_zcat()` function, even if that loading is just an option.

geordie666 (commented):
I tried running the code myself, and it generally works well. I appreciate that we're likely waiting for https://github.com/desihub/QuasarNP/pull/4 to be resolved before QN will be fully integrated. So, in the meantime, I've added a few more comments for you to address in addition to the two outstanding comments.

py/desitarget/lyazcat.py
eleanorlyke: Correct. Thanks for catching it. I removed the import.

py/desitarget/lyazcat.py
eleanorlyke: I fixed this in the version I have. I'll push.

py/desitarget/lyazcat.py
eleanorlyke: Perfect. Exactly what I was looking for. I copied them over and changed the filename calls.

py/desitarget/lyazcat.py
eleanorlyke: Switched to using the logger.

py/desitarget/lyazcat.py
eleanorlyke: @geordie666 Model filenames are generated by separate functions now and use environment variables. Model files are loaded by their own convenience functions.

py/desitarget/lyazcat.py
eleanorlyke: @geordie666 I switched this over to using the bin script method.

eleanorlyke:
MgII Absorption and comparisons between QuasarNP, SQUEzE, MgII absorption, and redrock have yet to be added. I'll start on MgII absorption next.

geordie666:
@eleanorlyke: Can we merge this to make continued progress and you can continue development in another branch?

eleanorlyke:
> @eleanorlyke: Can we merge this to make continued progress and you can continue development in another branch?

I haven't had a chance to add the MgII absorption yet, but if you like to merge this branch and I can make a new one for adding that functionality.

geordie666:
Sounds good, I think I'll merge it as its already a significant amount of new work.

PR #716, opened 2021-04-28
Explicitly permit SCND bits that can update match MWS primaries

apcooper:
Explicitly permit secondary bits that can change the state of matched MWS primary targets.

Previous behaviour was to allow such updates to be driven by any target with SCND_ANY. Now checks for `mwsupdate=True` flag in target bit definition.


sbailey:
Thanks @apcooper .  Please update the variable names and comments to avoid "whitelist" (and implied "blacklist") terminology.  allow / permit / include / override are possible alternatives.

geordie666:
@apcooper: Just so you know, I took the updates you made in the bit-mask yaml file in this branch and ported them to the Main Survey bit-mask yaml file in my working branch. So, you won't need to also do that in this PR. In other words, I have retained your `mwsupdate=True/False` choices for the Main Survey in my working branch.

apcooper:
@geordie666 OK thanks for doing that. You say I won't need to do that -- I've done it already, so do I need to undo anything?

@sbailey noted, will fix.

geordie666:
Yeah, if you could remove any changes in `py/desitarget/data/targetmask.yaml` that'd be great. I already have those changes in my working branch (and I have a lot more additional changes in my branch). It might be easier for me to merge the branches if you just undo anything in `py/desitarget/data/targetmask.yaml`. Thanks!

apcooper:
@geordie666  I think this is working -- I've tested it a bit and have corresponding unit tests, but some of them assume things are set up as expected in the `targetmask.yaml`, which is not the case now per your request. I can add these later once everything is merged, if needed.

The result of this PR is that the previous behavior is recovered if and only if updatemws=True.

@sbailey I've changed the terminology.

apcooper:
The last commit seems harmless and allows the tests to pass in without corresponding changes to `targetmask.yaml`, but I'm not sure if it fits with the style for the other flavor-like flags.

py/desitarget/secondary.py
```diff
@@ -576,20 +576,42 @@ def match_secondary(primtargs, scxdir, scndout, sep=1.,
...
+    scnd_update = targs[desicols[0]] == desi_mask['SCND_ANY']
+    log.info('{} scnd targets will have initial state set as secondary-only'.format(scnd_update.sum()))
+
+    # APC The exception to the rule above is that a subset of bits flagged with
+    # APC updatemws=True in the targetmask can drive initial state for a subset of
+    # APC primary bits corresponding to MWS targets and standards. We first create
+    # APC a bitmask of those permitted seconday bits.
+    permit_scnd_bits = 0
+    for name in scnd_mask.names():
+        if hasattr(scnd_mask[name], 'updatemws'):
```
geordie666, line 591: I'd rather this failed if the needed attribute wasn't present, to ensure we always remember to include the `updatemws` attribute. But, I understand that if you remove this unit tests will fail. I suggest we proceed as follows:

- You leave the `hasattr` tests in for #716.
- I merge #718 before you merge #719 (preferred anyway, as it's easier to merge a smaller PR into a larger PR).
- You remove the `hasattr` tests in #719.

Then I think everything will work.

py/desitarget/secondary.py
```diff
@@ -576,20 +576,42 @@ def match_secondary(primtargs, scxdir, scndout, sep=1.,
...
+    # APC they have NO primary target bits set (hence == on next line).
+    scnd_update = targs[desicols[0]] == desi_mask['SCND_ANY']
+    log.info('{} scnd targets will have initial state set as secondary-only'.format(scnd_update.sum()))
+
+    # APC The exception to the rule above is that a subset of bits flagged with
+    # APC updatemws=True in the targetmask can drive initial state for a subset of
+    # APC primary bits corresponding to MWS targets and standards. We first create
+    # APC a bitmask of those permitted seconday bits.
+    permit_scnd_bits = 0
+    for name in scnd_mask.names():
```
geordie666, line 590: Is this backwards compatible? For instance, if we needed to run the `sv3` version of MTL again, will this fail because the `sv3` bit-masks have no concept of the `updatemws` attribute?

It seems to me that you might need an `if survey=="main"` clause somewhere?

geordie666 (commented):
Thanks @apcooper. This looks like carefully written code, as always. I checked a few basic use cases and it seems to work fine.

I have a couple of comments regarding backwards compatibility and the tangled web of PRs we've been creating. But, feel free to merge once those comments are addressed.

py/desitarget/secondary.py
apcooper: OK, will do that.

py/desitarget/secondary.py
apcooper: I've added logic to only use `updatemws` if `survey==main`. Otherwise all bits are permitted to drive updates to MWS targets as before this PR.

geordie666:
Looks good! Feel free to merge once tests pass.

PR #717, opened 2021-04-30
Added MgII Absorption

eleanorlyke:
I added the functionality for MgII Absorption and cleaned up the bin file.


geordie666:
Looks good, thanks.

PR #719, opened 2021-05-02
MWS Faint

apcooper:
Following discussion on the desisurvey list, this PR implements in the main survey a fainter extension of the MWS sample almost the same as MWS_FAINT in SV1, magnitude limited `19 < r < 20`.

The small difference with SV1 is the addition of a parallax cut for stars with Gaia measurements (`paramssolved == 31`) and a slightly fainter hard limit of `r_obs = 20.5`.

The MWS-FAINT cuts for the main survey are documented on the wiki:
https://desi.lbl.gov/trac/wiki/TargetSelectionWG/TargetSelection

MWS will sign off on this on our Monday call on the 3rd May, then it should be ready to merge.


py/desitarget/data/targetmask.yaml
```diff
@@ -273,6 +276,9 @@ priorities:
         MWS_MAIN_RED:                 SAME_AS_MWS_MAIN_BLUE
         MWS_MAIN_RED_NORTH:           SAME_AS_MWS_BROAD_NORTH
         MWS_MAIN_RED_SOUTH:           SAME_AS_MWS_BROAD_NORTH
+        MWS_FAINT:                    {UNOBS: 1300, MORE_ZWARN: 2, MORE_ZGOOD: 2, DONE: 2, OBS: 1, DONOTOBSERVE: 0, MORE_MIDZQSO: 0}
```
geordie666, line 279: I believe that @djschlegel suggested that these should enter at the lowest priority. Even lower than all of the secondary targets. so, I believe the `UNOBS` priority here needs to be more like `50`.

apcooper: I set UNOBS=50 for MWS-FAINT, but then tests fail on a check that MORE_ZGOOD for BGS (=1000) is always higher than the minimum priority for an unobserved MW target (good test!).

I could update the test to permit this case (i.e. BGS reobservations ahead of MWS-FAINT), or alternatively lower the BGS MORE_ZGOOD. Someone would have to approve the second option.

I guess my biased opinion is that, given the tiny number and patchy distribution of targets we're talking about, unobserved MWS-FAINT should be preferred to BGS reobservations. 

There's also the question of whether BGS reobservations should be higher priority than secondary targets, and whether MORE_ZGOOD=1000 matches whatever the decision is about that. 

Maybe these are questions for @djschlegel?

geordie666: For now, just update the test to except `MWS_FAINT`. I think the BGS group is tending towards _not_ having a high priority for `MORE_ZGOOD` observations. My guess is that `MORE_ZGOOD` will end up as the `DONE` priority for all target classes, as was the case in `sv3`.

apcooper: OK, done.

apcooper:
Ready to merge once tests pass, I think. Will hold off until this has been discussed at the MWS telecon today (5/3).

geordie666:
If you could also hold off until I've merged #718, that'd be great. And, after I merge #718, also remove the `hasattr` tests you added to MTL and check that the unit tests then pass.

apcooper:
Discussion in MWS today brought up requests for a few tweaks, which I will get done in the next ~day, specifically:

- for bookkeeping, separate the FAINT sample into equal-priority BLUE and RED samples as for the MAIN selection.
- apply slightly different PM and plx cuts on the red side only.

I'll also remove the `hasattr`.

apcooper:
Should be ready to go if tests pass. I'll raise a separate PR regarding DR2->EDR3 changes for MWS cuts tomorrow.

apcooper:
@geordie666 please merge this or tell me when I can go ahead.

geordie666:
I'm in meetings for the next few hours. I _think_ everything looks good, but give me a couple of hours to check and then I'll merge it myself.

PR #720, opened 2021-05-03
MWS Gaia DR2->EDR3

apcooper:
Changes in response to switching MWS targeting to EDR3.

Including:
- Update AEN threshold to 2 from 3
- Lower error floor in RED parallax cut to 0.5 from 1 mas

Some final checks are in progress in MWS to validate these cuts on the EDR3 sweeps. Timeframe is 1-2 days.

Will need to update FAINT cuts in #719 once merged.

### DIFF
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -182,7 +182,7 @@ def isGAIA_STD(ra=None, dec=None, galb=None, gaiaaen=None, pmra=None, pmdec=None
                parallax=None, parallaxovererror=None, ebv=None, gaiabprpfactor=None,
                gaiasigma5dmax=None, gaiagmag=None, gaiabmag=None, gaiarmag=None,
                gaiadupsource=None, gaiaparamssolved=None,
-               primary=None, test=False, nside=2, dr="dr2"):
+               primary=None, test=False, nside=2):
     """Standards based solely on Gaia data.
 
     Parameters
@@ -195,8 +195,6 @@ def isGAIA_STD(ra=None, dec=None, galb=None, gaiaaen=None, pmra=None, pmdec=None
     nside : :class:`int`, optional, defaults to 2
         (NESTED) HEALPix nside, if targets are being parallelized.
         The default of 2 should be benign for serial processing.
-    dr : :class:`str`, optional, defaults to "dr2"
-        Name of a Gaia data release. Options are "dr2", "edr3".
 
     Returns
     -------
@@ -239,7 +237,7 @@ def isGAIA_STD(ra=None, dec=None, galb=None, gaiaaen=None, pmra=None, pmdec=None
                       gaiagmag=gaiagmag, gaiabmag=gaiabmag, gaiarmag=gaiarmag)
 
     # ADM restrict to point sources.
-    ispsf = gaia_psflike(gaiaaen, gaiagmag, dr=dr)
+    ispsf = gaia_psflike(gaiaaen, gaiagmag)
     std &= ispsf
 
     # ADM de-extinct the magnitudes before applying color cuts.
@@ -2833,7 +2831,6 @@ def apply_cuts_gaia(numproc=4, survey='main', nside=None, pixlist=None,
 
     # ADM determine if a target is a Gaia-only standard.
     primary = np.ones_like(gaiaobjs, dtype=bool)
-
     std_faint, std_bright, std_wd = targcuts.isGAIA_STD(
         ra=ra, dec=dec, galb=galb, gaiaaen=gaiaaen, pmra=pmra, pmdec=pmdec,
         parallax=parallax, parallaxovererror=parallaxovererror, ebv=ebv,
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -970,7 +970,7 @@ def isMWS_main_colors(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=No
     # ADM main targets are point-like based on DECaLS morphology
     # ADM and GAIA_ASTROMETRIC_NOISE.
     mws &= _psflike(objtype)
-    mws &= gaiaaen < 3.0
+    mws &= gaiaaen < 2.0
 
     # ADM main targets are 16 <= r < 19
     mws &= rflux > 10**((22.5-19.0)/2.5)
@@ -1136,8 +1136,8 @@ def isMWS_bhb(primary=None, objtype=None,
 
     # APC no gaia duplicated sources.
     mws &= ~gaiadupsource
-    # APC gaia astrometric excess noise < 3.
-    mws &= gaiaaen < 3.0
+    # APC gaia astrometric excess noise < 2.
+    mws &= gaiaaen < 2.0
 
     # APC BHB extinction-corrected color range -0.35 <= gmr <= -0.02.
     mws &= (gmr >= -0.35) & (gmr <= -0.02)
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -182,7 +182,7 @@ def isGAIA_STD(ra=None, dec=None, galb=None, gaiaaen=None, pmra=None, pmdec=None
                parallax=None, parallaxovererror=None, ebv=None, gaiabprpfactor=None,
                gaiasigma5dmax=None, gaiagmag=None, gaiabmag=None, gaiarmag=None,
                gaiadupsource=None, gaiaparamssolved=None,
-               primary=None, test=False, nside=2):
+               primary=None, test=False, nside=2, dr="dr2"):
     """Standards based solely on Gaia data.
 
     Parameters
@@ -195,6 +195,8 @@ def isGAIA_STD(ra=None, dec=None, galb=None, gaiaaen=None, pmra=None, pmdec=None
     nside : :class:`int`, optional, defaults to 2
         (NESTED) HEALPix nside, if targets are being parallelized.
         The default of 2 should be benign for serial processing.
+    dr : :class:`str`, optional, defaults to "dr2"
+        Name of a Gaia data release. Options are "dr2", "edr3".
 
     Returns
     -------
@@ -237,7 +239,7 @@ def isGAIA_STD(ra=None, dec=None, galb=None, gaiaaen=None, pmra=None, pmdec=None
                       gaiagmag=gaiagmag, gaiabmag=gaiabmag, gaiarmag=gaiarmag)
 
     # ADM restrict to point sources.
-    ispsf = gaia_psflike(gaiaaen, gaiagmag)
+    ispsf = gaia_psflike(gaiaaen, gaiagmag, dr=dr)
     std &= ispsf
 
     # ADM de-extinct the magnitudes before applying color cuts.
@@ -2787,6 +2789,9 @@ def apply_cuts_gaia(numproc=4, survey='main', nside=None, pixlist=None,
     if survey == 'main':
         import desitarget.cuts as targcuts
         from desitarget.targetmask import desi_mask, mws_mask
+
+        # APC Use Gaia EDR3 in main survey
+        dr = "edr3"
     elif survey[:2] == 'sv':
         try:
             targcuts = import_module("desitarget.{}.{}_cuts".format(survey,
@@ -2799,6 +2804,9 @@ def apply_cuts_gaia(numproc=4, survey='main', nside=None, pixlist=None,
             log.critical(msg)
             raise ModuleNotFoundError(msg)
         desi_mask, mws_mask = targmask.desi_mask, targmask.mws_mask
+
+        # We used DR2 in SV
+        dr = "dr2"
     else:
         msg = "survey must be either 'main'or 'svX', not {}!!!".format(survey)
         log.critical(msg)
@@ -2837,7 +2845,7 @@ def apply_cuts_gaia(numproc=4, survey='main', nside=None, pixlist=None,
         gaiabprpfactor=gaiabprpfactor, gaiasigma5dmax=gaiasigma5dmax,
         gaiagmag=gaiagmag, gaiabmag=gaiabmag, gaiarmag=gaiarmag,
         gaiadupsource=gaiadupsource, gaiaparamssolved=gaiaparamssolved,
-        primary=primary, nside=nside, test=test)
+        primary=primary, nside=nside, test=test, dr=dr)
 
     # ADM Construct the target flag bits.
     mws_target = backup_bright * mws_mask.BACKUP_BRIGHT
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -672,7 +672,7 @@ def isSTD_gaia(primary=None, gaia=None, astrometricexcessnoise=None,
 
     # ADM no obvious issues with the astrometry solution.
     std &= astrometricexcessnoise < 1
-    std &= paramssolved == 31
+    std &= paramssolved >= 31
 
     # ADM finite proper motions.
     std &= np.isfinite(pmra)
@@ -913,7 +913,7 @@ def isMWS_faint_colors(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=N
     # APC main targets are point-like based on DECaLS morphology
     # APC and GAIA_ASTROMETRIC_NOISE.
     faint &= _psflike(objtype)
-    faint &= gaiaaen < 3.0
+    faint &= gaiaaen < 2.0
 
     # APC faint targets are 19 <= r < 20
     faint &= rflux > 10**((22.5-20.0)/2.5)
@@ -935,7 +935,7 @@ def isMWS_faint_colors(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=N
     # ADM Turn off any NaNs for astrometric quantities to suppress
     # ADM warnings. Won't target these, using cuts on paramssolved
     # ADM (or will explicitly target them based on paramsssolved).
-    ii = paramssolved != 31
+    ii = paramssolved < 31
     parallax = parallax.copy()
     parallax[ii] = 0.
 
@@ -945,9 +945,9 @@ def isMWS_faint_colors(gflux=None, rflux=None, zflux=None, w1flux=None, w2flux=N
     # APC MWS-FAINT-RED also has parallax < max(3parallax_err,1) mas
     # APC and proper motion < 3 (lower than MAIN-MAIN-RED)
     # ACP and all astrometric parameters are measured.
-    faint_red &= parallax < np.maximum(3*parallaxerr, 1)
+    faint_red &= parallax < (3 * parallaxerr + 0.3)
     faint_red &= pm < 3.
-    faint_red &= paramssolved == 31
+    faint_red &= paramssolved >= 31
 
     # APC There is no equivalent of MWS-MAIN-BROAD for the faint extension, any
     # APC stars failing the astrometry cuts or without the Gaia parameters are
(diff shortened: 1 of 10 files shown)


geordie666:
Note that we only update parallax and proper motion to EDR3. The AEN still remains Gaia DR2. I think we could also update the AEN to EDR3. But, we don't currently. Updating the AEN would require an additional change to this function:

https://github.com/desihub/desitarget/blob/8e3b3d6d23cbbfea51693ee2c7072473415901bf/py/desitarget/gaiamatch.py#L195

(at least, I think that's the meat of where I decided to make the DR2->EDR3 substitutions in the sweeps).

As a rule, we should only allows updates to parameters to EDR3 if they are only used by the MWS (and standard star) selections.

geordie666:
Also, I was hedging my bets on whether we would make the DR2->EDR3 updates. If we're definitely going to do that, we might want to switch the sense of this:

https://github.com/desihub/desitarget/blob/8e3b3d6d23cbbfea51693ee2c7072473415901bf/bin/select_targets#L43-L44

So that the default is `gaiasub` (for the Main Survey) and `nogaiasub` would need to be passed to turn that behavior _off_.

apcooper:
Thanks for the comments, @geordie666.

> Note that we only update parallax and proper motion to EDR3. The AEN still remains Gaia DR2. I think we could also update the AEN to EDR3. But, we don't currently. 

I feel we should, but @segasai's opinion would be welcome on how important/useful this is.

> As a rule, we should only allows updates to parameters to EDR3 if they are only used by the MWS (and standard star) selections.

I can't see it being used anywhere except these cases. BGS have a different star/galaxy separation.

apcooper:
Also, just noticed that AEN is used in the WD selection, so we should double check with Boris that this change is not significant there.

apcooper:
@segasai and @geordie666 might also want to consider whether we need to update this function with EDR3 cuts if AEN is updated to EDR3:

https://github.com/desihub/desitarget/blob/8e3b3d6d23cbbfea51693ee2c7072473415901bf/py/desitarget/gaiamatch.py#L167-L168

geordie666:
If we do update the `gaia_psflike` function to have different cuts for `edr3`, we will need to retain backwards compatibility for any calls in other codebases that use it. For example, with a `dr="dr2"` kwarg that, if passed as `edr3` uses the different set of cuts.

BGaensicke:
We were planning to use for the WD selection phot_bp_rp_excess_factor and astrometric_sigma5d_max: 

```
    # ADM As of DR7, photbprpexcessfactor and astrometricsigma5dmax are not in the
    # ADM imaging catalogs. Until they are, ignore these cuts
    if photbprpexcessfactor is not None:
        # ADM remove problem objects, which often have bad astrometry
        mws &= photbprpexcessfactor < 1.7 + 0.06*br*br

    if astrometricsigma5dmax is not None:
        # ADM Reject white dwarfs that have really poor astrometry while
        # ADM retaining white dwarfs that only have relatively poor astrometry
        mws &= ((astrometricsigma5dmax < 1.5) |
                ((astrometricexcessnoise < 1.) & (parallaxovererror > 4.) & (pm > 10.)))
```

but I'm not sure if those parameters were available in the DR2 sweeps, in other words, we may have run the selection without those? If that's the case, then adding them now would probably cause much pain and we should just drop those statements.

geordie666:
@BGaensicke: We definitely use those. The "primary" targets are run on the Legacy Surveys sweeps files, not the Gaia files. Those columns are [included in the Legacy Surveys sweeps files](https://www.legacysurvey.org/dr9/files/#sweep-catalogs-region-sweep).

BGaensicke:
If we do update AEN to EDR3, we should also update PHOT_BP_RP_EXCESS_FACTOR and ASTROMETRIC_SIGMA5D_MAX, so that the selection runs on a consistent set of Gaia data. The WD selection algorithm would remain the same, just using the updated data.

geordie666:
`PHOT_BP_RP_EXCESS_FACTOR` is used by other selections (most notably the standard stars). So, if it's critical to have consistency in the WD selection then we _shouldn't_ update AEN either.

BGaensicke:
OK, I think I see: the Gaia photometry will remain DR2, right? 

I think that the biggest impact that EDR3 has is in the parallaxes and proper motions, updating  PHOT_BP_RP_EXCESS_FACTOR and ASTROMETRIC_SIGMA5D_MAX will only result in subtle changes, so if that causes headaches with other target classes, sticking to DR2 for those two parameters is fine.

geordie666:
Thanks @BGaensicke. And, yes, the Gaia photometry will all remain DR2.

segasai:
I looked at the code, and I think we must use the astrometric_params_solved from EDR3 and we need replace the astrometric_nparams==31 by >=31 (because it can have values of 3, 31, 95).
Regarding AEN, we are good with AEN<2 new cut. and it's unclear if we need to change gaia_psflike() as I don't see it being called by any target selection code.
I'll send info on new plx,pm selection

geordie666:
Thanks for your checks Sergey. It would be great if we don't update `gaia_psflike()`. It _is_ called by functions outside of the main target selection code (notably to select GFAs).

Great catch on `astrometric_params_solved`. I agree that will need to be updated to `>=31` everywhere it's used.

segasai:
Sure. I've created a different  branch ( I didn't know if I should tinker with this one or not)
https://github.com/desihub/desitarget/commits/mws_edr3x
https://github.com/desihub/desitarget/commit/b05985d073478fd25b2b9c76eae6a92be0ddca8a
with the 
* Andrew's AEN change, 
* the nparams fix.
* Update to the MW plx cut. 
I intend to apply a final tweak to the MWS proper motion cut as well tomorrow (or keep it as it is)

geordie666:
@segasai: It's Andrew's branch so it's up to him whether he wants to let you tinker with it directly. I'm fine with you doing that if @apcooper signs off on direct tinkering, though.

apcooper:
@segasai, I'm teaching a class now (!)so I'll look at this again later tonight but please go ahead and tinker with this branch!

apcooper:
OK, so @segasai, if you can please merge your branch with this one after you've made the decision about the proper motion cut, then I'll take care of any loose ends like `gaia_psflike`, update the target selection page on the wiki and do the final merge.

Regarding `gaia_psflike`, it seems to be used in selecting standards and setting the `TYPE` column for the gfa targets. The question is whether the cut of 3 in `gaia_psflike` would be harmful in those cases if we apply it to the EDR3 AEN values. I'm leaning towards Adam's idea of having a switch and changing 3 to 2 in the EDR3 case, rather than using `AEN < 3` for all Gaia releases. 

However, having read the above discussion I'm still unclear what if anything changes in the WD selection and which columns are EDR3 and which are DR2. I *think* `PHOT_BP_RP_EXCESS_FACTOR`, `ASTROMETRIC_SIGMA5D_MAX` and the photometry are staying as DR2, but AEN will be EDR3.

BGaensicke:
>
> However, having read the above discussion I'm still unclear what if
> anything changes in the WD selection and which columns are EDR3 and which
> are DR2. I *think* PHOT_BP_RP_EXCESS_FACTOR, ASTROMETRIC_SIGMA5D_MAX and
> the photometry are staying as DR2, but AEN will be EDR3.
>
yes, that was the conclusion.

segasai:
Sorry, it took me a while. I've updated the mws_edr3x branch with the new proper motion cut (as in the desi-mw email) and ensured that the NPARAMS are substituted as well from EDR3  
I can't merge my commits unfortunately here, because this branch does not include the commits from 
#719 while my branch was split after #719 landed. So either this branch need to be rebased, or we'll just work with mws_edr3x.

apcooper:
OK I have a bit of time now so I'll rebase this branch and merge yours, Sergey.

apcooper:
I've now merged @segasai 's branch with mine. 

I've also added the dr2/edr3 switch in `gaia_psflike` as discussed above, with default of dr2 for backwards compatibility. This is used in `isSTD_gaia`, which I've updated to use edr3 for the main survey only.

I'm not sure what to do about the one other place `gaia_psflike` is used, in the GFA targets:

https://github.com/desihub/desitarget/blob/05db4981cbd9ed918924dd7d2d97b91290ec5f8f/py/desitarget/gfa.py#L56-L57

called from

https://github.com/desihub/desitarget/blob/05db4981cbd9ed918924dd7d2d97b91290ec5f8f/py/desitarget/gfa.py#L244-L245

in
https://github.com/desihub/desitarget/blob/05db4981cbd9ed918924dd7d2d97b91290ec5f8f/py/desitarget/gfa.py#L161-L163

Either (1) the DR comes down through the stack above this, (2) it gets deduced somehow from the files that are being read here, or (3) we just ignore this one and leave it as DR2. @geordie666?

Finally, there are two other places AEN is used, perhaps we should double check. In `isSTD_gaia`:

https://github.com/desihub/desitarget/blob/e49c5b6640f8ab63447dc043975072746e38c733/py/desitarget/cuts.py#L675-L677

and in the WD selection we discussed above.

https://github.com/desihub/desitarget/blob/e49c5b6640f8ab63447dc043975072746e38c733/py/desitarget/cuts.py#L1250-L1254

These look fine, though they use a stricter cut than we have for the main sample.

If all that is ok, it seems like this branch is ready.

py/desitarget/cuts.py
```diff
@@ -2789,13 +2839,25 @@ def apply_cuts_gaia(numproc=4, survey='main', nside=None, pixlist=None,
...
     # ADM determine if a target is a Gaia-only standard.
     primary = np.ones_like(gaiaobjs, dtype=bool)
-    std_faint, std_bright, std_wd = targcuts.isGAIA_STD(
-        ra=ra, dec=dec, galb=galb, gaiaaen=gaiaaen, pmra=pmra, pmdec=pmdec,
-        parallax=parallax, parallaxovererror=parallaxovererror, ebv=ebv,
-        gaiabprpfactor=gaiabprpfactor, gaiasigma5dmax=gaiasigma5dmax,
-        gaiagmag=gaiagmag, gaiabmag=gaiabmag, gaiarmag=gaiarmag,
-        gaiadupsource=gaiadupsource, gaiaparamssolved=gaiaparamssolved,
-        primary=primary, nside=nside, test=test)
+
```
geordie666, line 2842: Revert all changes in `apply_cuts_gaia()`.

py/desitarget/cuts.py
```diff
@@ -2757,6 +2804,9 @@ def apply_cuts_gaia(numproc=4, survey='main', nside=None, pixlist=None,
             log.critical(msg)
             raise ModuleNotFoundError(msg)
         desi_mask, mws_mask = targmask.desi_mask, targmask.mws_mask
+
```
geordie666, line 2807: Revert all changes in `apply_cuts_gaia()`.

py/desitarget/cuts.py
```diff
@@ -182,7 +182,7 @@ def isGAIA_STD(ra=None, dec=None, galb=None, gaiaaen=None, pmra=None, pmdec=None
                parallax=None, parallaxovererror=None, ebv=None, gaiabprpfactor=None,
                gaiasigma5dmax=None, gaiagmag=None, gaiabmag=None, gaiarmag=None,
                gaiadupsource=None, gaiaparamssolved=None,
-               primary=None, test=False, nside=2):
+               primary=None, test=False, nside=2, dr="dr2"):
```
geordie666, line 185: Consider reverting all changes in `isGAIA_STD()`.

py/desitarget/cuts.py
```diff
@@ -195,6 +195,8 @@ def isGAIA_STD(ra=None, dec=None, galb=None, gaiaaen=None, pmra=None, pmdec=None
     nside : :class:`int`, optional, defaults to 2
         (NESTED) HEALPix nside, if targets are being parallelized.
         The default of 2 should be benign for serial processing.
+    dr : :class:`str`, optional, defaults to "dr2"
```
geordie666, line 198: Consider reverting all changes in `isGAIA_STD()`.

py/desitarget/cuts.py
```diff
@@ -237,7 +239,7 @@ def isGAIA_STD(ra=None, dec=None, galb=None, gaiaaen=None, pmra=None, pmdec=None
                       gaiagmag=gaiagmag, gaiabmag=gaiabmag, gaiarmag=gaiarmag)
 
     # ADM restrict to point sources.
-    ispsf = gaia_psflike(gaiaaen, gaiagmag)
+    ispsf = gaia_psflike(gaiaaen, gaiagmag, dr=dr)
```
geordie666, line 242: Consider reverting all changes in `isGAIA_STD()`.

py/desitarget/gaiamatch.py
```diff
@@ -222,7 +237,8 @@ def sub_gaia_edr3(filename, objs=None):
 
     # ADM read the gaia sweep.
     cols = ["PARALLAX", "PARALLAX_IVAR",
-            "PMRA", "PMRA_IVAR", "PMDEC", "PMDEC_IVAR"]
+            "PMRA", "PMRA_IVAR", "PMDEC", "PMDEC_IVAR",
+            "GAIA_ASTROMETRIC_PARAMS_SOLVED"]
```
geordie666, line 241: I think this will fail? The parameter is called `EDR3_ASTROMETRIC_PARAMS_SOLVED` in the Gaia EDR3 sweeps files, so for this additional parameter you need to swap out `GAIA` for `EDR3` rather than just prepending `EDR3`?

py/desitarget/gaiamatch.py
```diff
@@ -222,7 +237,8 @@ def sub_gaia_edr3(filename, objs=None):
 
     # ADM read the gaia sweep.
     cols = ["PARALLAX", "PARALLAX_IVAR",
-            "PMRA", "PMRA_IVAR", "PMDEC", "PMDEC_IVAR"]
+            "PMRA", "PMRA_IVAR", "PMDEC", "PMDEC_IVAR",
```
geordie666, line 240: It looks like you made changes on AEN in the MWS main survey cuts. Are you absolutely sure you don't want to use the EDR3 version of astrometric_excess_noise, which would require transferring it in here?

geordie666 (changes requested):
@apcooper, I've a few requested changes. The main important point is that `apply_cuts_gaia()` will still use Gaia DR2, regardless of any changes you've made so far, unless we change the [call to all_gaia_in_tiles()](https://github.com/desihub/desitarget/blob/38413c0c020dfc9cb969caac301a38d0ce50454c/py/desitarget/cuts.py#L2808-L2811) to acknowledge the existence of EDR3. Changing `all_gaia_in_tiles()` will have knock-on effects in lots of places in the code, though.

I suggest that we take the easy route; don't update `apply_cuts_gaia()` at all, and just acknowledge the fact that the `BACKUP` targets assigned by `apply_cuts_gaia()` will still run with Gaia DR2 parameters. That choice will also mean that your changes to `isGAIA_STD()` are obsolete, so you could remove those changes to (although I think they're harmless, so that's up to you).

py/desitarget/gaiamatch.py
apcooper: Good catch, thanks.

py/desitarget/gaiamatch.py
apcooper: Again, good catch. I've added AEN to this list and removed the GAIA prefix from APS.

py/desitarget/cuts.py
apcooper: Reverted all changes.

py/desitarget/cuts.py
apcooper: Reverted all changes here too.

py/desitarget/cuts.py
apcooper: Reverted.

py/desitarget/cuts.py
apcooper: Reverted.

py/desitarget/gaiamatch.py
geordie666: This is almost correct now, but not quite. You need to turn [these columns](https://github.com/desihub/desitarget/blob/5b2a280bb303eef17d49ec748e631047c4e8fb3a/py/desitarget/gaiamatch.py#L63-L66):
```
    ('GAIA_ASTROMETRIC_SIGMA5D_MAX', '>f4'), ('GAIA_ASTROMETRIC_PARAMS_SOLVED', '>i1'),
    ('PARALLAX', '>f4'), ('PARALLAX_IVAR', '>f4'),
    ('PMRA', '>f4'), ('PMRA_IVAR', '>f4'),
    ('PMDEC', '>f4'), ('PMDEC_IVAR', '>f4')
```
Into [these columns](https://github.com/desihub/desitarget/blob/5b2a280bb303eef17d49ec748e631047c4e8fb3a/py/desitarget/gaiamatch.py#L93-L98):
```
    ('EDR3_ASTROMETRIC_SIGMA5D_MAX', '>f4'), ('EDR3_ASTROMETRIC_PARAMS_SOLVED', '>i1'), 
    ('EDR3_PARALLAX', '>f4'), ('EDR3_PARALLAX_IVAR', '>f4'),
    ('EDR3_PMRA', '>f4'), ('EDR3_PMRA_IVAR', '>f4'),
    ('EDR3_PMDEC', '>f4'), ('EDR3_PMDEC_IVAR', '>f4')
```
So the `PMRA` and `PARALLAX` columns need an `EDR3` prepended but the `GAIA_ASTROMETRIC_SIGMA5D_MAX` and `GAIA_ASTROMETRIC_PARAMS_SOLVED` columns need `GAIA` switched to `EDR3`. So, some code logic like:

```
gaiacols = [col.replace("GAIA", "EDR3") if "GAIA" in col else "EDR3_{}".format(col) for col in cols]
```
should work.

Sorry this is so tricky. I inherited the Gaia DR2 data model from the Legacy Surveys catalogs.

geordie666:
@apcooper: Sorry, but one more change requested (see the conversation that I "unresolved" above). Once you've corrected that, you can merge it if you like. I'll test the result and make any additional fixes (if I find that any are necessary) early tomorrow morning my time.

py/desitarget/gaiamatch.py
apcooper: OK, thanks. Sorry I'm clearly not following very well how the columns get copied. Will make those changes.

Also I thought the discussion with Boris above ended on GAIA_ASTROMETRIC_SIGMA5D_MAX being left as DR2. For consistency it seems better to me to change it, so I will follow your suggestion.

apcooper:
@geordie666, many thanks indeed for your help with this. I made those changes but I also noted that we hadn't discussed GAIA_DUPLICATED_SOURCE -- it gets used in the MWS selection so I guess it should also be updated, although I hope that doesn't lead to some inconsistency for the selections that are still using dr2...

I'll hold off on the merge until either you or @segasai or both have had chance to comment on that (and perhaps GAIA_ASTROMETRIC_SIGMA5D_MAX, although I think that's a safer change).

geordie666:
@apcooper:

- Regarding `GAIA_DUPLICATED_SOURCE`: I think it's fine to also update that. It's only every used in a `True`/`False` sense so I don't think it can ever be dangerous to substitute the EDR3 version into the mix. i.e. there won't be any changes to parameters/cuts.
- Regarding `GAIA_ASTROMETRIC_SIGMA5D_MAX`: It's only used (I think) by the WDs in the clause:
```
       mws &= ((astrometricsigma5dmax < 1.5) |
                ((astrometricexcessnoise < 1.) & (parallaxovererror > 4.) & (pm > 10.)))
```
I think it's fine to change it _if_ we don't need to change the 1.5 in the clause above, such that the EDR3 and DR2 parameter cuts remain identical for WDs.

apcooper:
@BGaensicke: please can you confirm (one last time, sorry) that if we change `astrometricsigma5dmax` to the EDR3 value, we can keep the value of 1.5 in `(astrometricsigma5dmax < 1.5)` for the WD selection?

apcooper:
@geordie666 if this branch is holding things up I think it's OK to merge now, I guess there might still be a chance to tweak astrometricsigma5dmax later...

geordie666:
There might not be a chance to tweak anything else starting from about 6 hours from now. So, let's leave this open as long as we can. It's not holding anything up yet.

BGaensicke:
@geordie666: The tests looked all good, no further changes from the WD side.

PR #722, opened 2021-05-04
Lookup for dynamic sky fiber placement

dstndstn:
This is the code we're using in fiberassign to do the lookup for good sky locations in the "skybricks" data product.  Stephen requested it go in desitarget so he can use it in the spectro pipeline.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -77,6 +77,9 @@ desitarget API
 .. automodule:: desitarget.secondary
     :members:
 
+.. automodule:: desitarget.skybricks
+    :members:
+
 .. automodule:: desitarget.skyfibers
     :members:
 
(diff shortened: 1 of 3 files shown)


geordie666:
@dstndstn: This PR looks fine, and it's OK to merge it. Before you do, though, make sure to update [the changes documentation](https://github.com/desihub/desitarget/blob/master/doc/changes.rst) and the [API for read-the-docs](https://github.com/desihub/desitarget/blob/master/doc/api.rst).

sbailey:
Thanks; I also tested that this works as expected, and updated the changes.rst and api.rst so that @dstndstn can focus on other topics.  I'll merge after tests re-pass.

dstndstn:
Wow, thanks @sbailey!

PR #723, opened 2021-05-04
Update baseline LRG selection

rongpu:
This PR updates the baseline LRG selection (assuming it will be the final main survey selection). Changes from SV3 include:
    * Change the zfiber faint limit from 21.7 to 21.6
    * Change the low-z limit from z>0.3 to z>0.4
    * Change the overall density from 800/sq.deg. to 600/sq.deg.
    * Remove the LRG_LOWDENS target bit

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -25,7 +25,7 @@ desitarget Change Log
      to swap EDR3 proper motions/parallaxes for values in sweeps files.
 * New function and bin script to make QSO redshift catalogs [`PR #714`_].
    * Incorporates functionality from QuasarNET and SQUEzE.
-* Update the baseline LRG selection [`PR #723`_]. Changes from SV3 includes:
+* Update the baseline LRG selection [`PR #723`_]. Changes from SV3 include:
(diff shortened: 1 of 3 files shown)


geordie666:
@rongpu: This all looks good to me, so no need for a detailed review.

One request. Can you update the docstring in `isLRG()` from:
```
- Current version (03/31/21) is version 15 on `the SV3 wiki`_.
```
to
```
- Current version (05/04/21) is version 254 on `the wiki`_.
```
Also, I think we can remove the `LRG_LOWDENS` bits from the [bit-mask yaml file](https://github.com/desihub/desitarget/blob/master/py/desitarget/data/targetmask.yaml), as those bits have never been used in official Main Survey target catalogs. If you like, I can deprecate those in my working branch, though.

geordie666:
Thanks! I'll merge once tests pass.

rongpu:
Thanks @geordie666. I've made the change in the docstring.

And yes, if you could remove the `LRG_LOWDENS` bits from the yaml file that would be great.

PR #732, opened 2021-05-14
add SKYBRICKS_DIR and make module file more portable using DESI_ROOT

sbailey:
@geordie666 this PR updates the etc/desitarget.module file to
  * be more portable by using `$DESI_ROOT` instead of `/global/cfs/cdirs/desi`
  * set `SKYBRICKS_DIR=$DESI_ROOT/target/skybricks/v3`
  * switch `$DUST_DIR` from `...cosmo/data/dust/v0_1` to `...desi/external/dust/v0_1`, which at NERSC is the same thing via links, but allows everything to be under `$DESI_ROOT` for environment portability

You can inspect the environment that would be installed with this at NERSC via:
```
source /global/cfs/cdirs/desi/software/desi_environment.sh 21.3
module unload desitarget
module load desitarget/modulefile
```
(the last two lines are the same as a `module swap desitarget/modulefile`, but allows you to inspect the environment in between to ensure that it really is unsetting `CMX_DIR` etc. before resetting correctly with the new environment file.

I'd like to tag this as 1.0.1 to get the `$SKYBRICKS_DIR` update.

### DIFF
--- a/etc/desitarget.module
+++ b/etc/desitarget.module
@@ -76,16 +76,17 @@ setenv [string toupper $product] $PRODUCT_DIR
 #
 # Add any non-standard Module code below this point.
 #
-setenv TARG_DIR /global/cfs/cdirs/desi/target/catalogs
-setenv MTL_DIR /global/cfs/cdirs/desi/target/mtl
-setenv DUST_DIR /global/cfs/cdirs/cosmo/data/dust/v0_1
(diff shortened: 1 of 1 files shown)


geordie666:
@sbailey: I tried out the `21.3` environment, and all of the directories behave exactly as I would expect. Feel free to merge this and tag `1.0.1` whenever you get a chance.

Don't forget to update the changes docs!

geordie666 (approved):

PR #733, opened 2021-05-16
Fix bug in MWS FAINT selection

apcooper:
Fix bug in MWS FAINT selection.


apcooper:
I'm testing this now -- although my understanding of how to run `desitarget` is a bit out of date. Will report here when I confirm the MWS_FAINT targets are included correctly.

apcooper:
I confirm the target files produced with this code contain sensible-looking numbers of FAINT_BLUE and FAINT_RED targets.

geordie666:
Thanks @apcooper.

I'll merge this, make a new tag, and run new targets in a few hours.

PR #739, opened 2021-05-27
Allow initial ledgers to use a preordained timestamp

geordie666:
This PR adds an option to pass a timestamp when making initial ledgers. The passed timestamp will then be used in place of the `TIMESTAMP` column generated by `make_mtl()`. `MWS_FAINT` targets can be exempted from this forced timestamp, and will instead adopt the usual "recent" timestamp generated by `make_mtl()`.

The purpose of these changes is to address the issue where `1.0.0` ledgers (and target files) were generated without `MWS_FAINT` targets, as fixed in #733. This PR should allow `1.1.0` ledgers to be generated with `TIMESTAMP`s that correspond to `1.0.0` ledgers for all targets except `MWS_FAINT`, as suggested by @sbailey.

Some other minor updates are included in this PR:

- The data model has been altered for the ledgers:
  * ``ZS`` and ``ZINFO`` are replaced by ``Z_QN``, as QuasarNP redshifts are the only non-redrock information that we'll be using to make LyA decisions.
  * This data-model-update is _not_ backwards compatible. So, if you're using `desitarget` functionality to read the never-to-be-spoken-of-again `1.0.0` ledgers, then you'll have to use version `1.0.0` of `desitarget`.
- A bug where an extraneous `hpxlist` was passed to `write_secondary()`, that was introduced in #734, and was [noticed by Anand ](https://github.com/desihub/desitarget/pull/738#issuecomment-849159262) has been fixed.

### DIFF
--- a/py/desitarget/mtl.py
+++ b/py/desitarget/mtl.py
@@ -625,8 +625,13 @@ def make_ledger_in_hp(targets, outdirname, nside, pixlist, obscon="DARK",
         _ = check_timestamp(timestamp)
         origts = mtl["TIMESTAMP"].copy()
         mtl["TIMESTAMP"] = timestamp
-        # ADM don't use the bespoke timestamp for MWS_FAINT targets.
+        # ADM don't use the bespoke timestamp for MWS_FAINT_* targets.
         if exemptmf:
+            # ADM do not update the timestamp for targets for which
+            # ADM MWS_FAINT_* is controlling the priority. As MWS_FAINT_*
+            # ADM targets are the lowest-priority UNOBSERVED targets,
+            # ADM this is equivalent to not updating the timestamp for
+            # ADM targets that are purely MWS_FAINT.
             ii = np.array(["MWS_FAINT" in ts for ts in mtl["TARGET_STATE"]])
             mtl["TIMESTAMP"][ii] = origts[ii]
 
--- a/py/desitarget/mtl.py
+++ b/py/desitarget/mtl.py
@@ -594,7 +594,10 @@ def is_pure_mws_faint(targets):
     # ADM to look up all combination of just MWS_FAINT_ bits.
     pure_mf_bits = []
     for bitvals in set(targets[mws_target_col]):
-        mf = np.all(["MWS_FAINT" in names for names in mws_mask.names(bitvals)])
+        # ADM at a minimum, the bit combination must have MWS_FAINT_ set...
+        mf = np.any(["MWS_FAINT" in names for names in mws_mask.names(bitvals)])
+        # ADM ...but it must ONLY have combinations of MWS_FAINT_ set.
+        mf &= np.all(["MWS_FAINT" in names for names in mws_mask.names(bitvals)])
         if mf:
             pure_mf_bits.append(bitvals)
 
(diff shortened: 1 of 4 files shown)


geordie666:
A few example ledgers that include timestamps that have been "forced" for every target class except `MWS_FAINT` are available at NERSC in `/global/cscratch1/sd/adamyers/PR739`.

py/desitarget/mtl.py
```diff
@@ -557,20 +593,43 @@ def make_ledger_in_hp(targets, outdirname, nside, pixlist, obscon="DARK",
...
+        hdr["TIMESTAMP"] = timestamp
+        hdr["EXEMPTMF"] = exemptmf
+        # ADM check the timestamp is valid.
+        _ = check_timestamp(timestamp)
+        origts = mtl["TIMESTAMP"].copy()
+        mtl["TIMESTAMP"] = timestamp
+        # ADM don't use the bespoke timestamp for MWS_FAINT targets.
+        if exemptmf:
+            ii = np.array(["MWS_FAINT" in ts for ts in mtl["TARGET_STATE"]])
+            mtl["TIMESTAMP"][ii] = origts[ii]
```
sbailey, line 631: It appears that this sets the original (i.e. later, current) timestamp for any target with MWS_FAINT set, when I think we want this only for targets that have *only* MWS_FAINT set, i.e. could not have appeared in a bright mtl due to another bit being set.  In the test healpixels I haven't found a case where that happens, but there are cases where they also pass LRG cuts (e.g. TARGETID 39627601496506987 in mtl-bright-hp-6466.ecsv), and I think in principle this could happen for bright bits as well.  Please check.

sbailey (commented):
Thanks for adding this, and thanks for providing pre-made example outputs to inspect.  See two comments inline.

py/desitarget/mtl.py
```diff
@@ -557,20 +593,43 @@ def make_ledger_in_hp(targets, outdirname, nside, pixlist, obscon="DARK",
...
 
     # ADM in case an integer was passed.
     pixlist = np.atleast_1d(pixlist)
 
     # ADM execute MTL.
     mtl = make_mtl(targets, obscon, trimcols=True)
 
+    # ADM if requested, substitute a bespoke timestamp.
+    if timestamp is not None:
+        hdr["TIMESTAMP"] = timestamp
```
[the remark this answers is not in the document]
geordie666: Nice catch. I honestly remember thinking when I wrote this that the word "TIMESTAMP" only had 8 letters! Worrying. I changed it to "TSFORCED".

py/desitarget/mtl.py
geordie666: This doesn't set the current timestamp for any target with `MWS_FAINT` set. It only sets the current timestamp for targets _for which `MWS_FAINT` is driving the priority_. As `MWS_FAINT` is at the lowest priority for first-pass targets, this correctly only finds objects that have _only_ `MWS_FAINT` set.

It's a bit of trickery, admittedly, but I originally spent an hour tracking down all of the corner-cases where bits related to `MWS_FAINT_*` were set but no other bits, and no secondary matches, and only in bright time. This became very complicated. I'm far more confident that targets for which `MWS_FAINT` is "in control" are the targets that are new. And if `MWS_FAINT_*` _weren't_ the lowest-priority targets, we wouldn't want to add them anyway, as they'd bump other targets.

I think we can be confident that this was a reasonable choice, given the extensive vetting of the new ledgers by you, me and @ashleyjross this morning.

I've added a longer comment explaining this snippet of code.

geordie666:
@sbailey: I think this is ready to merge, and I'd like to merge it in about an hour so I can tackle the MTL loop after lunch without worrying about creating merge conflicts.

sbailey:
Thanks for the explanation, looks good.  Merging now.

PR #740, opened 2021-05-27
subpriority overrides, take 2

sbailey:
This PR replaces #736 which I will close after submitting this.

This provides tools for overriding SUBPRIORITY when generating target files, to work around the accidental SUBPRIORITY reassignments from fiberassign in May 2021:
  * `desitarget.subpriority.override_subpriority(targets, override)` - override targets['SUBPRIORITY'] in-place, for those TARGETIDs that appear in the override table.
  * `desitarget.subpriority.get_fiberassign_subpriorities` used to scrape fiberassign TARGETS HDUs to build override lists for dark, bright, and sky.
    * examples in /global/cscratch1/sd/sjbailey/desi/targets/subpriorities*.fits
    * this can be called from the command line but purposefully isn't put into the default desitarget installation path, e.g.
```
fadir=/global/cfs/cdirs/desi/target/fiberassign/tiles/trunk
python $DESITARGET/py/desitarget/subpriority.py -o . \
    -i $fadir/001/*.fits.gz $fadir/002/*.fits.gz \
       $fadir/020/*.fits.gz $fadir/021/*.fits.gz
```
  * `select_targets --dark-subpriorities ... --bright-subpriorities ...` options to provide override lists
  * `select_skies --sky-subpriorities ...` and `supplement_skies --sky-subpriorities subpriorities-sky.fits` options to provide sky SUBPRIORITY override lists
  * more unit tests related to subpriority overrides and writing files

One change of behavior in `write_targets` et al:
  * subpriority=True means set any SUBPRIORITY==0.0, but leave any SUBPRIOITY>0.0 alone (different than current master which overrides all or nothing)
  * if you really want to override all, set input SUBPRIORITY==0.0 before calling `write_targets(..., subpriority=True)`.

Examples in `/global/cscratch1/sd/sjbailey/desi/targets/default` and `override/`.  I believe I have tested the following but would appreciate additional checks:
  * targets appear in the same order regardless of override options
  * targets that differ in SUBPRIORITY are in the override list
  * targets with the same SUBPRIORITY are *not* in the override list
  * TARGETIDs in the override list have the SUBPRIORITIES from that list

@araichoor @geordie666

### DIFF
--- a/bin/select_secondary
+++ b/bin/select_secondary
@@ -33,6 +33,10 @@ ap.add_argument("--nomasking", action='store_true',
 ap.add_argument("--maskdir",
                 help="Name of the specific directory (or file) containing the bright star mask (defaults to the most recent directory in $MASK_DIR)",
                 default=None)
+ap.add_argument("--dark-subpriorities", type=str,
+                help='Optional file with dark TARGETID:SUBPRIORITY overrides')
+ap.add_argument("--bright-subpriorities", type=str,
+                help='Optional file with bright TARGETID:SUBPRIORITY overrides')
 
 ns = ap.parse_args()
 do_mask = not(ns.nomasking)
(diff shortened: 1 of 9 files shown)


dstndstn:
This all looks reasonable to me

PR #750, opened 2021-06-03
zqso robust to missing *_SCND_TARGET columns; record quasarnp version

sbailey:
This PR fixes a `make_zqso_files` bug when the input file FIBERMAP is missing a `*_SCND_TARGET` input column (because the input fiberassign file was also missing that column).  e.g. from the example in PR 737:
```
[cori06 ~] make_zqso_files -t 80605 -n 20210205 -o $CSCRATCH/desi/zqsotest
INFO:lyazcat.py:66:tmark: 
    Loading QuasarNP Model file and lines of interest: 2021-06-03 | 14:58:14
INFO:lyazcat.py:66:tmark: 
      QNP model file loaded: 2021-06-03 | 14:58:14
INFO:make_zqso_files:141:<module>: processing TILEID=80605, NIGHTID=20210205, petal=0
INFO:lyazcat.py:66:tmark: 
    Making redrock zcat: 2021-06-03 | 14:58:14
Traceback (most recent call last):
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desitarget/master/bin/make_zqso_files", line 143, in <module>
    create_zcat(input_dir, output_dir, tile=tile,
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desitarget/master/py/desitarget/lyazcat.py", line 754, in create_zcat
    zcat = make_new_zcat(zbestfn, qn_flag, sq_flag, abs_flag, zcomb_flag)
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desitarget/master/py/desitarget/lyazcat.py", line 141, in make_new_zcat
    zcat[col] = fms[zid][col]
ValueError: no field of name SV1_SCND_TARGET
```
This PR logs an error and leaves that column as zeros in the output file:
```
[cori06 ~] make_zqso_files -t 80605 -n 20210205 -o $CSCRATCH/desi/zqsotest
INFO:lyazcat.py:66:tmark: 
    Loading QuasarNP Model file and lines of interest: 2021-06-03 | 15:00:35
INFO:lyazcat.py:66:tmark: 
      QNP model file loaded: 2021-06-03 | 15:00:35
INFO:make_zqso_files:141:<module>: processing TILEID=80605, NIGHTID=20210205, petal=0
INFO:lyazcat.py:66:tmark: 
    Making redrock zcat: 2021-06-03 | 15:00:35
[...]

PR #751, opened 2021-06-04
Add the final QSO decision logic to the full Main Survey MTL loop.

geordie666:
This PR should be the final piece needed for the full MTL loop for the Main Survey. It adds the logic for deciding which quasars are observed at high priority on repeat passes. To try to succinctly describe the logic:

- Let `Z` denote the redrock redshift.
- Let `Z_QN` denote the QuasarNP redshift.
- Let `IS_QSO_QN` denote the QuasarNP decision as to whether something is a quasar or not.

Then:

- If a QSO target has `Z > 2.1` OR (`Z_QN > 2.1` AND `IS_QSO_QN==1`) it is confirmed as a **LyA quasar**. It receives 3 more observations (for a total of 4 observations) at a very high priority (just below an unoberved quasar).
- If a QSO target has `Z < 1.6` OR `Z_QN < 1.6` OR `IS_QSO_QN !=1` it is confirmed as a (possible) **tracer quasar**. It receives 1 more observation (for a total of 2 observations) at a priority lower than all other targets but above "filler" targets.
- All other QSO targets are considered to be **"mid-z" quasars** (requested by a secondary program). They receive 3 more observations (for a total of 4 observations) at a priority lower than all other targets but above "filler" targets.

Additionally:

- LyA QSOs are "locked in" to their priority state as soon as they are determined to be a LyA quasar. This is to ameliorate a bout of unlucky spectroscopy consigning them to be tracer quasars from which it may be difficult to recover.
- All other quasar targets are allowed to be promoted to LyA quasars should they meet the LyA criteria on subsequent passes. No quasar target ever receives more than 4 total observations, though.
- "Standalone" secondary QSO programs (`WISE_VAR_QSO`, `QSO_RED`) share the same logic as for primary QSOs. Almost all (standalone) targets in these programs fall in the tracer category, though. This is generally because of the shallower WISE selection(s) used by these programs, which skews the redshift distribution to z~1 compared to quasars selected with more overall weight given to optical data.

I'll use this branch to make MTL ledgers for review once the zqso catalogs are ready at NERSC. I'll subsequently merge this PR if those MTL ledgers look reasonable to the Operations team and some select quasologists.

### DIFF
--- a/py/desitarget/subpriority.py
+++ b/py/desitarget/subpriority.py
@@ -11,6 +11,7 @@ from desiutil.log import get_logger
 from desitarget.targetmask import desi_mask
 from desitarget.geomask import match
 
+
 def override_subpriority(targets, override):
     """
     Override SUBPRIORITY column in targets for those in override table
@@ -50,6 +51,7 @@ def override_subpriority(targets, override):
 
         return ii
 
+
 def get_fiberassign_subpriorities(fiberassignfiles):
     """
     Return table of TARGETID, SUBPRIORITY used in input fiberassign files
@@ -62,13 +64,13 @@ def get_fiberassign_subpriorities(fiberassignfiles):
     """
     log = get_logger()
 
-    #- allow duplicate inputs, but don't process multiple tiles
+    # - allow duplicate inputs, but don't process multiple tiles
     processed = set()
 
     subprio_tables = dict(dark=list(), bright=list(), sky=list())
 
     for filename in fiberassignfiles:
-        #- Have we already processed this file (e.g. from an earlier expid)?
+        # - Have we already processed this file (e.g. from an earlier expid)?
         basename = os.path.basename(filename)
         if basename in processed:
             continue
@@ -98,7 +100,7 @@ def get_fiberassign_subpriorities(fiberassignfiles):
             log.info(f'Reading {filename}')
             sp = fx['TARGETS'].read(columns=['TARGETID', 'SUBPRIORITY', 'DESI_TARGET'])
 
-        #- Separate skies from non-skies
+        # - Separate skies from non-skies
         skymask = desi_mask.mask('SKY|SUPP_SKY|BAD_SKY')
         iisky = (sp['DESI_TARGET'] & skymask) != 0
 
@@ -109,13 +111,13 @@ def get_fiberassign_subpriorities(fiberassignfiles):
     for program in subprio_tables.keys():
         subprio_tables[program] = np.hstack(subprio_tables[program])
 
-    #- QA checks on basic assumptions about uniqueness
+    # - QA checks on basic assumptions about uniqueness
     log.info('Checking assumptions about TARGETID:SUBPRIORITY uniqueness')
     for program in ['dark', 'bright', 'sky']:
         subprio = subprio_tables[program]
         tid, sortedidx = np.unique(subprio['TARGETID'], return_index=True)
 
-        #- sky can appear multiple times, but with same SUBPRIORITY
+        # - sky can appear multiple times, but with same SUBPRIORITY
         if program == 'sky':
             subpriodict = dict()
             for targetid, sp in zip(
@@ -125,7 +127,7 @@ def get_fiberassign_subpriorities(fiberassignfiles):
                         log.error(f'{program} TARGETID {targetid} has multiple subpriorities')
                 else:
                     subpriodict[targetid] = sp
-        #- but other programs should have each TARGETID exactly once
+        # - but other programs should have each TARGETID exactly once
         else:
             if len(tid) != len(subprio):
                 log.error(f'Some {program} TARGETIDs appear multiple times')
@@ -135,14 +137,15 @@ def get_fiberassign_subpriorities(fiberassignfiles):
 
     return subprio_tables
 
+
 if __name__ == "__main__":
     import argparse
 
     p = argparse.ArgumentParser()
     p.add_argument('-i', '--infiles', nargs='+', required=True,
-            help='Input fiberassign files with TARGETS HDU')
+                   help='Input fiberassign files with TARGETS HDU')
     p.add_argument('-o', '--outdir', required=True,
-            help='Output directory to keep dark/bright/sky TARGETID SUBPRIORITY tables')
+                   help='Output directory to keep dark/bright/sky TARGETID SUBPRIORITY tables')
 
     args = p.parse_args()
     log = get_logger()
@@ -168,6 +171,3 @@ if __name__ == "__main__":
         outfile = os.path.join(args.outdir, f'subpriorities-{program}.fits')
         fitsio.write(outfile, subprio, extname='SUBPRIORITY', header=hdr, clobber=True)
         log.info(f'Wrote {outfile}')
-
-
-
--- a/py/desitarget/targets.py
+++ b/py/desitarget/targets.py
@@ -576,14 +576,19 @@ def calc_numobs_more(targets, zcat, obscon):
             # ADM If a QSO target is confirmed to have a redshift at
             # ADM z < midzcut it will need to drop by 2 total observations
             # ADM (midzcut is defined at the top of this module).
-            ii = targets[desi_target] & desi_mask.QSO != 0
+            isqso = targets[desi_target] & desi_mask.QSO != 0
             # ADM the mocks may not include the secondary targets.
             if scnd_target in targets.dtype.names:
                 for scxname in scnd_mask.names():
                     if scnd_mask[scxname].flavor == "QSO":
-                        ii |= targets[scnd_target] & scnd_mask[scxname] != 0
-            ii &= (zcat['ZWARN'] == 0)
-            ii &= (zcat['Z'] < midzcut)
+                        isqso |= targets[scnd_target] & scnd_mask[scxname] != 0
+            # ADM the definition used for "not-LyA" in calc_priority.
+            midz = (zcat['Z'] < zcut) & ((zcat['Z_QN'] < zcut)
+                                         | (zcat["IS_QSO_QN"] != 1))
+            # ADM the likely low-z sources get fewer (2) observations.
+            loz = ((zcat['Z'] < midzcut) | (zcat['Z_QN'] < midzcut) |
+                   (zcat["IS_QSO_QN"] != 1))
+            ii = isqso & midz & loz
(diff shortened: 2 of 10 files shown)


py/desitarget/targets.py
```diff
@@ -860,26 +909,53 @@ def calc_priority(targets, zcat, obscon, state=False):
...
+                        # ADM Main Survey decisions are made using QN/Redrock.
+                        if survey == "main":
+                            good_hiz = (zcat['Z'] >= zcut) | ((zcat['Z_QN'] >= zcut) &
+                                                              (zcat["IS_QSO_QN"] == 1))
+                            # ADM all non-LyA-QSOs need more low-priority passes
+                            # ADM in the Main Survey. The mid-z QSOs get 4 passes
+                            # ADM at this lower priority, as requested by some
+                            # ADM secondaries, which is set in calc_numobs_more.
+                            good_midz = (zcat['Z'] < zcut) & ((zcat['Z_QN'] < zcut) |
+                                                              (zcat["IS_QSO_QN"] != 1))
```
sbailey, line 924: Is `!= 1` the correct test here, or should it be `== 0`?

py/desitarget/subpriority.py
```diff
@@ -62,13 +64,13 @@ def get_fiberassign_subpriorities(fiberassignfiles):
     """
     log = get_logger()
 
-    #- allow duplicate inputs, but don't process multiple tiles
+    # - allow duplicate inputs, but don't process multiple tiles
```
sbailey, line 67: FWIW, I suspect that some style checker converted all my `#-` comments into `# -`.  You don't need to use my style in desitarget and I try to follow your `# INITIALS` style in desitarget when I'm thinking of it, but sometimes habit slips through.  The basic idea of `#-` is that it distinguishes genuine comments from `# ` which I use for commenting out code.  In practice, I'm the only person who follows that in desi code so `#-` and now `# -` defacto means `# SB`.  :)

sbailey (commented):
The text description looks good, and I don't see any obvious errors in the code, but the chain of logic is subtle enough that I would trust extensive tests more than I would trust my personal reading of the logic.  In that spirit, it doesn't look like all the special cases are covered by the tests (locking in LyA even if a future redshift claims otherwise, promoting a second observed tracer to LyA, low-, mid-, high- z quasars, etc.).  Please check if the tests have explicit full coverage for all the logical cases that could occur.

Note that PR #750 should be merged prior to me running zqso on sv daily data, which you list as a pre-requisite for finalizing this PR, so effectively that PR is a blocking factor for this PR.

py/desitarget/targets.py
geordie666: This is correct. `IS_QSO_QN` can have values of -1, 0 and 1.

py/desitarget/subpriority.py
geordie666: That's fine and understood. I'm just trying to (pedantically) pass the PEP 8 style-checker as much as possible. One PEP 8 check is that there's a space after every comment.

geordie666:
The basic/main QSO cases (tracer/secondary-mid-z-QSO/LyA-QSO) are already in the unit tests. There are possibly one or two that I've missed, which I can add relatively quickly.

Adding "exhaustive" tests will require significant new functionality that we have never previously included. First, we would also need to check decisions on passes 2, 3 and 4 of the overlapping observations, not just on passes 0 and 1. Second, we would also need to include secondary targets to be truly exhaustive. 

There are potentially dozens of individual cases as a few of the secondary targets request additional observations and could, in theory, merge with a primary QSO that takes precedence in dark time but not in bright time. Checking many of these cases using _real_ data took me several days of work for this PR. Writing such tests with ersatz data will potentially take a great deal of time and delay merging this PR (and the creation of the subsequent MTL catalogs). I doubt I can add all of the cases by Monday.

We also don't have tests that check the full MTL loop, and never have. To some extent, I see checking pass-2/3/4 decisions and looking for corner cases as the purview of mocks rather than of unit tests. We have never unit-tested beyond passes 0 and 1.

geordie666:
I added unit tests to:

- Check dark-time targets behave as expected after 0, 1, 2, 3, 4 passes, in some basic sense.
- Check quasar targets are "locked into" their priority state once they become a LyA QSO.

If I find some time over the weekend I'll add a few more cases.

geordie666:
I extended the "locked in" test to an additional pass, and added a new unit test, to check that:

- A QSO promoted to being a LyA QSO on the _second_ pass retains that status for the final pass.
- Mid-z QSOs are not "locked in," _i.e._ they can revert to being tracer QSOs as soon as a redshift combination is measured that suggests they might be tracers.

geordie666:
I extended the "locked in" test further, to check that:

- A tracer QSO can also be promoted to being a LyA QSO, and is then "locked in" to being a LyA QSO.
- Regardless of updating redshifts in different ways on different passes, every QSO is still `DONE` after a full 4 passes.

sbailey:
Thanks for adding the extra tests.  Can you confirm that the LyA followup choices purposefully do *not* use ZWARN or DELTACHI2?  I don't see that in the description, which surprises me, especially for more dramatic ZWARN bits like NODATA, Z_FITLIMIT, or BAD_MINFIT.  Or is that caught earlier before a target is even considered for LyA followup?

sbailey:
Running `make_mtl` alters the input `zcat` table by adding a `NUMOBS_MORE` column.  Ideally `make_mtl` should not alter the inputs, or otherwise this side effect should be documented (including why).  If you are only adding a column for bookkeeping convenience but not otherwise modifying the rows of the pre-existing columns, you could start with
```
zcat = zcat.copy(copy_data=False)
```

sbailey:
@geordie666 are your criteria applied in order and the first matching one determines the state?  e.g.

`Z=0.5, QN_Z=2.5, IS_QSO_QN==1` matches your first criterion as a **LyA quasar** but matches your second criterion as a **tracer quasar**.  Does the first criterion win?

[apologies to others getting N>>1 emails via this ticket from me; more are likely coming...]

sbailey:
minor: `make_mtl` docstring says that `targets` requires columns `TARGETID, DESI_TARGET, NUMOBS_INIT, PRIORITY_INIT`; experimentally it also requires `BGS_TARGET`, `MWS_TARGET`, and if `zcat` is not None then it also requires `PRIORITY` (i.e. if zcat is not None then targets needs to be a previous mtl, not an original targets table, even if this is the first iteration).  That matches how targets -> mtl ledger -> obs -> mtl updates happens in practice, but is somewhat inconsistent with the docstring.

sbailey:
@geordie666 FWIW I independently tested all the initial cases of RR/QN for the first MTL update.  Initially I found a bunch of errors, then realized that I had the master branch checked out; switching to this branch fixed all the failing cases.  Good!  Attached is a gzipped python script I used (GitHub doesn't allow direct .py attachments) in case it covers any cases that you don't think you have already covered in unit tests.

[check_lya_mtl.py.gz](https://github.com/desihub/desitarget/files/6603621/check_lya_mtl.py.gz)

I have *not* independently tested the "locked in" and "promoted" cases, though it sounds like you have those included in your unit tests.

Summarizing my tests/questions for the night:
  * Please confirm that make_mtl purposefully does *not* use DELTACHI2, and NODATA is the only ZWARN bit used
  * make_mtl alters zcat, which is undesireable
  * Docstring comments
  * I have independently tested the functionality *except* for the LyA lockin and promotion, and didn't find any problems with this branch

geordie666:
@sbailey: In answer to your comments:

- For the **_Main Survey_**, `make_mtl()` purposefully does not use `DELTACHI2`, and `NODATA` is the only `ZWARN` bit used. But for **_SV_**, both `DELTACHI2` and `ZWARN` were needed and used, so are retained throughout for backwards compatibility.
- `make_mtl( )` has added a `NUMOBS_MORE` column to the `zcat` for ~5 years, I think?

numobs_more_blame

So, I'm not sure what your request is, here? If the issue is that the _number of rows_ in the `zcat` can potentially change, that's been possible since you updated the code back in late 2018 (specifically, your `zcat = zcat[ok]` line can alter the number of rows in the `zcat`):

alter zcat

My only change in this initial zcat-altering-behavior was to log that this alteration was a possibility. Essentially, I don't think this specific PR changes existing allowed behavior in any way, it merely inherits behavior from earlier PRs?
- Ah, yes, I updated the docstrings in `calc_numobs_more` and `calc_priority` but not in `make_mtl`. I'll update that before I merge this PR. Thanks!
- It's great that you've independently tested the functionality. To answer one more of your questions: `Z=0.5, QN_Z=2.5, IS_QSO_QN==1 matches your first criterion as a LyA quasar but matches your second criterion as a tracer quasar. Does the first criterion win?`. I should have specified that the tracer/mid-z QSOs are only ever drawn from the population of quasars that are explicitly _not_ LyA quasars. So, yes, the first (LyA) criterion wins.

geordie666:
Sorry, I misunderstood your 2nd point, above, by conflating a discussion of the MTL table with a discussion of the `zcat`. I think I see what you mean, now. So, to summarize: 

- It's OK to alter the number of rows in the `zcat` (based on your 2018 PR) but; 
- It's not OK to alter the columns/dtype of the `zcat`, which makes the data model hard to track.

If that's a correct synopsis, I can fix the addition of the `NUMOBS_MORE` column, sure. Although the `zcat` has been modified-in-place by `make_mtl()` for as long as I can remember, nothing about processing the "real" data requires that modification, so I've always assumed it was a necessary feature for the mocks.

sbailey:
Thanks for the additional info @geordie666 .  I'm apparently 5 years late in noticing that `make_mtl` modifies `zcat` by adding a new `NUMOBS_MORE` column.  I suspect that is an unintended side-effect of the bookkeeping rather than something actually used for mocks, but in the interest of not breaking things at the last moment let's just document this behavior while we're noticing it, but not change it right now.  i.e. this is a technical debt/maintenance issue to document whenever a function modifies its inputs (as opposed to just using them to derive outputs).

Additionally, if you know of any other inputs that are modified, or know whether `make_mtl` would use a pre-existing `NUMOBS_MORE` column, that should be documented too.

Otherwise I'm satisfied with the functionality of this PR.  Thanks.

geordie666:
I expect that this is ready to merge, now, but I'm still going to run a few more tests through the rest of the evening. I'll likely self-merge this early tomorrow.

Issue #749, opened 2021-06-02
read_targets in 1.1.1 not properly reading SV3 MTL ledgers


araichoor:
thanks to an issue reporting from Eddie:

the ['SV3_DESI_TARGET', 'SV3_BGS_TARGET', 'SV3_MWS_TARGET', 'SV3_SCND_TARGET'] columns are not read anymore when one reads the SV3 MTL ledgers with the 1.1.1 tagged version:

with 1.1.1:
```
>>> import desitarget
>>> desitarget.__version__
'1.1.1'
>>> from desitarget.io import read_targets_in_box
>>> d = read_targets_in_box("/global/cfs/cdirs/desi/survey/ops/surveyops/trunk/mtl/sv3/dark", radecbox=[270,271,60,61], mtl=True)
>>> [key for key in d.dtype.names if key[:3]=="SV3"]
[]
```

with 1.0.1:
```
>>> import desitarget
>>> desitarget.__version__
'1.0.1'
>>> from desitarget.io import read_targets_in_box
>>> d = read_targets_in_box("/global/cfs/cdirs/desi/survey/ops/surveyops/trunk/mtl/sv3/dark", radecbox=[270,271,60,61], mtl=True)
>>> [key for key in d.dtype.names if key[:3]=="SV3"]
['SV3_DESI_TARGET', 'SV3_BGS_TARGET', 'SV3_MWS_TARGET', 'SV3_SCND_TARGET']
```

geordie666:
Addressed in #755.

PR #753, opened 2021-06-09
Speed-ups and bug-fixes for I/O routines

geordie666:
This PR fixes a bug when reading MTL ledgers. It also implements some speed-ups in the functions that read targets, which were suggested by @araichoor. To summarize:

- When we moved to the final MTL data model, we broke the code that read Main Survey ledgers created by version `1.0.0` of desitarget (see also PR #742). The quick-fix that was implemented to address this issue broke the reader for SV targets (see issue #749). This PR implements a less fragile fix that facilitates reading all types of MTL ledgers, including SV3 and Main Survey `1.0.0`.
- This PR implements a faster version of `io.read_targets_in_tiles()` via a new `io.read_targets_in_tiles_quick()` function.
- This PR also speeds up `mtl.inflate_ledger()` using a new `mtl.match_ledger_to_targets()` function.


geordie666:
@araichoor: I think you can switch your bespoke code in [fba_launch_io.py](https://github.com/desihub/fiberassign/blob/208717798c1f71988c72e5c3b7e41d0b1374039e/py/fiberassign/fba_launch_io.py#L1179-L1200) over to just two lines of code using the following schema:

1. Read the targets using `targets=io.read_targets_in_tiles(hpdirname, tiles=tiles, quick=True)` or by directly using `targets=io.read_targets_in_tiles_quick(hpdirname, tiles=tiles)`
2. Match the targets and the mtl using the quick version of `mtl.inflate_ledger()`, which is `mtl_with_target_columns = mtl.match_ledger_to_targets(mtl, targets)`.

You might also want to just try `mtl_with_target_columns = mtl.inflate_ledger(mtl, targdir, quick=True)` directly, as there doesn't appear to be a huge speed-up from switching to using `io.read_targets_in_tiles_quick()`.

It would be great if you could try this set-up for some pass 0 tiles and check that results are identical and (hopefully) faster than the standard `desitarget.io` routines.

PR #754, opened 2021-06-16
add ZWARN bad QA bits

sbailey:
This PR adds ZWARN bits BAD_SPECQA and BAD_PETALQA and updates the MTL logic to reject those in addition to NODATA.  I'll open a companion desispec PR after this to set those bits when making zmtl (formerly zqso formerly zcat) files.

Example desispec outputs are in /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/zmtl, though I have not tested how mtl updates would use them.  Unit tests did catch some typos so the updates are at least somewhat tested.

@geordie666 please take a look.  I did *not* update the $ZCAT_DIR environment variable, but let's consider renaming that to $ZMTL_DIR too in both the code and etc/desitarget.module .

Also: I did *not* remove the zqso code from desitarget yet, even though it is moving over to desispec as part of the aforementioned PR.  Let's make sure it fully works there (as zmtl) before removing it from desitarget as a separate PR.

### DIFF
--- a/py/desitarget/data/targetmask.yaml
+++ b/py/desitarget/data/targetmask.yaml
@@ -223,6 +223,7 @@ targetid_mask:
 # ADM ZWARN mask (bits that can be set to warn that a redshift is bad).
 # ADM inherited from https://github.com/desihub/redrock/blob/master/py/redrock/zwarning.py
 zwarn_mask:
+    # SB bits 0-15 set by redrock
     - [SKY,                0, "sky fiber"]
     - [LITTLE_COVERAGE,    1, "too little wavelength coverage"]
     - [SMALL_DELTA_CHI2,   2, "chi-squared of best fit is too close to that of second best"]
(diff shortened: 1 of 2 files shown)


geordie666:
Just so you know: To test this I've flipped two of the tiles in `/global/cfs/cdirs/desi/users/sjbailey/spectro/redux/zmtl/tiles.csv` to `zdone=true`.

geordie666 (approved):
@sbailey: I've now run your example files through the full MTL loop and everything works. Reviewing the code, everything looks as I would expect, too. Thanks!

Unless you have a strong objection, let's _not_ change the `ZCAT_DIR` environment variable. For SV3, we used the `zbest` files directly, rather than reading the new `zmtl` files. So, `ZCAT_DIR` seems like a more general name for all iterations of MTL.

In addition, (by default) the code is looking for whatever files are in `/global/cfs/cdirs/desi/spectro/redux/daily`. That seems more like a "ZCAT" location than specifically a "ZMTL" location.

Finally, the `zcat` nomenclature is used in a number of places in `bin/run_mtl_loop` and `mtl.py` so we aren't going to scrub it completely without more effort (and substantial care).

geordie666:
Oh, and I've now flipped the two tiles I tested in `/global/cfs/cdirs/desi/users/sjbailey/spectro/redux/zmtl/tiles.csv` back to `zdone=false`.

sbailey:
Thanks for the quick review.  I'll leave $ZCAT_DIR as-is, and update the change log after merging so that I don't have to wait for tests again.

PR #756, opened 2021-07-15
Modified read_mtl_ledger to change datamodel for different survey types

jalasker:
While creating the code for the alternate MTLs (currently in https://github.com/desihub/LSS/tree/AltFATools/py/LSS/SV3/altmtltools), I had trouble with getting desitarget.io to read the SV3 MTLs in a way that would not cause errors involving the code later searching for "DESI_TARGET" columns rather than "SV3_DESI_TARGET" columns. This minor change appears to fix the issue.


jalasker:
Issue was with environment variables not code. Closing without merging.

PR #757, opened 2021-07-17
add redrock ZWARN bit 11=POORDATA

sbailey:
Adds ZWARN bit 11 "POORDATA" which was recently added to the redrock ZWARN bitmask, and which desitarget unit tests properly caught hadn't been updated here yet.  (background: desitarget doesn't depend upon redrock both to avoid circular dependencies and to simplify what is required for desitarget to run, but unit tests check that it's list of redrock ZWARN bits is current).

I think the equivalent of this bit is already propagated through tile QA and zmtl in FIBERSTATUS, so I don't think any change is needed on the MTL side, but I will leave this open for a cross check and/or a second look on a non-weekend day.  The change on the redrock side was to set the bit for positioner offset > 30 microns at the time of running redrock, and not just post-facto as part of tile QA.

### DIFF
--- a/py/desitarget/data/targetmask.yaml
+++ b/py/desitarget/data/targetmask.yaml
@@ -235,6 +235,7 @@ zwarn_mask:
     - [BAD_TARGET,         8, "catastrophically bad targeting data"]
     - [NODATA,             9, "No data for this fiber, e.g. because spectrograph was broken during this exposure (ivar=0 for all pixels)"]
     - [BAD_MINFIT,        10, "Bad parabola fit to the chi2 minimum"]
+    - [POORDATA,          11, "Poor input data quality but try fitting anyway"]
     # SB bits 16-24 available for DESI to set post-redrock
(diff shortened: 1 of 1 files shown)


geordie666:
Just to make absolutely sure, when `POORDATA` is set, one of:

```
NODATA
BAD_SPECQA
BAD_PETALQA
```

is also always set, right? Otherwise we need an update in `desitarget.mtl.make_mtl()`.

(and don't forget to update the changes file).

sbailey:
Tracing backwards:
  * Redrock sets ZWARN mask bit POORDATA when desispec FIBERSTATUS mask bit POORPOSITION is set.
    * Redrock uses the more generic name "POORDATA" for the ZWARN bitmask since there might be other causes of poor data in the future, and this is a catchall warning bit that something is problematic about the spectrum, but redrock will try fitting it anyway.
  * desispec tile QA -> zmtl also uses FIBERSTATUS.POORPOSITION to set the BAD_SPECQA bit, so MTL gets it through that route.

So we're ok on the mtl side; merging now (and then will update change log in master)

PR #765, opened 2021-09-30
Change the location of the (zdone) tile file for SV

geordie666:
This PR updates the MTL loop to use `surveyops/ops/tiles-specstatus.ecsv` as the location of the tile file that contains `zdone` information for SV. 

We'd already updated the location of this tile file for the Main Survey, and were searching for a legacy copy of `tiles.csv` for SV. With this PR, we should no longer need to rely on `tiles.csv` at all.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,9 @@ desitarget Change Log
 1.3.1 (unreleased)
 ------------------
 
-* No changes yet.
+* Also use the ops/tiles-specstatus.ecsv tile file for SV [`PR #765`_].
+
+.. _`PR #765`: https://github.com/desihub/desitarget/pull/765
 
 1.3.0 (2021-09-20)
 ------------------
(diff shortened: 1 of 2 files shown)


jalasker (changes requested):
This change does not allow the phrase "mtl" to appear anywhere in the path name to the MTLs. Currently I store the sets of alt mtls in /global/cfs/cdirs/desi/survey/catalogs/SV3/LSS/altmtl/ with further contains the directory "copied_orig_mtls" for a directory in which I am trying to reproduce the original SV3 observations. 

This leads to this code looking for the file in /global/cfs/cdirs/desi/survey/catalogs/SV3/LSS/altops/debug_jl/copied_orig_opss

I could try replacing instances of "mtl" with "MTL" so it doesn't get caught by the replacement command in line 1417. However this could potentially break things on my end and would involve moving all of the alternate MTLs and bitweights that currently exist between directories.

sbailey:
@jalasker good catch, though it appears that this problem pre-dates this particular PR.

Let's update the line 1417 you reference:
```
opsdir = os.path.dirname(mtltilefn).replace("mtl", "ops")
```
to only replace the trailing "mtl" and not all cases of "mtl" instead of requiring users to not have "mtl" in their higher level directory path.

geordie666:
@jalasker: I made a change to simplify your construction of the directory names. Also, the `opsdir = os.path.dirname(mtltilefn).replace("mtl", "ops")` construction was used in one other place in the code other than line 1417, so I updated that too.

Can you check that my updated code still works with the alternate MTLs? If so, we can merge this PR and move on.

jalasker:
This works with the alternate MTLs. Merging pull request.

PR #770, opened 2021-11-08
fix code typos

segasai:
Fix 4 instances of variable names typos in cuts.py (probably most are harmless)


geordie666:
Thanks, Sergey. As you noted, most of these typos are in clauses that we don't actually trigger in the code. Some of the clauses were set up in a certain, formulaic way to facilitate working with mocks as well as with the real data.

The only typo that, I suppose, could have been harmful is the `gaiamag = gaiagmag.copy()` line. But, I think that will just lead to warnings on comparisons to `NaN` rather than affecting the actual selection.

In any case, feel free to merge this small PR _after_ documenting it in `doc/changes.rst`.

segasai:
Done.
Thanks, Adam.

PR #773, opened 2021-11-11
Add an ARCHIVEDATE column to the mtl done files and data model 

geordie666:
This PR adds an `ARCHIVEDATE` column to the data model for the MTL done files (`mtl-done-tiles.ecsv` and `scnd-mtl-done-tiles.ecsv`). 

It also includes the convenience script that was used to add the intial `ARCHIVEDATE` values to the MTL done files.

### DIFF
--- a/bin/add_archive_to_mtl
+++ b/bin/add_archive_to_mtl
@@ -1,0 +1,69 @@
+#!/usr/bin/env python
+
+import os
+import numpy as np
+from astropy.table import Table
+
+from desitarget.mtl import get_mtl_dir, get_mtl_tile_file_name, mtltilefiledm
+from desitarget.mtl import get_ztile_file_name
+from desitarget.geomask import match_to
+from desitarget import io
+
+from desiutil.log import get_logger
+log = get_logger()
+
+from argparse import ArgumentParser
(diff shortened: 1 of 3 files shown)


geordie666:
Hold off on merging this. I need to make a change.

geordie666:
Should be good now. No further changes expected.

sbailey:
Code looks good, with the caveat that I haven't tried running/testing it.

For consideration, I see that mtl-done-tiles.ecsv treats ZDATE and ARCHIVEDATE as strings, while tiles-specstatus.ecsv treats QANIGHT(=ZDATE), LASTNIGHT, and ARCHIVEDATE as integers, which is mildly inconvenient when matching the two files.  Options:
* mtl switches to int
* we live with the discrepancy
* (I'm not willing to switch tiles-specstatus to strings; I want to keep them as ints :)

I'll note that this discrepancy pre-dates this PR in ZDATE vs. QANIGHT+LASTNIGHT; ARCHIVEDATE just adds to the discrepancy.

geordie666:
As long as we're breaking backwards compatibility I can change those columns over to integers for ease-of-comparison. It'll take me a little time, though. So, we just have to remember to block MTL updates in the interim.

geordie666:
I made @sbailey's requested change. I'll merge this once tests pass and tag `2.0.0` so we can continue with "normal mode" MTL updates.

PR #775, opened 2021-11-21
Add SCND_TARGET column to BACKUP targets, even when nosec is passed

geordie666:
This PR adds a `SCND_TARGET` column to files of `BACKUP` targets, even when the `--nosecondary` argument is passed. This should ensure that `BACKUP` targets have the same data model as `BRIGHT` and `DARK` targets, even though there are typically no secondary targets in the `BACKUP` program.


geordie666:
This is a relatively minor PR. I'll merge it in about an hour so I can tag a new version of `desitarget` and make official files of targets for the `BACKUP` program.

PR #776, opened 2021-11-22
Skyhealpixs gaia v2

araichoor:
This PR is a small modification of the PR https://github.com/desihub/desitarget/pull/771.
It introduces the following changes in `skyhealpixs.py`:
- the healpix file basename now is `skymap*`, instead of `gaia_skymap*` (see [desi-survey 3412]);
- the `(nside, nest)` values are now arguments, defaulting to the previously hard-coded values `(64,True)` (in case we want to test a different healpix scheme);
- adds some log infos, expliciting the settings.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,7 @@ desitarget Change Log
 2.2.1 (unreleased)
 ------------------
 
-* Small modification on skyhealpixs.py [`PR #776`]:
+* Small modification on skyhealpixs.py [`PR #776`_]:
     * adjust the per-healpix file name
     * set (nside, nest) arguments, defaulting to (64, True)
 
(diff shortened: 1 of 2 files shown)


geordie666:
@araichoor: Can you please update the `doc/changes.rst` file to describe this PR?

araichoor:
thanks again for the reminder, I ve just done it.

geordie666:
I think you need an _ after 776 in `changes.rst`? In any case, if tests pass then I'll just merge this PR and deal with that syntax myself when I make the tag.

araichoor:
ah right, sorry: as we are not in rush, I can make the change now, let me know.

geordie666:
Good point, if you want to do that, great. As you say, I don't need to tag urgently. Thanks!

araichoor:
typo corrected; apologies for that..

Issue #779, opened 2021-12-12
ragged array deprecation warning with numpy 1.20.3


sbailey:
Fuji housecleaning after updating to latest versions of external dependencies:

`desitarget.geomask.circle_boundaries` generates a deprecation warning with numpy 1.20.3 when `nloc` is different per target:
```
/global/common/software/desi/users/sjbailey/desitarget/py/desitarget/geomask.py:764: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
  offdec = np.array([rad*np.sin(np.arange(ns)*2*np.pi/ns)
```

e.g. the unit tests try with `r=[ 1.001, 20.02 , 10.01 ], nloc=[ 2, 21, 11]`.  The list comprehension in
```python
    offdec = np.array([rad*np.sin(np.arange(ns)*2*np.pi/ns)
                       for ns, rad in zip(nloc, radius)]).transpose()
```
has 3 lists of length 2,21, and 11, but then numpy is unhappy about stuffing those into a single 2D array.  I think the downstream code is doing the right thing by guessing `dtype=object` which ends up returning ra, dec arrays of length 34=2+21+11.

@geordie666 please check and confirm that dtype=object is the intended behavior for this case where each ra,dec gets a different number of circles around them.

geordie666:
I'm fixing this now. `dtype=object` is _almost_ correct. But, not quite, for interesting reasons. If an array is ragged, `dtype=object` makes the (well, _my_) expected choice for a ragged-array construction:

```
np.array([np.arange(i) for i in [1, 2, 3]], dtype=object)
Out[]: array([array([0]), array([0, 1]), array([0, 1, 2])], dtype=object)
np.shape(_)
Out[]: (3,)
```
This result would be the same `for i in [2, 4, 7]` or `for i in [2, 21, 11]`, etc.

But, for arrays that aren't ragged, `dtype=object` makes a different choice:

```
np.array([np.arange(i) for i in [2, 2, 2]], dtype=object)                                                                                                   
Out[]: 
array([[0, 1],
       [0, 1],
       [0, 1]], dtype=object)
np.shape(_)
Out[]: (3, 2)
```
Fortunately, the unit tests caught this discrepancy and failed on the second (non-ragged) case.

[...]

PR #781, opened 2021-12-14
Fix ragged array warning in geomask.circle_boundaries()

geordie666:
This PR fixes a ragged array deprecation warning in [geomask.circle_boundaries()](https://github.com/desihub/desitarget/blob/6a2c55c6725279608e848fcbff43833998455548/py/desitarget/geomask.py#L736).

The fix was to convert the ragged arrays to `dtype=object` while being careful to convert them back to `dtype=float` when needed.

This addresses #779.


geordie666:
@sbailey: If possible I'd like to merge this soon to include it in the tag for running MTL reprocessing (i.e. I'm about to merger #780 and tag).

PR #785, opened 2021-12-24
Extra functionality for splitting random catalogs by HEALPixel

geordie666:
This PR is an extension to #783.

Certain random catalogs already include an `HPXPIXEL` column, which #783 also added. This PR checks for the existence of an `HPXPIXEL` column before attempting to add it.

In addition, as random catalogs can be split in various ways, this PR adds a new directory name — based on the input random catalog filename — to the output HEALPixel-split directory structure. This makes it easier to track the location of — and better facilitates applying the `desitarget.io` reading routines to — HEALPixel-split random catalogs.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,6 +5,8 @@ desitarget Change Log
 2.3.1 (unreleased)
 ------------------
 
+* Extra functionality for splitting randoms by HEALPixel [`PR #785`_]:
+    * Correctly process catalogs with an existing ``HPXPIXEL`` column.
 * Functionality to split random catalogs by HEALPixel [`PR #783`_].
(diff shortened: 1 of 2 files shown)


geordie666:
@ashleyjross: New versions of the HEALPixel-split randoms catalogs are available in:

`/global/cscratch1/sd/adamyers/forashley/dr9/2.3.0.dev5334/randoms/resolve/randoms-1-*`

These were built from the input catalogs:

`/global/cfs/cdirs/desi/target/catalogs/dr9/0.49.0/randoms/resolve/randoms-1-*`

geordie666:
I'm going to merge this once tests pass. We can always revisit this later if the new style of random catalogs isn't as desired.

PR #786, opened 2022-01-06
add [grz]fiberflux optional arguments to fix (now-broken) desisim/templates code

moustakas:
Our `desisim.templates` code has been broken for some time (see https://github.com/desihub/desisim/issues/553) because `desisim` lagged behind the changes made to `desitarget`.

This PR is the minimal set of changes to get `desisim.templates` working again, although I want to add some additional tests on the `desisim` side (companion PR is forthcoming) so let's not merge yet.

Issue #747, opened 2021-05-29
MPI_Select_Mock_Target


zhengyun0215:
Hi all , I try to use MPI_Select_Mock_Target to make mtl file. But I got this error:
```
ERROR:mpi_select_mock_targets:182:<module>: Pixels [28120] failed after 1.7 minutes

ERROR:mpi_select_mock_targets:185:<module>: Traceback (most recent call last):
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/master/py/desispec/parallel.py", line 453, in stdouterr_redirected
    yield # allow code to be run with the redirected output
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desitarget/master/bin/mpi_select_mock_targets", line 173, in <module>
    targets_truth(params, output_dir=args.output_dir, seed=rankseeds[i],
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desitarget/master/py/desitarget/mock/build.py", line 988, in targets_truth
    write_skies(output_dir, skytargets.as_array(), nside=nside, nsidefile=nside,
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desitarget/master/py/desitarget/io.py", line 1210, in write_skies
    subpseed += int(hpxlist[0])
IndexError: invalid index to scalar variable.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desitarget/master/bin/mpi_select_mock_targets", line 173, in <module>
    targets_truth(params, output_dir=args.output_dir, seed=rankseeds[i],
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/conda/lib/python3.8/contextlib.py", line 131, in __exit__
    self.gen.throw(type, value, traceback)
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/master/py/desispec/parallel.py", line 477, in stdouterr_redirected
    with open(fname) as infile:
FileNotFoundError: [Errno 2] No such file or directory: '/global/cscratch1/sd/yunzheng/BGSmock/data/runmock1/281/28120/build-64-28120.log_0'
```
Which means the sky.fits file couldn't be made and I couldn't make build-64-28120.log_0 in my catalog.
It seems that I could run MPI_Select code last week but after updating, it failed.
[...]

Issue #745, opened 2021-05-29
mpi_select_mock_targets


zdplayground:
Hi,
Recently I tried to go through the notebook https://github.com/desihub/tutorials/blob/master/FiberAssignMocks.ipynb in NERSC. But it fails when running target selection for mtl file generation.
Here is the error:

ERROR:mpi_select_mock_targets:182:<module>: Pixels [4528] failed after 4.0 minutes
ERROR:mpi_select_mock_targets:185:<module>: Traceback (most recent call last):
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desitarget/master/bin/mpi_select_mock_targets", line 173, in <module>
    targets_truth(params, output_dir=args.output_dir, seed=rankseeds[i],
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desitarget/master/py/desitarget/mock/build.py", line 905, in targets_truth
    data, MakeMock = read_mock(params['targets'][target_name], log, target_name,
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desitarget/master/py/desitarget/mock/build.py", line 172, in read_mock
    data = MakeMock.read(mockfile=mockfile, mockformat=mockformat,
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desitarget/master/py/desitarget/mock/mockmaker.py", line 3857, in read
    data = MockReader.readmock(mockfile, target_name=self.objtype,
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desitarget/master/py/desitarget/mock/mockmaker.py", line 1389, in readmock
    gmmout = self.sample_GMM(nobj, target=target_name, isouth=isouth,
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desitarget/master/py/desitarget/mock/mockmaker.py", line 669, in sample_GMM
    samp[south] = _samp_iterate(samp[south], target=target, south=True, rand=rand,
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desitarget/master/py/desitarget/mock/mockmaker.py", line 647, in _samp_iterate
    itarg = colorcuts_function(gflux=gflux, rflux=rflux, zflux=zflux,
[...]

Issue #735, opened 2021-05-19
z-band extinction correction


mehdirezaie:
Hi! I am interested in the extinction coefficients for W1 and W2, used for the calculation of MW_Transmission attributes. I found this function in the repository:
https://github.com/desihub/desitarget/blob/acefce624846c54ac08261a85b297e99ba226da0/py/desitarget/mock/mockmaker.py#L321

However, I noticed that the z-band coefficient does not quite match the value reported in the documentation (Galactic Extinction Coefficients): https://www.legacysurvey.org/dr9/catalogs

moustakas:
That's a typo in the z-band coefficient, @mehdirezaie, thanks (it should be 1.211, not 1.221).

But note that all the "dust" code has been consolidated into `desiutil` now (see https://github.com/desihub/desiutil/pull/169) and so I'll be refactoring this part of the code to use those extinction coefficients (which are consistent with what we used for the imaging).

moustakas:
Fixed in #788.

PR #788, opened 2022-01-13
restore select_mock_targets to working order

moustakas:
[WIP]

Building on #786, and on the heals of the major fixes in the template-generating code in https://github.com/desihub/desisim/pull/559, the purpose of this PR is to restore `select_mock_targets` to working order (it's been broken for many months, if not a year) and to address a number of open bugs and issues, including #735, #745, #747, and others.


moustakas:
@geordie666 @weaverba137 @sbailey this PR is now ready for review. However, I need help setting up the unit tests to test the mock code (specifically, `test/test_mock_build`), which will need to bring in `desisim@master` as a dependency (at least until we cut a new tag).

For the record, the mock unit tests pass with a local check-out of this branch using 
```
source /global/cfs/cdirs/desi/software/desi_environment.sh master
```

A couple quick notes:

* I was originally developing and testing this branch on my laptop but I ran into this issue using `desitarget.internal.sharedmem`--https://github.com/rainwoodman/sharedmem/issues/24
However, when I switched to NERSC I didn't run into this issue. Nevertheless, it'd be good to keep an eye on this change since most of the production desitarget code uses this module.
 
* This PR commits a new notebook which develops Gaussian mixture models for the various target classes using DR9 photometry, `doc/nb/gmm-dr9.ipynb`, which may be of interest to others and for other projects.

* There are many(!) places where `select_mock_targets` could be improved based on actual DESI data and updated cosmological mocks. However, that work is outside the scope of this PR, whose goal was to just get everything back into a working state.

weaverba137:
For the record: it looks like desitarget already depends on desisim, and of course, desisim already depends on desitarget.  Thus we have a circular dependency.  It hasn't caused problems so far, but those are always a potential source of pain.

For the sharedmem issue, it would be interesting to compare what version of Python you are using on your laptop to the version at NERSC.  If you're using the DESI software stack at all, I think that is still Python 3.8, which would appear to contradict the ticket.

moustakas:
> For the sharedmem issue, it would be interesting to compare what version of Python you are using on your laptop to the version at NERSC. If you're using the DESI software stack at all, I think that is still Python 3.8, which would appear to contradict the ticket.

On my laptop I was using `Python 3.8.12` vs `Python 3.8.3` for the DESI software stack. I didn't dig any deeper into the issue other than to celebrate when everything "just worked" after I switched from my laptop to NERSC.

doc/changes.rst
```diff
@@ -5,6 +5,7 @@ desitarget Change Log
 2.3.1 (unreleased)
 ------------------
 
+* Restore ``select_mock_targets`` to a working state [`PR #788`].
```
geordie666, line 8: You're missing the trailing underscore on 
```
[`PR #788`]
[`PR #788`_]
```
that would link this to the url for the PR.

py/desitarget/io.py
```diff
@@ -1293,7 +1293,7 @@ def write_skies(targdir, data, indir=None, indir2=None, supp=False,
         else:
             subpseed = 718
         if hpxlist is not None:
-            subpseed += int(hpxlist[0])
+            subpseed += int(np.atleast_1d(hpxlist)[0])
```
geordie666, line 1296: This looks sensible, but the docstring does specify that `hpxlist` should be a list, so the situation you're guarding against isn't meant to occur. Please update the docstring to specify that `hpxlist` _can_ be passed as an integer.

geordie666 (changes requested):
@moustakas: As you mention, most of your changes are specific to the mocks, which I didn't review too carefully. For the more general parts, I've added two very minor comments for you to address. I'm happy to sign off on this, though.

doc/changes.rst
moustakas: Done in https://github.com/desihub/desitarget/pull/788/commits/83c8a2421dcb6f91550b717804e57c7985cfcf0d.

py/desitarget/io.py
moustakas: Done in https://github.com/desihub/desitarget/pull/788/commits/83c8a2421dcb6f91550b717804e57c7985cfcf0d.

PR #799, opened 2022-11-18
Refresh test infrastructure

### LINKED ISSUE #798: Coverage of sv1, sv2, sv3 has dropped dramatically
As reported in #797, coverage dropped dramatically after merging that PR.  The culprit is that `desitarget.sv1`, `desitarget.sv2`, and `desitarget.sv3` tests are not being run or otherwise skipping that code. For example coverage of `desitarget.sv1` dropped from about 85% to about 17%.

No other components of desitarget show a significant change in coverage.

weaverba137:
This PR closes #798 by refreshing the test infrastructure.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -17,14 +17,14 @@ jobs:
             fail-fast: true
             matrix:
                 os: [ubuntu-latest]
-                python-version: [3.8]  # fitsio does not like Python < 3.7, astropy does not like Python < 3.8, might be too soon for 3.9.
-                astropy-version: ['==4.0.1.post1', '==5.0']  # everest & fuji versions
-                fitsio-version: ['==1.1.2', '==1.1.6']  # everest & fuji versions
+                python-version: [3.9]
+                astropy-version: ['==5.0', '<6']  # fuji version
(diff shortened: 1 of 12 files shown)


weaverba137:
@geordie666, I think I found the answer: pytest>=7 and pytest-cov>=4 give lower coverage than pytest<7; pytest-cov<4 which is what we have at NERSC.  I've pinned to those older versions, and the coverage value now matches the NERSC value.

`desitarget/svX/svX_cuts.py` is imported by `importlib.import_module` instead of a standard `import` statement, & this seems to be incompatible with how recent versions of `pytest-cov` counts coverage.  Further research will be needed to fix this long-term, but for now pinning the versions works.

geordie666 (approved):
Thanks @weaverba137, I appreciate you running carefully through some of these abstruse cases. This all looks good to me, so feel free to merge when you're ready.

FYI (in case you were thinking of tagging now that tests pass): I'll need to make a couple of updates in the next few days to run randoms for DR10 of the Legacy Surveys. I'll make a new tag when I've completed those updates.

weaverba137:
I don't actually need to tag this immediately, so go ahead and tag based on your own changes when you're ready.

PR #803, opened 2023-08-23
Fix bug where data types are altered when override MTLs are empty

geordie666:
This PR fixes a bug introduced when incorporating the MTL "override" ledgers into reprocessing.

When an override ledger didn't exist, or was otherwise empty, the data types for some columns were updated to standard types for an empty numpy array. This could then conflict with the data types for an overall ledger. 

The fix is to never use standardized column updates for empty override ledgers.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,7 @@ desitarget Change Log
 2.6.1 (unreleased)
 ------------------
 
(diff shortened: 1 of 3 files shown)


sbailey:
Update looks fine; thanks for sorting out photutils dependency pin for tests.  OK to merge this when ready.  Please open a separate ticket about the photutils dependency stuff.

Issue #804, opened 2023-08-23
Clean up unit tests


geordie666:
The `desitarget` unit tests are currently flagging some deprecated modules and functions. Clean those up.

geordie666:
In particular, the `photutils` dependency is now pinned to version `1.6.0`. We should revisit the best way to maintain this dependency.

moustakas:
I think #829 addresses at least some of this ticket. Note that I didn't see any issues with `photutils` there.

geordie666:
I think we just need to retain a copy of `photutils` for some of the core code that produces sky-location targets. But, the DESI stack doesn't actually ostensibly _support_ `photutils`.

But, if the `photutils` dependency is still working then maybe we don't worry about it until it breaks? It's currently being maintained as an `astropy` affiliated package, so it likely won't be a problem in the short term.

geordie666:
Oh, good point. Now that you've updated the other dependencies it may not be necessary to pin `photutils`. Certainly the unit tests cover `photutils`, so this would be worth investigating.

I don't think there's anything special about `photutils=1.6.0` other than that it was "the version that worked in our pre-numpy-2 environment."

Issue #802, opened 2023-08-02
Add "columns" option to desitarget.io.read_mtl_in_hp


jalasker:
Feature Request: Add an option to read_mtl_in_hp to select only certain columns to load into memory analogously to the read_targets_in_tiles "columns" option.

geordie666:
Addressed in #805.

PR #805, opened 2023-08-29
Add an option to only read a subset of columns from the HEALPixel-split MTLs

geordie666:
This PR adds an option in `desitarget.io.read_mtl_in_hp()` to facilitate reading a subset of columns from an MTL rather than the entire ledger. This should address #802.

This PR also addresses some deprecation warnings, which should at least partially address #804.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,13 @@ desitarget Change Log
 2.6.2 (unreleased)
 ------------------
 
-No changes yet.
+* Option to read subset of columns from HEALPix-split MTLs [`PR #804`_].
+    * Specifically in :func:`io.read_mtl_in_hp()`.
+    * Partial fix for `issue #802`_.
+    * Also address some deprecation warnings.
+
+.. _`issue #802`: https://github.com/desihub/desitarget/issues/802
+.. _`PR #804`: https://github.com/desihub/desitarget/pull/804
 
 2.6.1 (2023-08-23)
 ------------------
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,13 +5,15 @@ desitarget Change Log
 2.6.2 (unreleased)
 ------------------
 
-* Option to read subset of columns from HEALPix-split MTLs [`PR #804`_].
+* Option to read subset of columns from HEALPix-split MTLs [`PR #805`_].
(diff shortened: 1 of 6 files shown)


geordie666:
@jalasker: Before I merge this, could you check that you can indeed successfully use the new `columns` kwarg in `desitarget.io.read_mtl_in_hp()` as requested?

jalasker:
I tested this function and there is some slightly unexpected (to me) behavior. 

When you are requesting RA and Dec as columns, the function works properly. However, if you do not provide RA and Dec as columns, there is a step at the end of the function (https://github.com/desihub/desitarget/blob/ADM-cols-mtl-hp/py/desitarget/io.py#L3238)  which "restrict the targets to the actual requested HEALPixels" and which requires RA and Dec to be present. 

The function will raise a ValueError that there is no RA field in the event that RA/Dec are not listed as columns. 

I'm assuming this line is so you can provide healpixels with a different size than the ones the MTLs are partitioned into natively, since if the nside = 32, it shouldn't be possible for any trimming to change anything. 

A couple of changes which I think might work here in a rough order based on how intuitive they'd be for me to use combined with what I think would be easiest to implement. 
1. only evaluate the is_in_hp function if nside != filenside. This avoids the problem entirely if you are reading in using healpixels/pixlist simply as an alias for which MTL file you want to read in. 
2.  Check if "columns" contains RA/Dec and if not (2a) add them to the list of columns or (2b) raise the exception before doing any file loading/processing. 
3. Combine 1 and 2 by only doing the check in 2 if nside != filenside
4. Store RA and Dec as separate arrays/tables from anything that will be returned and evaluate is_in_hp regardless of the considerations above.
5. Number 4, but only do that if RA/Dec aren't in columns

jalasker:
One more semi-related comment: the 'columns' argument can now be propagated through "read_targets_in_hp" to "read_mtl_in_hp" when mtl == True.

geordie666:
Thanks for the careful and thorough review @jalasker. I chose your option (2a) and always add the RA, DEC columns even if they aren't requested in the `columns` keyword. I think this is defensible given that the coordinates are a crucial aspect of working with HEALPixels. The docstring of `read_mtl_in_hp()` also now documents this behavior.

Please run your check again and let me know if you want further updates.

jalasker:
Works as expected and ready to merge.

Issue #807, opened 2023-09-19
Add sanity check that archiving is correct


geordie666:
Add something like this to `run_mtl_loop`:

https://github.com/desihub/desisurveyops/issues/141

PR #808, opened 2023-10-06
Add check to run_mtl_loop that archived directories match the tiles-specstatus file

geordie666:
This PR adds a `check_archiving()` function that is used in `run_mtl_loop` to test whether archived tile directories match the `ARCHIVEDATE` column in the `tiles-spectstatus.ecsv` file. The check is restricted to bright and dark tiles from the Main Survey.

See #807 and https://github.com/desihub/desisurveyops/issues/141 for more context.

`run_mtl_loop` now also includes a `--noarchivecheck` flag that can be used to bypass the archiving check (e.g., to speed up simulations or the alt-MTL process).

I tested this pretty extensively by:

- Creating a mock up of the `archive` directories in `/pscratch/sd/a/adamyers/arxivsandbox/tiles/archive/`.
- Copying the `tiles-specstatus.ecsv` file to `/pscratch/sd/a/adamyers/arxivsandbox/ops/`.
- Moving the archive directories for tiles `20994` and `25442` to `/pscratch/sd/a/adamyers/arxivsandbox/tiles/original/`.
- Removing tile `3333` from `/pscratch/sd/a/adamyers/arxivsandbox/ops/tiles-specstatus.ecsv` (the original file is retained at `/pscratch/sd/a/adamyers/arxivsandbox/ops/tiles-specstatus.ecsv.orig`).
- Then running, e.g., ` run_mtl_loop DARK --zcatdir /pscratch/sd/a/adamyers/arxivsandbox/ --mtldir /pscratch/sd/a/adamyers/arxivsandbox/mtl`

The resulting error message was:
```
ERROR:mtl.py:494:check_archiving: archived tiles not in specstatus file: [3333]; tiles in specstatus file that aren't archived: [20994 25442]
```

@araichoor: I'd appreciate a second pair of eyes on this PR before I tag `desitarget` and we make the check a standard part of daily ops. 

I don't think this is critical to resolve on the timescale of the long weekend, though!

### DIFF
--- a/bin/run_mtl_loop
+++ b/bin/run_mtl_loop
@@ -41,7 +41,8 @@ if ns.obscon == "BACKUP" and ns.survey == "main":
     log.error(msg)
 
 if not ns.noarchivecheck:
-    check_archiving(zcatdir=ns.zcatdir, mtldir=ns.mtldir)
+    check_archiving(ns.obscon, survey=ns.survey, zcatdir=ns.zcatdir,
+                    mtldir=ns.mtldir)
 
 scndstates = [False]
 
--- a/bin/run_mtl_loop
+++ b/bin/run_mtl_loop
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-from desitarget.mtl import loop_ledger
+from desitarget.mtl import loop_ledger, check_archiving
 from desiutil.log import get_logger
 log = get_logger()
 
@@ -30,6 +30,8 @@ ap.add_argument("-nosec", "--nosecondary", action='store_true',
                 there are no secondaries to process")
 ap.add_argument("--reprocess", action='store_true',
                 help="Run reprocessed tiles (instead of the standard loop)")
+ap.add_argument("--noarchivecheck", action='store_true',
+                help="Do NOT check if tiles-specstatus and archived tiles match")
 
 ns = ap.parse_args()
 
@@ -37,6 +39,9 @@ if ns.obscon == "BACKUP" and ns.survey == "main":
     msg = "Updating BACKUP ledgers is not yet supported for the Main Survey!!!"
     raise RuntimeError(msg)
     log.error(msg)
+
+if not ns.noarchivecheck:
+    check_archiving(zcatdir=ns.zcatdir, mtldir=ns.mtldir)
 
 scndstates = [False]
 
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -6,6 +6,7 @@ desitarget Change Log
 ------------------
 
 * Add check that archived tiles match tiles-specstatus file [`PR #808`_].
+* Reproducible output from :func:`io.read_targets_in_hp()` [`PR #806`_].
 * Option to read subset of columns from HEALPix-split MTLs [`PR #805`_].
     * Specifically in :func:`io.read_mtl_in_hp()`.
     * Addresses `issue #802`_.
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,7 @@ desitarget Change Log
 2.6.2 (unreleased)
 ------------------
 
-* Reproducible output from :func:`io.read_targets_in_hp()` [`PR #806`_].
+* Add check that archived tiles match tiles-specstatus file [`PR #808`_].
 * Option to read subset of columns from HEALPix-split MTLs [`PR #805`_].
     * Specifically in :func:`io.read_mtl_in_hp()`.
     * Addresses `issue #802`_.
@@ -16,6 +16,7 @@ desitarget Change Log
 .. _`issue #804`: https://github.com/desihub/desitarget/issues/804
 .. _`PR #805`: https://github.com/desihub/desitarget/pull/805
 .. _`PR #806`: https://github.com/desihub/desitarget/pull/806
+.. _`PR #808`: https://github.com/desihub/desitarget/pull/808
 
 2.6.1 (2023-08-23)
 ------------------
--- a/py/desitarget/mtl.py
+++ b/py/desitarget/mtl.py
@@ -431,21 +431,67 @@ def check_archiving(zcatdir=None, mtldir=None):
       "archive" sub-directory of the redshift catalog directory matches
       the `ARCHIVEDATE` recorded in the `tiles-specstatus.ecsv` file.
     - The check is limited to BRIGHT/DARK main survey tiles.
+    - Based on work by Anand Raichoor.
     """
-    # ADM use the default $ZCAT_DIR, if needed.
-    zcatdir = get_zcat_dir(zcatdir)
-    # ADM grab the tile-based sub-directories in the archive directory.
-    archivedir = os.path.join(zcatdir, "tiles", "archive")
-    fns = sorted(glob(os.path.join(archivedir, "*", "*")))
-
     # ADM grab tile files, use default environment variables, if needed.
     ztilefn = get_ztile_file_name(survey="main")
     mtldir = get_mtl_dir(mtldir)
     opsdir = os.path.join(os.path.dirname(mtldir), "ops")
     ztilefn = os.path.join(opsdir, ztilefn)
-    # ADM need to check which specstatus tiles are Main Survey tiles.
+
+    # ADM need to check which tiles are in Main Survey and bright/dark.
     maintilefn = ztilefn.replace("specstatus", "main")
+    maintiles = Table.read(maintilefn)
+    dorb = (maintiles["PROGRAM"] == "DARK") | (maintiles["PROGRAM"] == "BRIGHT")
+    maintiles = maintiles[dorb]
+
+    # ADM grab the tile-based sub-directories in the archive directory.
+    # ADM use the default $ZCAT_DIR, if zcatdir was not passed.
+    zcatdir = get_zcat_dir(zcatdir)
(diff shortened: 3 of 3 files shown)


geordie666:
Pinging @araichoor as a reminder. This still probably still isn't urgent yet, but it would be nice to merge this this week.

araichoor:
sorry that fell off my radar.
I ll have a look, and comment what I can.

bin/run_mtl_loop
```diff
@@ -38,6 +40,9 @@ if ns.obscon == "BACKUP" and ns.survey == "main":
     raise RuntimeError(msg)
     log.error(msg)
 
+if not ns.noarchivecheck:
+    check_archiving(zcatdir=ns.zcatdir, mtldir=ns.mtldir)
```
araichoor, line 44: for consistency, would it make sense to propagate in the call to `check_archiving()` the `ns.survey` and `ns.obscon` argument?
for instance if `run_mtl_loop` is called for main/bright, then the check is done for main/bright only?

I am aware that cases other than main/bright and main/dark may never be used -- and `check_archiving()` would be irrelevant for those; so maybe you can just request in `check_archiving()` that only those cases are valid?

doc/changes.rst
```diff
@@ -5,7 +5,7 @@ desitarget Change Log
 2.6.2 (unreleased)
 ------------------
 
-* Reproducible output from :func:`io.read_targets_in_hp()` [`PR #806`_].
```
araichoor, line 8: is it intended that the line for PR806 is removed here?

py/desitarget/mtl.py
```diff
@@ -412,6 +412,93 @@ def get_mtl_ledger_format():
     return ff
 
 
+def check_archiving(zcatdir=None, mtldir=None):
```
araichoor, line 415: as suggested above, what about adding additional `survey="main"` and a `obscon="DARK"` arguments?

py/desitarget/mtl.py
```diff
@@ -412,6 +412,93 @@ def get_mtl_ledger_format():
...
+    -----
+    - The basic check is that the latest archive date for tiles in the
+      "archive" sub-directory of the redshift catalog directory matches
+      the `ARCHIVEDATE` recorded in the `tiles-specstatus.ecsv` file.
+    - The check is limited to BRIGHT/DARK main survey tiles.
+    - Based on work by Anand Raichoor.
+    """
+    start = time()
+    log.info("Checking archived tiles match tiles-specstatus file...")
+
```
araichoor, line 438: then you could add here something like:
```
if (survey != "main") or (obscon not in ["BRIGHT", "DARK"]):
    msg = "survey={} and obscon={} => only survey='main' and obscond='BRIGHT' or 'DARK' are currently handled".format(survey, obscon)
    log.error(msg)
    raise ValueError(msg)

File: py/desitarget/mtl.py [hunk 5]
```diff
@@ -412,6 +412,93 @@ def get_mtl_ledger_format():
...
+    # ADM grab tile files, use default environment variables, if needed.
+    ztilefn = get_ztile_file_name(survey="main")
```
araichoor, line 440: if adding a `survey` argument, you ll have to propagate the argument value here.

py/desitarget/mtl.py
```diff
@@ -412,6 +412,93 @@ def get_mtl_ledger_format():
...
+    mtldir = get_mtl_dir(mtldir)
+    opsdir = os.path.join(os.path.dirname(mtldir), "ops")
+    ztilefn = os.path.join(opsdir, ztilefn)
+
+    # ADM need to check which tiles are in Main Survey and bright/dark.
+    maintilefn = ztilefn.replace("specstatus", "main")
+    maintiles = Table.read(maintilefn)
+    dorb = (maintiles["PROGRAM"] == "DARK") | (maintiles["PROGRAM"] == "BRIGHT")
+    maintiles = maintiles[dorb]
+
```
araichoor, line 450: same, here you d have to adjust for the `obscon` argument, if choosing that approach.

py/desitarget/mtl.py
```diff
@@ -412,6 +412,93 @@ def get_mtl_ledger_format():
...
+    # ADM grab tile files, use default environment variables, if needed.
+    ztilefn = get_ztile_file_name(survey="main")
+    mtldir = get_mtl_dir(mtldir)
+    opsdir = os.path.join(os.path.dirname(mtldir), "ops")
+    ztilefn = os.path.join(opsdir, ztilefn)
+
+    # ADM need to check which tiles are in Main Survey and bright/dark.
+    maintilefn = ztilefn.replace("specstatus", "main")
+    maintiles = Table.read(maintilefn)
+    dorb = (maintiles["PROGRAM"] == "DARK") | (maintiles["PROGRAM"] == "BRIGHT")
```
araichoor, line 448: side comment: do we want to also cut on `maintiles["IN_DESI"]` here?
I m not sure, but I mention in case.

araichoor (commented):
ok, so here are minor comments/suggestions, which you can ignore if you think it s better to do so.

bin/run_mtl_loop
geordie666: I think this is a smart idea. 

I added a required `obscon` argument and a `survey=survey` kwarg to `check_archiving()` and made the appropriate changes to propagate these arguments throughout the function.

doc/changes.rst
geordie666: Nice catch! This was a mistake on my part. I added this line back into the `changes.rst` file.

bin/run_mtl_loop
geordie666: Also, `run_mtl_loop` now additionally passes `obscon` and `survey` to the `check_archiving()` function.

py/desitarget/mtl.py
geordie666: Addressed, see my response to your first comment, above.

py/desitarget/mtl.py
geordie666: Addressed, see my response to your first comment, above.

py/desitarget/mtl.py
geordie666: Addressed, see my response to your first comment, above.

py/desitarget/mtl.py
geordie666: Addressed, see my response to your first comment, above.

py/desitarget/mtl.py
geordie666: I _don't_ think we need an additional check on `maintiles["IN_DESI"]`, here. I think the result should be the same either way. I'd appreciate an opinion from @schlafly, though.

geordie666:
Thanks for your careful review @araichoor. I addressed all of your comments. I went in the direction you suggested of splitting out the archiving check by `main/bright` and `main/dark`.

I've left one request for an opinion from @schlafly as part of my response to the final comment of your review. Otherwise I think this PR should be ready for merging.

I rechecked the code using the same test as detailed at the [top of this thread](https://github.com/desihub/desitarget/pull/808#issue-1930977746).

araichoor:
thanks for implementing my suggestions!

I just thought of a possible minor drawback of making the check on {survey,obscon} only, but I think there s nothing to change in the code.

I thought about the following:
- the dailyops person runs `run_mtl_loop` for the {main,bright}, and the archiving check fails;
- then one may want to verify the archiving situation for {main,dark} also, without running `mtl_loop_run` for {main,dark} (because if only {main,bright} tiles have archiving issue, {main,dark} would run smoothly, and for the time to resolve the issue, bright/dark would be "out-of-sync" from an mtl-point-of-view; but maybe that s fine because we d then sort out + fix the main,bright case right away);
- but if we want to check the archiving for {main,dark} without running `run_mtl_loop`, one can just run from python:
```
from desitarget.mtl import check_archiving
check_archiving("DARK")
```
(and we could put that instruction in the dailyops wiki).

geordie666:
I think "dark" and "bright" being "out-of-sync" is fine, though, right? 

As far as I know we treat "bright" and "dark" as completely separate surveys? So, although we've never just run `run_mtl_loop DARK` or `run_mtl_loop BRIGHT` (because of the way the instructions are structured on the wiki) there's nothing, in principal, that's actually _wrong_ with just running one or the other?

Sometimes we run an MTL update where there are, e.g., some new dark tiles but no new bright tiles and the `run_mtl_loop BRIGHT` command just does nothing in that case.

Another way of looking at this (which is why I liked your suggestion) is that we might want to proceed with an MTL update for dark tiles even if there's an archiving issue for bright tiles.

Am I missing something that intertwines the bright and dark parts of the survey?

araichoor:
'_I think "dark" and "bright" being "out-of-sync" is fine, though, right?_' => yes, sure, no big deal actually

'_Am I missing something that intertwines the bright and dark parts of the survey?_' => no you re right!

I was just slightly worried that we end up in a case where `run_mtl_loop` has been run for one program but not for another one; but you re totally right that those are fully disjoint, and that this actually already happens if we do an mtl update but one program has no new tiles.

sorry for the noise!

py/desitarget/mtl.py
schlafly: Let's not add an additional check on IN_DESI at the moment.  It's not clear to me what we want, here, but we probably do want to make noise about any files in the archive directory that aren't recorded as such in the specstatus file.

PR #809, opened 2023-12-03
Add functionality to write all columns from csv files out to Gaia DR3 FITS files

geordie666:
This PR has two main purposes:

- It adds functionality to, optionally, write out the **full** set of _Gaia_ columns to the _Gaia_ DR3 FITS files instead of just the subset used for DESI targeting.
- It adds a new higher-priority bit for the diffuse interstellar bands BACKUP program (see DESI project 398).

While writing this PR, I discovered a bug that should be included in the DESI DR1 paper: 

The string form of Booleans changed for the _Gaia_ csv files moving from _Gaia_ DR2 to _Gaia_ EDR3, necessitating [an update to how these strings were parsed](https://github.com/desihub/desitarget/commit/75251191e3ddea51f45501817f6f6c51f3f32be0). This update was not made, however, meaning that no Boolean columns could be interpreted as ``True`` for _Gaia_ EDR3 or _Gaia_ DR3. Therefore, all EDR3 or DR3 _Gaia_ FITS files made by versions of `desitarget` prior to this PR had Boolean columns set entirely to ``False``. The only column this would have affected for DESI targeting would be the _Gaia_ `DUPLICATED_SOURCE` column — i.e. effectively no source in DESI targeted using _Gaia_ EDR3 (see section 4.1.4 of the [desitarget paper](https://ui.adsabs.harvard.edu/abs/2023AJ....165...50M/abstract)) would have been masked on `DUPLICATED_SOURCE`.

This PR also adds a `boolfix` kwarg to `gaia_csv_to_fits()` allowing this bug to be corrected when generating _Gaia_ EDR3 files. The default behavior of `gaia_csv_to_fits()`  subsequent to this PR is:

- To _always_ correct for _Gaia_ DR3 (and we should probably update the relevant _Gaia_ DR3 FITS files at NERSC).
- But to **not** correct the _Gaia_ EDR3 files by default. The reason for this choice is that the _Gaia_ DR3 files were never used for DESI targeting, but the _Gaia_ EDR3 files were. 

The bug does not affect _Gaia_ DR2 files, either way.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,6 +5,8 @@ desitarget Change Log
 2.6.2 (unreleased)
 ------------------
 
+* Option to write all columns for Gaia DR3 FITS files [`PR #809`_].
+    * Also adds a new higher-priority bit for the BACKUP DIB program.
 * Add check that archived tiles match tiles-specstatus file [`PR #808`_].
 * Reproducible output from :func:`io.read_targets_in_hp()` [`PR #806`_].
 * Option to read subset of columns from HEALPix-split MTLs [`PR #805`_].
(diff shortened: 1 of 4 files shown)


geordie666:
I'll leave this PR open until tomorrow evening and then I'll merge and make a new tag, mainly so we can deploy the work from PR #808 as part of daily operations.

PR #820, opened 2024-05-06
Add function to match RA/Dec positions to Main Survey targets

geordie666:
This PR adds a function to match RA/Dec positions to Main Survey targets (`desitarget.secondary.match_to_main_survey()`).

The purpose of this function is to provide a straightforward way for proposers of spare-fiber programs to check how many of their likely targets already match main-survey targets.

### DIFF
--- a/py/desitarget/secondary.py
+++ b/py/desitarget/secondary.py
@@ -129,7 +129,7 @@ def match_to_main_survey(ras, decs, sep=1.):
     Notes
     -----
     - Must have set up the correct environment by running, e.g.,
-      source /global/cfs/projectdirs/desi/software/desi_environment.sh
+      source /global/common/software/desi/desi_environment.sh
     - Prints summary statistics of the number of matches to screen.
     """
     t0 = time()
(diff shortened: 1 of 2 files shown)


geordie666:
This PR adds a standalone function with no other dependencies so should be uncontroversial. I'll merge it soon in preparation for an impending call for new spare-fiber targets.

py/desitarget/secondary.py
```diff
@@ -102,6 +103,113 @@
...
+            BRIGHT: ``True`` for locations matching a bright-time target.
+            DARK: ``True`` for locations matching a dark-time target.
+            BRIGHTSEC: ``True`` for matching a bright-time secondary.
+            DARKSEC: ``True`` for matching a dark-time secondary.
+
+    Notes
+    -----
+    - Must have set up the correct environment by running, e.g.,
+      source /global/cfs/projectdirs/desi/software/desi_environment.sh
+    - Prints summary statistics of the number of matches to screen.
```
moustakas, line 133: I think the preferred, current location of the desi software environment is `/global/common/software/desi/desi_environment.sh`.

moustakas:
Nice. Quick comment to consider: Would it be useful to add the targeting columns (`DESI_TARGET`, `BGS_TARGET`, etc.) as well as `TARGETID` (for matches) so the user can understand *why* an object was targeted in the main survey?

geordie666:
Yep, good idea. I'll add those.

geordie666:
I've added the `TARGETID`s to the output file. I didn't add the other columns @moustakas suggested: The issue is that all of the `_TARGET` columns can be different for each of the `dark`, `bright`, `secondary/dark`, `secondary/bright` combinations, so the extra information in the file quickly proliferates.

Users interested in provenance should be able to use the `TARGETID`s to look up extra information.

Issue #818, opened 2024-04-29
Include an option to read the MTLs using "ascii.ecsv" instead of "ascii.basic"


geordie666:
When running the alt-MTLs for mocks, we occasionally trip on reading a file. This appears to be fixed by adopting the `ascii.ecsv` reader instead of `ascii.basic`. So, it would be useful to propagate a kwarg throughout desitarget.io that allows this option when reading MTLs.

geordie666:
Addressed in #822.

PR #822, opened 2024-05-17
Allow different astropy Table.read() reader options when accessing MTLs

geordie666:
This PR updates the desitarget I/O code for MTLs to allow different reader options to be passed to the core `Table.read()` reader function.

Context:
 - We originally fixed the reader for MTLs to `ascii.basic`, as this reader is ~3-4x faster than, e.g., `ascii.ecsv`. 
   * The `ascii.basic` option will remain the default after this PR.
 - The alt-MTL/mocks work is occasionally tripping on formatting errors that could be related to strict parsing using `ascii.ecsv` instead of `ascii.basic`.
   * This has never been a problem for the real MTLs, though.

This PR should address issue #818.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,11 +5,15 @@ desitarget Change Log
 2.7.1 (unreleased)
 ------------------
 
+* Allow different Table reader options when reading MTLs [`PR #822`_].
+    * Addresses `issue #818`_.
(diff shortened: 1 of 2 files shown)


geordie666:
Unit tests will fail in this branch, but have been fixed in #814.

Issue #821, opened 2024-05-10
select_targets can select north (south) targets for objects with south (north) photometry


moustakas:
Off-list, @geordie666 and I have been tracking down a nefarious, confusing bug which we finally think we understand.

This came up in the context of the Y3 template work (see https://github.com/desihub/rrevaluator and https://github.com/desihub/redrock/issues/279). 

As part of this work, we're using the set of ~19,000 objects observed as part of sv1 with VI redshifts. However, we want to focus on the subset of those  (~9000) objects which *would have been* observed as part of the main survey.

Initially, we simply matched the sample to the catalogs in `/global/cfs/cdirs/desi/target/catalogs/dr9/1.1.1/targets/main/resolve/{dark,bright}` and that worked fine.

However, for reasons that I don't recall, I decided I wanted more control over the results and instead: 
1. Gathered the Tractor photometry for these targets using `desispec.io.photo.gather_tractorphot`; and
2. Ran `desitarget.cuts.select_targets` myself.

However, this procedure results in 25 ostensibly "main-survey" targets which are not in the `dr9/1.1.1` target catalogs. Bad!
```
<Table length=25>
    TARGET_RA          TARGET_DEC          TARGETID
     float64            float64             int64
------------------ ------------------ -----------------
35.813700262224884 -5.206945199328692 39627658597767428
 35.95611507231263 -5.359925700877448 39627658601958684
 37.53560160975776 -4.511371192758242 39627676696188831
150.15391038279543 1.8945674771823369 39627835563838694
149.20043940392307 2.5159829263416227 39627847626657859
151.38949868826157 2.6027204998624245 39627847664405076
 149.6921862712477  2.892975597340467 39627859714641687
106.00075653648892  54.95950884025398 39633321159821042
105.75946082450504  55.14225192191247 39633324632703364
107.14453820974144  55.22266773697883 39633324645287048
106.18602683962725  55.56817690184516 39633328084619301
105.22266236194238  55.75898232607414 39633331507171632
108.30252293795569 55.844273985288034 39633331536531297
104.17873124885628   56.1161651740171 39633334904557083
104.30242744427088  55.93876986965283 39633334908748247
107.89623333467183  55.97002538469384 39633334942302972
109.46876774084396 55.877223289905885 39633334954887818
105.13664946129848  56.55590539431955 39633341665771821
105.51332906710496 56.413102795203805 39633341665775786
106.07870149933537  56.57395545784924 39633341674160813
[...]

PR #823, opened 2024-05-20
When resolving, only set bits for targets that pass northern (southern) cuts in northern (southern) imaging.

geordie666:
This PR addresses #821 by only strictly setting bits for targets that pass the northern cuts in northern imaging, and vice versa, when resolving is turned on.

I tested this against the objects mentioned by @moustakas in #821, and the offending target bits are no longer set:

```
import fitsio
from desitarget.cuts import select_targets

targetids = [39627658597767428, 39627658601958684, 39627676696188831,
39627835563838694, 39627847626657859, 39627847664405076, 39627859714641687,
39633321159821042, 39633324632703364, 39633324645287048, 39633328084619301,
39633331507171632, 39633331536531297, 39633334904557083, 39633334908748247,
39633334942302972, 39633334954887818, 39633341665771821, 39633341665775786,
39633341674160813, 39633345012829156, 39633345025413416, 39633348355689677,
39633351639826525, 39633351648217151]

tractorfile = '/pscratch/sd/i/ioannis/Y3-templates/sample/tractor-iron-vi-main.fits'
alltargets = select_targets(tractorfile, backup=False, numproc=1)
for targetid in targetids:
    print(targetid, alltargets[alltargets['TARGETID'] == targetid])
    
39627658597767428 []
39627658601958684 []
39627676696188831 []
39627835563838694 []
39627847626657859 []
39627847664405076 []
...
...
...
```

I also checked the `main` branch against this `branch` for a couple of sweeps files (one from the north and one from the south), e.g.:

```
from desitarget.cuts import select_targets
import fitsio
swfile = "/global/cfs/cdirs/cosmo/data/legacysurvey/dr9/north/sweep/9.0/sweep-270p035-280p040.fits"
alltargets = select_targets(swfile, backup=False, numproc=1)
fitsio.write("/pscratch/sd/a/adamyers/ADM-photsys-branch-targets.fits", alltargets)
```

All resulting targets and columns are identical. Some example output files are here:

```
/pscratch/sd/a/adamyers/main-targets.fits
/pscratch/sd/a/adamyers/ADM-photsys-branch-targets.fits
```

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -7,11 +7,13 @@ desitarget Change Log
 
 * When resolving, only set bits for northern cuts/imaging [`PR #823`_].
     * And, similarly, for southern cuts in southern imaging.
+    * Addresses `issue #821`_.
 * Function to match RA/Dec positions to Main Survey targets [`PR #820`_].
 * Bump astropy from 5.0 to 5.3.3 (dependabot) [`PR #815`_].
 
 .. _`PR #815`: https://github.com/desihub/desitarget/pull/815
 .. _`PR #820`: https://github.com/desihub/desitarget/pull/820
+.. _`issue #821`: https://github.com/desihub/desitarget/issues/821
 .. _`PR #823`: https://github.com/desihub/desitarget/pull/823
 
 2.7.0 (2023-12-05)
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,11 +5,14 @@ desitarget Change Log
 2.7.1 (unreleased)
 ------------------
 
+* When resolving, only set bits for northern cuts/imaging [`PR #823`_].
+    * And, similarly, for southern cuts in southern imaging.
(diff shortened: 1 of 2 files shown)


moustakas (approved):
Thanks for the fix, @geordie666. I tested the case I documented in #821, and the changes in this branch work as expected.

My only thought or comment, which I haven't tested, is what happens if `PHOTSYS` is neither `N` or `S`? In the most general case, can't some (secondary, backup, etc.) targets have `G` or even blank `PHOTSYS` values, and does the code already handle this (perhaps upstream)?

geordie666:
I think it's pretty clear in the documentation for `select_targets` that only sweeps or Tractor files are valid inputs. I think that circumvents the possibility of having non-valid values of `PHOTSYS`.

The code base does, indeed, handle Gaia-only and secondary target cases in other places.

PR #824, opened 2024-05-21
Add a function to help secondary proposers check the format of their target files

geordie666:
This PR adds a function to check the format of secondary target files for the new call for secondary targets.

As the PR only adds a small new function and doesn't change existing code, it's likely to be uncontroversial and I'll merge it later today.


geordie666:
Tests will fail in this branch but have been fixed in #814.

PR #826, opened 2024-07-18
Clean up environment variables defined in desimodules

weaverba137:
As of desihub/desimodules#52, some environment variables will be defined in desimodules.

Please wait until the PR mentioned above is merged *and deployed* before merging.


sbailey (approved):

sbailey:
FTR: I merged, updated changes.rst, reinstalled desitarget/main at NERSC, and verified that these environment variables are still set by "source desi_environment.sh main", but unlike before they are no longer unset by "module unload desitarget" (which is a good thing, because they are also needed by desispec).

Issue #811, opened 2024-01-03
geomask.match(): check for duplicates


araichoor:
the `desitarget.geomask.match()` function only works if there are no duplicates in the input arguments.
this is correctly written in the docstr; however, not everyone carefully reads a docstr :)

if I m correct, with the current version, when the arguments contain duplicates, the code just proceeds, but the outputs are not meaningful; so it s likely the user will be hurt by that.

that s why it would be better/safer to add a check that none of the arguments have duplicates; and just return an errror if they do.
something like:
```
if np.unique(A).size != len(A):
    msg = "A contains duplicates"
    log.error(msg)
    raise ValueError(msg)
if np.unique(B).size != len(B):
    msg = "B contains duplicates"
    log.error(msg)
    raise ValueError(msg)
```

also, with writing the code snippet above, I m wondering if A and B should only be 1d-array...

araichoor:
@geordie666, another suggestion for this one:

actually, maybe:
- add an argument to say if the check for duplicates should be done or not;
- set the default to True;
e.g.: `match(A, B, check_for_dups=True)`; and also for the "twin" function: `match_to(A, B, check_for_dups=True)`.

would that make sense?

the motivation for that is that this check can take few seconds for 1e8 arrays, and if one is really looking for speed -- and already knows that the inputs don t have duplicates, this could be a "loss" of time.

[context: I ve a in-development fiberassign update using `match()`, to speed up the code when running at once several hundreds of tiles with several tens of millions of targets]

geordie666:
Ah, yes, sorry, this fell off my plate for a while.

Would it be better to set the default to `check_for_dups=False` for strict backward-compatibility? It's likely people use this function in other places where speed is important.

araichoor:
one suggestion here, if we re looking for speed for this check -- and this may justify setting `check_for_dups=True`:

if `check_for_dups=True`, maybe the check can be done just after these lines:
https://github.com/desihub/desitarget/blob/971d9d5c6b8beea27d1722306f2d54ee5ff82389/py/desitarget/geomask.py#L91-L93
with something like (I ve not tested the code):
```
    # check for duplicates?
    if check_for_dups:
        n_Adups = np.count_nonzero(tmpA[1:] == tmpA[:-1])
        n_Bdups = np.count_nonzero(tmpB[1:] == tmpB[:-1])
        msg = []
        if n_Adups > 0:
            msg.append("A has {} duplicates".format(n_Adups))
        if n_Bdups > 0:
[...]

PR #828, opened 2024-08-20
Update matching routines to check input arrays for duplicates

geordie666:
This PR addresses issue #811.

@araichoor: In the end I used your suggested approach, utilizing `np.count_nonzero()`, in full. I tested a bunch of corner cases and I think your approach always returns the desired result. And it is _very_ fast!

I also agree with @araichoor that it's actually _desirable_ to break backwards compatibility in this case, as the results without checking for duplicates are essentially garbage.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -6,6 +6,7 @@ desitarget Change Log
 ------------------
 
 * Change match routines to test input arrays for duplicates [`PR #828`_].
+  * Addresses `issue #811`_.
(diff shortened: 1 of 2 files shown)


araichoor:
thanks!

moustakas:
For the record, this PR looks good to me (I use the matching functionality quite extensively). Quick request: can a new tag be created after this PR (or soon after, in case there's some other stuff in progress)?

PR #829, opened 2024-08-22
fix pkg_resources, scipy.ndimage, and pytest deprecation warnings

moustakas:
This PR cleans up a number of deprecation warnings seen in the unit tests due to:
* `pkg_resources.resource_filename` (deprecated in favor of `importlib.resources.files`; see also https://github.com/desihub/desispec/pull/2157))
* `scipy.ndimage`; and
* `pytest` (deprecated `test_suite` functions).

The `pkg_resources` warnings aren't seen in the `desitarget` tests because tests are only done against `python/3.9`, but they are seen in other packages like `fastspecfit`, which are testing against `python/3.10` and `3.11`.

### DIFF
--- a/bin/make_new_sv
+++ b/bin/make_new_sv
@@ -6,7 +6,7 @@ import numpy as np
 #import warnings
 #warnings.simplefilter('error')
 
-from pkg_resources import resource_filename
+from importlib import resources
 from glob import glob
 from time import time
 start = time()
@@ -19,7 +19,7 @@ ap = ArgumentParser("Make a new sv directory, based on the highest existing sv d
 _ = ap.parse_args()
 
 # ADM the standard root name for a desitarget SV directory.
-svroot = resource_filename('desitarget', 'sv')
+svroot = str(resources.files('desitarget').joinpath('sv'))
 
 # ADM retrieve the highest current sv directory.
 fns = sorted(glob(svroot+'*'))
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,6 +5,7 @@ desitarget Change Log
 2.7.1 (unreleased)
 ------------------
 
+* Fix `pkg_resources`, `scipy.ndimage`, and `pytest` deprecation warnings [`PR #829`_].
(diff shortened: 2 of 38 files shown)


moustakas:
@geordie666 I accidentally updated `doc/changes.rst` for this branch in `main` (while belatedly checking to make sure I had the most recent commits`), but also added it here. Let me know if you want me to fix `main` (sorry!).

Also, coverage tests are failing (not just here, but for some time), but I'm not sure why.

geordie666:
Excellent, thanks for these updates. Reviewing now. I don't think it's an issue that you update `changes.rst` in `main`. That doc file is always easy to merge/revert.

geordie666:
I think everything looks good @moustakas, and only have one suggestion/request:

Could you change constructions like this (there are quite a few so I didn't want to flag them all):

```cmd = resources.files('desitarget').joinpath('urat/fortran/v1dump')```

to use `os.path.join()` or something similar? I know that we _mostly_ only support nix-like operating systems in `desitarget` but it's conceivable that some functions are useful on other OS's that don't use `/` as the path separator.

I'm not sure why the coverage tests are failing. It hasn't been a high priority to dig into, but we should open an issue for completeness.

moustakas:
> Could you change constructions like this (there are quite a few so I didn't want to flag them all):
> cmd = resources.files('desitarget').joinpath('urat/fortran/v1dump')
> to use os.path.join() or something similar? I know that we mostly only support nix-like operating systems in desitarget but it's conceivable that some functions are useful on other OS's that don't use / as the path separator.

@geordie666 I'm sorry to push back, but I think this is unnecessary busy work for me and I don't think I'd prioritize getting to these changes (whereas the deprecation warnings are making it annoying for me to identify genuine errors and other relevant warnings in `fastspecfit`). 

In the spirit of making the minimal set of changes in this PR, I simply inherited the use of `/` as the path separator currently in `desitarget/main`, so if we want to make this change, I recommend we open a new ticket and try to make these changes in a different PR. 

Hope that's alright.

geordie666:
That's fine. Go ahead and merge. I'll do the busywork in my next branch.

PR #832, opened 2025-01-29
fix install with latest desiutil without DesiTest

sbailey:
desihub/desiutil#212 removed `desiutil.setup.DesiTest`, which broke the ability to install desitarget when using the latest desiutil.  This PR fixes that by removing the DesiTest import from setup.py.  Now to run desitarget tests you have to use `pytest` and not `python setup.py test`.  That has been our recommendation for awhile (and how github actions and NERSC tests have been performed), but this PR removes support for the old `python setup.py test` option.

With current main this fails with "ImportError: cannot import name 'DesiTest' from 'desiutil.setup'":
```
conda create --yes --name desitest 'numpy<2' 'astropy<7'
conda activate desitest

pip install git+https://github.com/desihub/desiutil.git@main
pip install git+https://github.com/desihub/desitarget.git@main
```

Replacing that last line with this branch works
```
pip install git+https://github.com/desihub/desitarget.git@fix-install
```

After merging this branch, it should be possible to install desitarget/main when using desiutil/main.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -20,7 +20,7 @@ jobs:
                 python-version: ['3.10']
                 astropy-version: ['==5.0', '<6.1']
                 fitsio-version: ['==1.2.1']
-                numpy-version: ['<2']
+                numpy-version: ['<1.23']  # to keep asscalar, used by astropy
                 scipy-version: ['<1.9']
                 matplotlib-version: ['<3.7']
 
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -17,7 +17,7 @@ jobs:
             fail-fast: true
             matrix:
                 os: [ubuntu-22.04]
-                python-version: [3.10]
+                python-version: ['3.10']
                 astropy-version: ['==5.0', '<6.1']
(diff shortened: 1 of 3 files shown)


sbailey:
Ugh, tests are failing while installing cfitsio.  Will need some more work to restore that.

weaverba137 (approved):
This looks fine to me. It looks like you may need to sort out dependencies for compiling `fitsio`. The first thing to try is using `ubuntu-22.04` instead of `ubuntu-latest`.

sbailey:
Well that was unsatisfying.  matplotlib is trying to force an upgrade to numpy>2, which isn't compatible with the other packages.  Using --no-deps for matplotlib means it doesn't have other dependencies like PIL, which I haven't figured out how to install yet given the other package constraints...

weaverba137:
I went through something very similar last week. I'll point you to it in the morning.

weaverba137:
You might want to compare with desihub/desimodel#178.

sbailey:
Finally, success.  In the end the coverage and regular unit tests have a somewhat different matrix of dependency versions because that is just how it landed, but I'll declare that to be a feature for providing some examples of combinations of versions that work.

PR #834, opened 2025-02-28
Update RTD configuration

weaverba137:
An update to RTD configuration is needed to keep the RTD Search function working. See desihub/desispec#2445 for details.


weaverba137:
Self-merging at @sbailey's request.

PR #835, opened 2025-03-02
Mwextension

crockosi:
Changes to do target selection for streams and dSph galaxies for the DESI Extension.
Added to targetmask.yaml in the mws_mask, priorities and numobs.
Added more stream-specific info to the GD1 entry in streams.yaml, goal is for all stream params to be in streams.yaml and not hard-coded in the is_in functions.
Modified streams/cuts.py to implement new magnitude ranges in the target selection in is_in_GD1, and set_target_bits to set the new mws_mask bits.
Modified streams/target.py to set PRIORITY and NUMOBS in the mws_mask and not as secondary targets.


geordie666:
@crockosi: My latest branch allows caching in HEALPixels, which is what I think we'll ultimately want — i.e. we're going to want to process all of the MW extension targets together in a portion of the sky rather than processing individual streams/dwarfs one-by-one over "the full sky".

You can see how this new functionality works by running something like this at the command line:

(after grabbing an interactive node, e.g.: `salloc --nodes 1 --qos interactive --time 02:00:00 --constraint cpu`)

`select_stream_targets /global/cfs/cdirs/cosmo/data/legacysurvey/dr9/south/sweep/9.0 $SCRATCH/blat -t GD1,ORPHAN --numproc 16 --nside 1 --pixint 5 --donotreadcache`

this will create the cache and run the `GD1` and `ORPHAN` streams together in HEALPixel 5 at `nside` of 1. The cache will be placed in `$TARG_DIR/streamcache/dr9` (e.g. `/global/cfs/cdirs/desi/target/catalogs/streamcache/dr9`). Note the extra "dr9" directory in the cache names, now.

Once you've checked the caching works for you, you can run without the `donotreadcache` option:

`select_stream_targets /global/cfs/cdirs/cosmo/data/legacysurvey/dr9/south/sweep/9.0 $SCRATCH/blat -t GD1,ORPHAN --numproc 16 --nside 1 --pixint 5`

to read the cache without generating it.

Some additional notes:

- Eventually, to produce the full suite of MWS extension targets, we'd run the above commands over all possible HEALPixels (so, once for `--nside 1 --pixint 0`, once for `--nside 1 --pixint 1`, once for `--nside 1 --pixint 2`, etc.)
- I've currently commented out:
```
#    with open(fn2) as f:                                                                                                                                                                                                                                       
#        dwarf = yaml.safe_load(f)                                                                                                                                                                                                                              
#    dwarflist = list(dwarf.keys()) 
```
in `get_targthing_parameters()` in `utilities.py()`. Once you've `git add`ed `data/dwarfs.yaml` you can uncomment those lines.

- I think we probably want to move away from including the stream/dwarf names in the output filename when we run over everything at once (when we run with `readall=True`) as the output filenames are going to get long! That should be an easy fix but I only noticed after I pushed my updates.

- As you and I discussed offline, we should change `thing` to `mwext` throughout the code. Or just drop `thing` completely (i.e.  `get_targthing_parameters()` could just be `get_targ_parameters()`.

Let me know if this new caching implementation works for you (either here or over Slack).

geordie666:
My latest push addresses the third bullet point from above, i.e.:

- I think we probably want to move away from including the stream/dwarf names in the output filename when we run over everything at once (when we run with `readall=True`) as the output filenames are going to get long! That should be an easy fix but I only noticed after I pushed my updates.

geordie666:
The general code development looks good, here. A couple of items so we don't forget:

- `streams.io.write_targets()` currently writes a file with no extensions when there are zero targets to write. We should fix that to instead exit quickly with 0 targets rather than writing a file. We could return the `NoneType` for the filename in that instance (and mention that choice in the docstring).
- `streams.utilities` now uses a value `TARGMWEXT_RANK` to decide which stream a duplicate target is in. 
    * @crockosi: Should we be writing out this rank to the target files for provenance? It wouldn't show up anywhere in the DESI data model except in the target files. But, it seems like it might be useful information for modeling the streams that we shouldn't discard?

crockosi:
Your second point is very wise, Adam. We should definitely do that.

We should take care of the zero targets thing, too.  I'll try to do that
before I push the bug fixes I have ready to go.

--Connie

geordie666:
I'm merging this to work on the `BRIGHT1B` MTL updates in a separate branch as some of the required 1B MTL infrastructure is in `main`.

PR #845, opened 2025-07-15
Change LyA decisions for the DESI 1B phase

geordie666:
This PR updates the LyA decisions for the DESI 1B phase. It includes:

- A function to add 2 observations for every LyA quasar in the MTL ledgers.
  - This function also sets the priority for these quasars back to the most recent LyA-like one.
  - This is necessary, for example, because LyA quasars that are "done" have a very low priority and would never be revisited without a priority update.
- New MTL logic that sets 6 observations for every ELG target that is observed to have a redshift consistent with a LyA QSO.
  - Such quasars are set to a new priority level, named `LY_1B_ELG`.
  - This new priority level is just below a QSO target that turns out to be a LyA quasar.

I've mocked up ledgers for both of these new sets of LyA decisions:

`/pscratch/sd/a/adamyers/1bqsocheck`: Contains the results of updating a couple of ledgers by adding +2 observations to every existing LyA-like quasar.

`/pscratch/sd/a/adamyers/1belgcheck`: Contains the results of running the new ELG-target-is-a-LyA-QSO logic on 4 tiles with tile IDs `102276`, `102277`, `102281` and `102282`.

@ashleyjross and @LNapolitano: It would be great if you could run some tests on the mocked up ledgers to see if the results look good and make sense.

### DIFF
--- a/bin/update_lya_1b
+++ b/bin/update_lya_1b
@@ -1,0 +1,26 @@
+#!/usr/bin/env python
+
+from desitarget.mtl import update_lya_1b
+
+from argparse import ArgumentParser
+ap = ArgumentParser(description="Add two observations for LyA quasars for 1B.")
+ap.add_argument("--obscon",
+                help="String matching ONE obscondition in the bitmask yaml file \
+                (e.g. 'DARK'). Defaults to DARK.", default="DARK")
+ap.add_argument('--mtldir',
+                help="Full path to the directory that hosts the MTL ledgers.    \
+                Default is to use the $MTL_DIR environment variable.",
+                default=None)
+ap.add_argument('--timestamp',
+                help="Fixed time stamp at which to record updates.              \
+                Default is to use NOW.",
+                default=None)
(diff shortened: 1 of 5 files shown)


LNapolitano:
Sure thing, looks like all of the previously offending entries do have 'Z_QN > 2.1` and `IS_QSO_QN == 1`

image

LNapolitano:
I've been trying to recreate the list of 192 entries from `mtl-dark-hp-3072.ecsv` that are given a new entry in the 1B ledgers. There are 3 such entries with a final pre-1B status of  `TARGET_STATE == QSO|MORE_MIDZQSO`, and it's not immediately obvious to me why these should get additional observations. Given that they've all had only a single observation I don't believe they could have been considered LyA-QSO in a previous ledger entry. Do these make sense to you @geordie666?

image

LNapolitano:
Maybe a false alarm, looks like these all were considered LyA QSO at some previous state:

image

geordie666:
Right, thanks, I think the logic here is that these were all "once" Lyman-alpha quasars.

PR #863, opened 2025-10-28
Add capability to veto M31 (and potentially other) targets directly in svn

geordie666:
This PR adds the capability for a file to be added to, and tracked, in svn to veto M31 targets. It could easily be expanded to other programs if needed.

The general approach is that files can be added to a new `mtl/main/veto` directory that contain `TARGETID`s that should be "turned off" in the MTLs. An example such file (provided by Arjun) is at:

```
/pscratch/sd/a/adamyers/mtlsandbox/trunk/mtl/main/veto/bright1b/m31/m31-veto-bright1b.ecsv
```

New veto programs could be added by including new directories of the form, e.g., `program2/program2-veto-bright1b.ecsv`

The code reads the veto file (or files), checks for updates since the last MTL run, and changes any targets in the MTLs that are newly in the veto file to have `NUMOBS_MORE=0`, `PRIORITY=2` and`TARGET_STATE=SVN_VETO|DONE` at a given `TIMESTAMP`.

A new MTL done file called `mtl-done-svn-vetoes.ecsv` is then added to track the `TIMESTAMP` of the most recent update. An example such file is at:

```
/pscratch/sd/a/adamyers/mtlsandbox/trunk/mtl/mtl-done-svn-vetoes.ecsv
```

An example of how running the veto files updates the ledgers can be seen via, e.g.:

```
tail /pscratch/sd/a/adamyers/mtlsandbox/trunk/mtl/main/bright1b/mtl-bright1b-hp-5111.ecsv
```

The code does not make any updates for programs for which a veto directory does not exist. For example, because the directory `/pscratch/sd/a/adamyers/mtlsandbox/trunk/mtl/main/veto/dark` does not exist, the vetoing process is completely ignored for the `dark` program. This, coupled with `TIMESTAMP`s for any updates, should ensure backward compatibility.

As written, the code should seamlessly run with the daily operations `run_mtl_loop` updates once a veto file is added to svn.

This is a work-in-progress for now while I write meaningful unit tests. But, I'm opening the PR in case @sbailey finds time to take a look in the meantime.

### DIFF
--- a/py/desitarget/io.py
+++ b/py/desitarget/io.py
@@ -3674,9 +3674,6 @@ def read_targets_in_hp(hpdirname, nside, pixlist, columns=None, header=False,
           grouped HEALPixels, as fewer files will need to be read.
         - If `mtl` is ``True`` then this is just a wrapper on
           read_mtl_in_hp().
-        - Will generally fail if the target files in a given `hpdirname`
-          are formatted differently BUT tries to allow a special DESI
-          case of an M31-program-like file and a standard DESI file.
     """
     # ADM if quick is True, use the quick-code.
     if quick:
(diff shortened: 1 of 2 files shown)


py/desitarget/io.py
```diff
@@ -2727,6 +2727,73 @@ def find_target_files(targdir, dr='X', flavor="targets", survey="main",
...
+        A structured numpy array of the MTL veto file.
+
+    Notes
+    -----
+        - The checks that are performed include:
+            - The file should include these columns (with correct type):
+                `RA`, `DEC`, `TARGETID`, `TIMESTAMP`
+            - No TARGETID should be duplicated.
+            - No TIMESTAMP lower down in the file should be earlier in
+              time that a TIMESTAMP higher up in the file.
```
sbailey: This function allows optional columns to exist as long as the required columns are there.  I think that is good, e.g. to support end-users including a "REASONS" column, or other parameters that were measured that led to the reason for the veto.  At the same time, this function returns the full original table including those extra columns, which might have datamodel implications downstream depending upon how this table is merged with others.

Consider allowing the existence of optional columns in the input file, but stripping the data to only the columns you expect in only the order you want them so that all downstream usage gets a consistent data model.  e.g. with
```python
return vetodata['RA', 'DEC', 'TARGETID', 'TIMESTAMP']
```

sbailey (approved):
Overall, looks good. I put two comments inline for consideration. I wasn't able to view the example files in your $SCRATCH directory (unix permissions...) but the description makes sense.

I think the structure of your vetos directories means that an end user could svn checkout just the surveyops/trunk/mtl/main/veto directory without having to checkout the giant surveyops/trunk/mtl/main directory will all the MTL files.  That's good, safer, and easier for end users, but I'd appreciate a double check that you agree that veto providers only need a small svn checkout, not the very large one.

Quibble: the fact that the veto files are kept in svn is a technical detail that doesn't actually matter for the code, but "svn" appears multiple times in various variable and filenames. Does including "svn" in the name actually carry value, or is it just extra info that could get out of date someday if we outgrown svn?  v.s. eg. "mtl-done-vetos.ecsv" instead of "mtl-done-svn-vetos.ecsv"?

geordie666:
I think the new veto process has reasonable unit-test coverage, now, so I'll merge this tomorrow morning and make a new tag.

Issue #866, opened 2025-11-10
cuts.py masked elements as NaN


sbailey:
As reported by Ben in PR #864:
> I'm seeing these warnings on Python 3.12 and 3.13:
```
...
py/desitarget/test/test_cmx.py::TestCMX::test_astropy_table
py/desitarget/test/test_cuts.py::TestCuts::test_astropy_table
  /home/runner/work/desitarget/desitarget/py/desitarget/cuts.py:2195: UserWarning: Warning: converting a masked element to nan.
    refcat = np.array([refcat, ])
```

This might be harmless due to interplay between astropy Table with masked elements vs. numpy ndarrays.  At the same time, it would be useful to work around this so that we don't get in the habit of ignoring warnings, and it might reveal a surprise gotcha about why the elements were masked in the first place (earlier versions of astropy certainly had some surprise behaviors there).

sbailey:
Fixed by PR #870.

Issue #865, opened 2025-11-10
bool vs. int usage and bitwise inversion of photsys_north


sbailey:
As reported by Ben in PR #864:
> I'm seeing these warnings on Python 3.12 and 3.13:
```
...
py/desitarget/test/test_cuts.py::TestCuts::test_numpy_ndarray
  /home/runner/work/desitarget/desitarget/py/desitarget/cuts.py:2088: DeprecationWarning: Bitwise inversion '~' on bool is deprecated and will be removed in Python 3.16. This returns the bitwise inversion of the underlying int object and is usually not what you expect from negating a bool. Use the 'not' operator for boolean negation or ~int(x) if you really want the bitwise inversion of the underlying int.
    photsys_south = ~photsys_north
```

It looks like `photsys_north` was intended to be a bool, but something is turning it into an int.  The issues is that e.g. if `photsys_north==1`, then `photsys_south = ~photsys_north` ends up as -2 which is evaluated as True instead of False.

Belt-and-suspenders I suggest following the warning recommendation of using
```
photsys_south = not photsys_north
```
but also tracking down how/why `photsys_north` is an integer in the first place and forcing that to a bool, since as an int it might have other unintended consequences.

dylanagreen:
I was reading about this deprecation this weekend, as the bitwise ~ on a bool object is removed in the forthcoming Python 3.16, and just wanted to clear something up: the error message is poorly worded. `photosys_north` is a `bool` object when this warning occurs, not an integer. In python, however, `bools` are represented internally as integer objects, which is why the error refers to the `bitwise inversion of the underlying integer object`. This doesn't mean that `photosys_north` is an integer itself. 

We've probably been surreptitiously getting away with this for a long time, because if `photosys_north` is False, then `~photosys_north` will correctly evaluate to True. In a Python 3.10 ipython:
```
In [1]: blah = True

In [2]: type(blah)
Out[2]: bool

In [3]: ~blah
Out[3]: -2

In [4]: bool(~blah)
Out[4]: True

In [5]: blah = False

In [6]: ~blah
Out[6]: -1

In [7]: bool(~blah)
Out[7]: True
```

As an addendum, the representation of `bool` as an `int` internally in python is also why Python `bool` objects take up a the same space in memory in bytes, rather than the expected lower memory usage:
```
In [12]: x = 2

In [13]: sys.getsizeof(x)
Out[13]: 28

In [14]: y = True

In [15]: sys.getsizeof(y)
Out[15]: 28
[...]

PR #870, opened 2025-12-18
Circumvent warnings when running with Python >3.12

geordie666:
This PR addresses issues #865 and #866.

In both cases, the warnings that were being raised by the unit tests for Python 3.12 and 3.13 were benign.

- For issue #865 we were already converting single Booleans using `not` and Boolean arrays using `~`. I just had to restructure this conversion to place the `~` case in an `if/else` block.

- For issue #866 the conversion of a single-element masked array to `NaN` was expected. I just had to make this conversion more explicit to prevent a warning being triggered.

### DIFF
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -2085,10 +2085,11 @@ def _prepare_optical_wise(objects, mask=True):
     # ADM flag whether we're using northen (BASS/MZLS) or
     # ADM southern (DECaLS) photometry.
     photsys_north = _isonnorthphotsys(objects["PHOTSYS"])
-    photsys_south = ~photsys_north
-    # ADM catch case where single object or row is passed.
+    # ADM make sure to catch the case of a single object or row.
     if isinstance(photsys_north, bool):
         photsys_south = not(photsys_north)
+    else:
+        photsys_south = ~photsys_north
 
     # ADM the observed r-band flux (used for F standards and MWS, below).
     # ADM make copies of values that we may reassign due to NaNs.
--- a/py/desitarget/cuts.py
+++ b/py/desitarget/cuts.py
@@ -2199,7 +2199,8 @@ def _prepare_gaia(objects, colnames=None):
         # ADM the same outcome more explicitly without a warning.
         if np.ma.is_masked(refcat):
             refcat = np.array([np.nan])
-        refcat = np.array([refcat, ])
+        else:
+            refcat = np.array([refcat, ])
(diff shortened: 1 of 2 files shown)


py/desitarget/cuts.py
```diff
@@ -2192,6 +2193,12 @@ def _prepare_gaia(objects, colnames=None):
     gaia = objects['REF_ID'] > 0
     refcat = objects['REF_CAT']
     if _is_row(objects):
+        # ADM A single-element masked array is converted to array([nan])
+        # ADM by np.array([]), which is fine for our purposes, but as of
+        # ADM Python 3.12 this triggers a warning. This workaround gets
+        # ADM the same outcome more explicitly without a warning.
+        if np.ma.is_masked(refcat):
+            refcat = np.array([np.nan])
```
sbailey: Is there any danger here that `refcat` is a multi-element array where only some elements are masked? In that case `np.ma.is_masked(refcat)` is True and `refcat` would get replaced with a single element array of nan.

I'm confused about what is actually the triggering situation here.  Using python/3.12.11 and numpy/2.2.6 I haven't been able to generate the warning while converting single-element masked arrays:
```
In [1]: x = np.ma.array([1,], mask=[False,])

In [2]: np.array(x)
Out[2]: array([1])

In [3]: x = np.ma.array([1,], mask=[True,])

In [4]: np.array(x)
Out[4]: array([1])

In [5]: x = np.ma.array([np.nan,], mask=[True,])

In [6]: np.array(x)
Out[6]: array([nan])

In [7]: x = np.ma.array([np.nan,], mask=[False,])

In [8]: np.array(x)
Out[8]: array([nan])
```
Can you reproduce the behavior that you are fixing here, outside of the context of the unit tests?

geordie666:
@sbailey @dylanagreen: As we were discussing related deprecation problems, I revisited this PR and found the bug that impeded my initial fixes for issues https://github.com/desihub/desitarget/issues/865 and https://github.com/desihub/desitarget/issues/866.

Everything should be good now. If somebody wants to further review my work, great. Otherwise I'll merge this early next week.

dylanagreen:
@geordie666 I looked over your fix and the code and it looks good to me. I'll ping @sbailey to clear it to merge before you leave for vacation.

geordie666:
Thanks @dylanagreen. I think it's fine. Very little was actually impacted by the deprecation warnings, except for the case of passing a single target in an astropy Table through the code.

So, I'm going to go ahead and merge. I don't think Stephen needs to take a look.

PR #872, opened 2026-01-29
LSST ToO update

jbanovetz:
This update will divide the ToO-input.ecsv list into a primary folder that grows over time (anything with `-all` in the name) and a smaller folder of just targets that are available right now. Fiber assign on the fly will use the smaller folders for assigning ToOs, allowing for more ToOs to be uploaded in a day.

### DIFF
--- a/py/desitarget/ToO.py
+++ b/py/desitarget/ToO.py
@@ -475,7 +475,7 @@ def finalize_too(inledger, survey="main"):
     return outdata
 
 
-def ledger_to_targets(toodir=None, survey="main", ecsv=True, outdir=None, subtable=False):
+def ledger_to_targets(toodir=None, survey="main", ecsv=True, outdir=None):
     """Convert a ToO ledger to a file of ToO targets.
 
     Parameters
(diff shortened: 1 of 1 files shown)


py/desitarget/ToO.py
```diff
@@ -518,15 +524,32 @@ def ledger_to_targets(toodir=None, survey="main", ecsv=True, outdir=None):
...
-        fn = get_filename(tdir, outname=True, ender=form)
+        fn = get_filename(tdir, ender=form, outname=True)
     else:
-        fn = get_filename(outdir, outname=True, ender=form)
+        fn = get_filename(outdir, ender=form, outname=True)
 
     # ADM write out the results.
-    _write_too_files(fn, outdata, ecsv=ecsv)
+    # JB added if statement for writing out a subtable
+    if subtable:
```
geordie666: I'm OK with your approach if you find it easiest to parse, but consider whether you need two `if` statements; one for setting `subtable` and one for using it. Particularly as the `subtable` kwarg seems unnecessary.

Every relevant line of code for subsampling the data could presumably be embedded in a single `if todaymjd >= Time('2026-01-28').mjd:` block where you currently have the `if subtable:` block.

geordie666:
Thanks @jbanovetz. This looks good I'm going to merge and tag now.

PR #874, opened 2026-02-12
LSST ToO update

jbanovetz:
Trying to re-implement this after the last time. The major change here is that instead of calling `_write_too_files` twice as before, resulting in different `TIMESUBMIT`, the subtable just makes the cuts on the larger table. I also changed it so that the two `*all*` files get copied before starting, as that also resulted in different values for `TIMESUBMIT`. I have tested this locally and it does work. Waiting on some tests from Dylan before merging and possibly changing the start time for doing this (currently 2026/01/31)/


py/desitarget/ToO.py
```diff
@@ -531,17 +547,20 @@ def ledger_to_targets(toodir=None, survey="main", ecsv=True, outdir=None):
...
+        for file in ['ToO-all.ecsv','ToO-fiber-all.ecsv']:
+            if os.path.exists(f'{tdir}/{file}'):
+                continue
+            else:
+                os.system(f"cp {tdir}/{file.replace('-all','')} {tdir}/{file}")
+
         all_fn=f'{tdir}/ToO-all.{form}'
         log.info('Writing all ToOs to {}'.format(all_fn))
-        _write_too_files(all_fn, outdata, ecsv=ecsv)
+        _write_too_files(all_fn, outdata, ecsv=ecsv,subtable=True,date=date)
```
geordie666: Add spaces after commas for PEP 8 code style.

geordie666 (approved):
I mostly reviewed this offline before you issued the PR, and have tested it myself. Everything looks good except for minor code style.

geordie666:
Some additional notes:

- We will definitely need to change the start date to a future date rather than leaving it as a past date.
  * So don't merge this yet!
- We'll need to remember to initially `svn add` the new `-all-` files the first time the new code is run.
- I've been thinking about how to deal with `validate_lastnight_fba` if somebody runs it for a _past_ night, i.e. for a night _before_ we deploy these changes. I think the best I've come up with is to add an instruction to `svn up -r` to the appropriate version of the `$TOO_DIR` before running `validate_lastnight_fba` in such cases.

geordie666:
@jbanovetz: While we're waiting to merge, could you please update the `changes.rst` file with a description of this PR.

jbanovetz:
Resurrecting this again, I think that once NERSC is back up, I just need to update somethings and make sure everything works with the current `main` branch (what's causing the conflicts in the check above). Is there anything else that needs to be done @geordie666?

geordie666:
Other than resolving the conflicts, I don't think anything else needs to be done, here. Let's wait to merge until we get through the `validate_lastnight_fba` from NERSC downtime, though.

Then, I think we can merge during the Mayall shutdown and, conveniently, [change the start date](https://github.com/desihub/desitarget/pull/874#issuecomment-3893914555) to any date during the telescope shutdown. Say September 1st.

jbanovetz:
I made the changes though am having trouble with all of the CI checks and am getting this one a lot which I don't think is related to my PR:

```
RuntimeError: module was compiled against NumPy C-API version 0x12 (NumPy 2.0) but the running NumPy has C-API version 0x11. Check the section C-API incompatibility at the Troubleshooting ImportError section at https://numpy.org/devdocs/user/troubleshooting-importerror.html#c-api-incompatibility for indications on how to solve this problem.
=========================== short test summary info ============================
ERROR py/desitarget/test/test_brick_fluctuation.py
ERROR py/desitarget/test/test_brightmask.py
ERROR py/desitarget/test/test_cmx.py
ERROR py/desitarget/test/test_cuts.py
ERROR py/desitarget/test/test_geomask.py
ERROR py/desitarget/test/test_io.py
ERROR py/desitarget/test/test_mock_build.py
ERROR py/desitarget/test/test_mtl.py
ERROR py/desitarget/test/test_mtl_multiple.py
ERROR py/desitarget/test/test_priorities.py
ERROR py/desitarget/test/test_qa.py
ERROR py/desitarget/test/test_skyfibers.py
ERROR py/desitarget/test/test_subpriority.py
ERROR py/desitarget/test/test_sv.py
ERROR py/desitarget/test/test_targetids.py
ERROR py/desitarget/test/test_units.py
ERROR py/desitarget/test/test_veto.py
!!!!!!!!!!!!!!!!!!! Interrupted: 17 errors during collection !!!!!!!!!!!!!!!!!!!
```

Otherwise I have fixed the conflicts and the only thing that might need to happen in the future is changing the start date.

geordie666:
@jbanovetz: I have fixed the relevant/important unit tests and we could merge this PR when you're ready.

I believe we were waiting on [some tests from Dylan](https://github.com/desihub/desitarget/pull/874#issue-3934435285), so please ensure that those have been done before merging.

jbanovetz:
If I remember right, Dylan might have done the tests? @dylanagreen do you remember the status of those tests and if they were resolved?

dylanagreen:
I do not remember what the tests were. Do you have any recollection @jbanovetz? I have some notes in the discussion between the three of us on Slack on February 12 where I ran a test comparing the old style vs the new style ToO files and found that they were identical:
> I ran a test tile creattion + assignment for a BRIGHT tile, centered at (141.430427,-3.458271) using an arbitrary tileid (I used 20001 in this case), using the oldstyle/ToO-fiber.ecsv and newstyle/ToO-fiber.ecsv as custom too-tiles. (This is a detail without distinction, but I attempted to export TOO_DIR= and it turns out that fiberassign hard codes the location of the ToO files based on DESI_SURVEYOPS :neutral_face: , so I instead had to use the privided --custom-too-file input to fiberassign. Since it disallows TILE toos it only searches for ToO-fiber anyway). Based on this test, using the two different ToO-fiber.ecsv files, the fiberassign outputs are identical in all aspects.

So if that was the test, then yes, it was done.

Unfortunately the location where I put the results on my scratch has been purged by NERSC but I could rerun them I think if we need. I am surprised I did not put it in my CFS.

jbanovetz:
I think the tests were that everything was working on the `TIMESUBMIT` side and that it wouldn't be too much of a change to point to a new `ToO-fiber` file? I forget if there was something with `fiberassign on the fly` or not.

dylanagreen:
I've rerun some tests with this branch for fiber assignment. I did two tests, as follows:

## New Tile
I fiber assigned a new tile with a RA/DEC I picked to pick up some ToOs that were active today. I did this tile three times: once with  `ToO-fiber.ecsv` from main, once with `ToO-fiber-all.ecsv` from this branch, and once with `ToO-fiber.ecsv` from this branch. Since I am testing with the tile "obsdate" of today, I expected all three of these to produce identical assignments, and they did! 

Here is a sample command I ran for this test, which uses `fba_launch` to assemble input files:
```
fba_launch --doclean n --dr dr9 --dtver 1.1.1 --forcetileid y --gaiadr gaiadr2 --goaltime 180 --mintfrac 0.85 --outdir $FAOUT/too_all/ --pmcorr n --program BRIGHT --rundate 2026-08-21T17:00:49+00:00 --sbprof ELG --sky_per_petal 40 --standards_per_petal 10 --steps tiles,sky,gfa,targ,scnd,too,fa,zip,move --survey main --tileid 20002 --tiledec 12.8 --tilera 216  --goaltype BRIGHT --sky_per_slitblock 1 --margin_pos 0.05 --margin_petal 0.4 --margin_gfa 0.4  --mtltime 2026-08-21T20:00:49+00:00 --log_stdout --try_too_all
```

I reused an old TILEID, so --force_tileid was necessary. `--try_too_all` is the switch to use `ToO-fiber-all.ecsv` rather than `ToO-fiber.ecsv`

## Rerun
Using `fba_rerun` I reran an old tile from 2025 (TILELID=21124), which picked up 3 ToOs in its original assignment. e.g.
```
fba_rerun --infiberassign $DESI_TARGET/fiberassign/tiles/trunk/021/fiberassign-021124.fits.gz --o`ToO-fiber.ecsv`utdir $FAOUT/too_all --nosteps qa [--force_try_too_all]
```

where `--force_try_too_all` forces the rerun to use the `-all.ecsv` if found. This switch is active in `validate_lastnight_fba` so this replicates that process. As expected, when using the truncated `ToO-fiber.ecsv`, this fails reproducibility (but interestingly only fails in the sense that it misses "potential" assignments for the ToOs, on this tile serendipitously none of the ToOs were _actually assigned_). It passes if using `--force_try_too_all`, using `ToO-fiber-all.ecsv`. 

For anyone wanting to dig into the results they are in my cfs: `$CFS/desi/users/dylang/too_test` where the `diff*` files are the difference in output between the original tiels (main or the one in $DESI_TARGET) with either `branch` (uses `ToO-fiber.ecsv`) or `too_all` (uses `ToO-fiber-all.ecsv`).

Since this is all working as expected, `ToO-fiber-all.ecsv` successfully reproduces old assignments, and `validate_lastnight_fba` uses `fba_rerun` using this file I'd say this is good to go.

geordie666:
I've merged this PR, but should remember [this additional step](https://github.com/desihub/desitarget/pull/874#issuecomment-3893914555) before deployment:

- We'll need to remember to initially `svn add` the new `-all-` files the first time the new code is run.

Issue #503, opened 2019-05-27
Deprecate the sandbox


geordie666:
The sandbox is not under active development, and I'm not sure that anyone is actually using it. I'd therefore like to remove it. It would, of course, remain in earlier tags of the desitarget code.

I'll leave this issue open for about a month before I remove any code related to the sandbox. Please comment on this issue if you *are* using the sandbox.

moustakas:
@geordie666 I agree that it was a good idea when we started out, but (largely because of your outstanding management of both baseline and experimental code!) I agree that the sandbox is basically obsolete (AFAIK).

geordie666:
Addressed in #564.

PR #877, opened 2026-04-05
Add secondary targets for DARK1B and BRIGHT1B tiles, BACKUP tiles, and coming DR11-based BRIGHT and DARK tiles.

geordie666:
This PR adds secondary targets for `1B` tiles, for `BACKUP` tiles and for the new `BRIGHT` and `DARK` area that will be based on DR11 imaging.

This is a work in progress until @dylanagreen has checked the interface with `fiberassign`, as detailed in [issue #503](https://github.com/desihub/fiberassign/issues/503). But, it will be necessary for related work to use this branch of `desitarget` to ensure that all of the new targeting bits can be referenced.


geordie666:
I think I've added everything that's needed to include `1B` secondaries, at least on `1B` tiles. I'll merge this and make a tag in the next day or two.

Issue #883, opened 2026-05-27
`read_mtl_in_hp` concatenation of MTLs breaks with new versions of numpy


dylanagreen:
New versions of numpy break on concatenation when the order of the two MTL columns don't match exactly:
```  File "/global/common/software/desi/perlmutter/desiconda/20260227-2.3.1/code/fiberassign/main/py/fiberassign/fba_launch_io.py", line 1491, in create_mtl
    d = read_targets_in_tiles(
        mtldir,
    ...<6 lines>...
        maketwostyle=True, # AR add MTL_{HIGHEST,WANTED,CONTAINS} columns
    )
  File "/global/common/software/desi/perlmutter/desiconda/20260227-2.3.1/code/desitarget/main/py/desitarget/io.py", line 4183, in read_targets_in_tiles
    targets = read_targets_in_hp(
        hpdirname, nside, pixlist, columns=columnscopy, header=header,
        mtl=mtl, unique=unique, isodate=isodate, initial=initial, leq=leq,
        tabform=tabform, maketwostyle=maketwostyle)
  File "/global/common/software/desi/perlmutter/desiconda/20260227-2.3.1/code/desitarget/main/py/desitarget/io.py", line 3699, in read_targets_in_hp
    return read_mtl_in_hp(
        hpdirname, nside, pixlist, unique=unique, isodate=isodate,
        initial=initial, leq=leq, tabform=tabform, maketwostyle=maketwostyle)
  File "/global/common/software/desi/perlmutter/desiconda/20260227-2.3.1/code/desitarget/main/py/desitarget/io.py", line 3583, in read_mtl_in_hp
    mtl = np.concatenate(mtls)
  File "/global/common/software/desi/perlmutter/desiconda/20260227-2.3.1/conda/lib/python3.13/site-packages/numpy/_core/_internal.py", line 472, in _promote_fields
    raise DTypePromotionError(
            f"field names `{dt1.names}` and `{dt2.names}` mismatch.")
[...]

Issue #855, opened 2025-09-18
read_two_mtl_ledgers incorrectly changes some column values for targets appearing in both ledgers


dylanagreen:
## Overview of the Problem
There is an inconsistency between when you load two MTL ledgers using `read_two_mtl_ledgers` vs using `read_one_mtl_ledger` to load the two ledgers individually, specifically for targets that are in both ledgers. This inconsistency leads to incorrectly set columns. 

Curiously this behavior seems to be entirely dependent on the _order_ in which the two ledgers are passed into the former function. For this we will use the two healpix ledgers and their associated two ledger loads specified by:

```python
a_loc = ${DESI_ROOT}/survey/ops/surveyops/trunk/mtl/main/bright/mtl-bright-hp-7921.ecsv
b_loc = ${DESI_ROOT}/survey/ops/surveyops/trunk/mtl/main/bright1b/mtl-bright1b-hp-7921.ecsv

a_ledger = read_one_mtl_ledger(a_loc)
b_ledger = read_one_mtl_ledger(b_loc)

a_then_b = read_two_mtl_ledgers([a_loc, b_loc])
b_then_a = read_two_mtl_ledgers([b_loc, a_loc])
```

These two MTLs share 265 targets between the two of them, with ~18000 targets in the A ledger and ~12000 in the B ledger, for a total of ~30000 in the double loaded and merged ledger:

Of those 265 shared targets, only 4 of them have a higher PRIORITY_INIT in the A ledger, while the other 261 have a higher priority (of any kind) in the B ledger. Only one target has a remaining higher priority in A compared to B, the other three have a priority equal to "done" in the A ledger, meaning that overall 264 of the targets have a higher priority in B.

### Higher Priority in A
Taking targetid 39628059149600576 as a representative example of what happens during the merging process between A and B when A is higher priority, we can print the following rows of each of the four ledgers:

```
[...]

PR #884, opened 2026-05-28
Fix incorrect MTL concatenation in `read_mtl_in_hp` that would change entire ledger column values.

dylanagreen:
This fixes the crash identified in fiberassign triggered by `read_mtl_in_hp`, in issue #883. 

The fix is done, at base level, by using `np.recfunctions.stack_array` instead of `np.concatenate` to concatenate the MTLs. The former function correctly matches on columns, and stacks columns as expected. The latter in `numpy < 1.23.0` concatenated the recarrays without concern for column ordering, which caused problems because `read_two_mtl_ledgers` and `read_one_mtl_ledger` can return different MTL column orders. This happens when reading a tile that has some ledgers _only_ in 1B (read with `read_one_mtl_ledger`, with the 1B MTL data model) and some ledgers in both 1A and 1B (read with `read_two_mtl_ledgers` and thus using the 1A data model, due to the fix in #855).  In `numpy >= 1.23.0` this triggers a crash, because `np.concatenate` added a check that the data model of all input arrays is exactly identical and it is not in this case.

Fixing this crash additionally required changing the endianness of `mtlprimdatamodel` and `mtlsecdatamodel` dtypes so that `read_two_ledgers` returns the same endianness as `read_one_ledger`, even when the columns are not in the same order.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -12,9 +12,11 @@ desitarget Change Log
     * The case caught in `PR #879`_.
     * The case for which a `1B` tile contains no secondary targets.
 * Add code to make a DR9/DR11 bricks file [`PR #881`_].
+* Correct the concatenation of MTL ledgers in read_mtl_in_hp [`PR #884`_]
 
 .. _`PR #881`: https://github.com/desihub/desitarget/pull/881
 .. _`PR #882`: https://github.com/desihub/desitarget/pull/882
+.. _`PR #884`: https://github.com/desihub/desitarget/pull/884
 
 4.7.2 (2026-04-29)
 ------------------
--- a/py/desitarget/io.py
+++ b/py/desitarget/io.py
@@ -2944,11 +2944,11 @@ def read_mtl_ledger(filename, unique=True, isodate=None, initial=False,
         # ADM need to catch cases where only one file actually exists.
         f1exists = os.path.isfile(filename[0])
         f2exists = os.path.isfile(filename[1])
-        if f1exists & ~f2exists:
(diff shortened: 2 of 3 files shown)


dylanagreen:
Trying to reproduce the fiberassignment of the night of 20260526 originally triggered the crash on concatenate, as during the day of 20260527 the main environment was updated to have `numpy=2.3.5`.  Running with the new version of desitarget successfully generates a fiberassignment for the failing tile (122940) and interestingly correctly reproduces the fiberassignment for that night. This is probably because the PRIORITY column is matched correctly, so all target priorities were correct up to the subpriority (which would be changed but for at least this tile that does not seem to have changed anything major). I also tested the night of 20250526 (one year ago) and the new version also successfully reproduced all 25 tiles from that night, although there were no 1B tiles observed that night.

dylanagreen:
I've additionally added a switch that will trigger the old concat behaviour (set to false), and corrected a bitwise inversion of a boolean. 

We may never need to actually trigger the old switch, I tested the nights of 20260526, 20250526, and 20260520 (which gives a spread of DARK, DARK1B, BRIGHT1B and BACKUP) and for all three nights in both "new" and "old" mode the fiberassignment is exactly reproduced for all tiles on all three nights. But I'm leaving it in on the off chance there is a tile where we need the old behaviour to reproduce.

geordie666 (approved):
I've independently tested this and it works for me, too.

I haven't tested the backwards compatibility part, but we can always revisit that when we need to produce alt MTLs.

I'm happy to merge this whenever you are ready!

PR #885, opened 2026-05-28
Process targets and skies from DR11 imaging.

geordie666:
This PR makes the necessary updates to process targets and sky locations selected from the DR11 imaging. Major changes include:

- Updates to handle the DR11 imaging data model.
  * For instance, the new `RELEASE` numbers for `DR11/south`.
- Changes the FITS extension from which the blob maps are read.
  * The blob maps are used to determine blank-sky locations
  * As of DR11 we now read the maps from extension 1 not extension 0 (while maintaining backward compatibility to earlier data releases).
- Adds some slurm scripts to improve batching using GNU parallel.
  * We previously used naked `srun` commands on Cori but `parallel` yields better performance on Perlmutter.

### DIFF
--- a/bin/select_skies
+++ b/bin/select_skies
@@ -151,6 +151,8 @@ else:
         skies = np.array(skies, dtype=object)
         ii = [sk is not None for sk in skies]
         skies = np.concatenate(skies[ii])
+    else:
+        skies = np.concatenate(skies)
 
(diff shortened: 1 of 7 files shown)


geordie666:
I'll leave this open for about an hour while unit tests pass. Then I'll merge this and tag `desitarget/5.0.0` in preparation for running DR11 targets.

PR #886, opened 2026-05-30
Update the default masking on MASKBITS for DR11

geordie666:
This PR updates the default set of imaging mask bits for DR11 to additionally include the new `RESOLVED`, `MCLOUDS` and `WISE_GAIA` bits from the [MASKBITS](https://github.com/legacysurvey/legacypipe/blob/DR11.0.9/py/legacypipe/bits.py#L81-L102) bit mask.

I've tested that this does indeed remove a substantial number of targets in, e.g., the Large Magellanic Cloud. As the update occurs directly in the function that sets the default masking, it will propagate to both targets and random catalogs.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,10 @@ desitarget Change Log
 5.0.1 (unreleased)
 ------------------
 
-* No changes yet.
+* Update the default set of imaging mask bits for DR11 [`PR #886`_]:
+    * Now always include new `RESOLVED`, `MCLOUDS` and `WISE_GAIA` bits.
+
+.. _`PR #886`: https://github.com/desihub/desitarget/pull/886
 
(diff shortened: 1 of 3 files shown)


geordie666:
Unit tests identified a failure mode when re-running DR9 targets. So, I updated the code for more robustness to backward compatibility. 

The failure mode was triggered by the fact that between DR9 and DR10 we updated the type of the `MASKBITS` column to `int32` instead of `int16`, but the new bits of interest are `2**17`, `2**18` and `2**19`. This means that checking for the new bits fails for the old (DR9) casting of `MASKBITS`.

The new approach is to check that the `MASKBITS` column is `int32` before defaulting to the `MASKBITS` appropriate to DR11. If the `MASKBITS` column is `int16` then the old (DR9) behavior is retained.

PR #887, opened 2026-06-04
Final updates to facilitate running all of the different targeting file types on DR11 imaging.

geordie666:
This PR adds the (hopefully) final updates to run targeting files on DR11. It mainly focuses on skies, supplemental skies and MTLs, and updates:

- The `RELEASE` number/schema for `SAFE` / `BAD_SKY` skies.
  * They now get assigned a release number of `RELEASE//1000`, but only if `RELEASE//1000 > 10` for backward compatibility (i.e., only starting with DR11).
    - This prevents `TARGETID`s from being duplicated with DR9 `SAFE`/`BAD_SKY` locations, which previously had a generic `RELEASE` number of 0.
- The supplemental skies so that they use Gaia DR3.
  * This guarantees that DR11 supplemental skies can't share a `TARGETID` with DR9 supplemental skies.
- The slurming/bundling function to add a gnu parallel option for supplemental skies.
- Functionality that adds targets to the MTLs to facilitate only adding DR11 targets in DR11 bricks.
- The MTL software to include new code to turn off DR9 targets in DR11 bricks.

I'll merge this in an hour or so to start making official products for DR11 targets.

### DIFF
--- a/bin/add_to_mtl_ledgers
+++ b/bin/add_to_mtl_ledgers
@@ -19,12 +19,12 @@ ap.add_argument("targfile",
 ap.add_argument("--dest",
                 help="Full path to the output directory to host the ledger. The \
                 filename and full directory structure are built on-the-fly from \
-                file headers in targdirname. [defaults to {})".format(mtldir),
+                file headers in targdirname. [defaults to {}]".format(mtldir),
                 default=mtldir)
 ap.add_argument("--obscon",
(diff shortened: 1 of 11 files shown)


geordie666:
Unit tests aren't passing but this is because they're hanging and different unit tests are hanging for different Python/numpy combinations. Overall, each Python/numpy combination has passed for some iteration. So, I'm going to merge this to make progress.

PR #889, opened 2026-06-18
BACKUP ToOs

jbanovetz:
I have made changes to `desitarget` so that if there are nearby and bright ToOs, these ToOs can be ingested and we can get the ToO spectra. 

The changes I made are to include the BACKUP program in the `ToO.py` file and to add the BACKUP ToOs to the `targetmask.yaml` under the `bgs_mask`. 

I ran this through `fba_launch` and was able to successfully assign the BACKUP ToOs to a tile. The command I used was 

```
fba_launch --doclean n --dr dr9 --dtver 2.2.0 --fba_use_fabs 1 --forcetileid y --gaiadr gaiadr2 --goaltime 60.0 --goaltype BACKUP \
--ha -12.820331882448016 --hdr_faprgrm backup --hdr_survey main --lookup_sky_source gaia --margin_gfa 0.4 \
 --margin_petal 0.4 --margin_pos 0.05 --mintfrac 0.85 --mtltime 2026-06-25T20:04:30+00:00 --pmcorr n \
--pmtime_utc_str 2026-06-26T07:36:43+00:00 --program BACKUP --rundate 2026-06-20T17:00:55+00:00 --sbprof PSF \
--sky_per_petal 40 --sky_per_slitblock 1 --standards_per_petal 10 --steps tiles,sky,gfa,targ,scnd,too,fa,zip,move \
--survey main --tiledec -17.5 --tileid 46163 --tilera 148.357 --try_too_all --worldreadable  --log_stdout \
--outdir $BACKUP_TOO_DIR
```

And the `TOODIR` used was: `/global/cfs/cdirs/desi/users/banovetz/ToO_BACKUP_test`


py/desitarget/data/targetmask.yaml
```diff
@@ -392,6 +395,9 @@ priorities:
         BGS_WISE_SOUTH:   SAME_AS_BGS_FAINT_SOUTH
         BGS_WISE_NORTH:   SAME_AS_BGS_FAINT_SOUTH
 
+        BACKUP_TOO_HIP:  {UNOBS: 100,  MORE_ZGOOD: 2,    MORE_ZWARN: 2,    DONE: 2, OBS: 1, DONOTOBSERVE: 0, MORE_MIDZQSO: 0}
```
geordie666: Let's set `UNOBS` for `BACKUP_TOO_HIP` at the standard "very high" priority value of `9500` in case we ever add additional `BACKUP` targets at higher priority. For example, there are currently secondary targets (WDs) in the backup program that have an `UNOBS` priority of `1998`.

py/desitarget/ToO.py
```diff
@@ -364,7 +364,7 @@ def _check_ledger(inledger, survey="main"):
     # ADM and the priorities are all LO or HI.
     allowed = {"TOO_TYPE": {'FIBER', 'TILE'},
                "TOO_PRIO": {'LO', 'HI'},
-               "OCLAYER": {'BRIGHT', 'DARK'}}
+               "OCLAYER": {'BRIGHT', 'DARK','BACKUP'}}
```
geordie666: You're missing a space after a comma.

py/desitarget/ToO.py
geordie666: I think the code will fail, here, because `bitname` will be, e.g., `BACKUP_TOO_LOP` and `sMx` is the mask of _secondary_ bits. Because we have to store `BACKUP_TOO_LOP` in `mMx` (the mask of _MWS_ bits) `sMx[bitname]` is not defined.

I tested the code by adding this line to a mock-up of the `ToO-input.ecsv` file:

```
350.555538   0.404432 0.000000 0.000000 2000.0 XH/AP FIBER LO BACKUP 61219.42268462 61249.42268462 2999999999
```

And did, indeed, trigger such a failure. So, I think you need some extra lines indicating which bit mask to query to retrieve the `BACKUP_TOO` bits.

geordie666 (changes requested):
I think the code will currently fail for the reason stated in my 3rd review comment. Please update the code and test that `select_too` proceeds to completion.

Please also update the `doc/changes.rst` file to document this PR.

jbanovetz:
Thank you are the review and the catch! I forgot to rerun `select_too` once I changed the mask to BGS. I have made the changes and checked with both `select_too` and `fba_launch` and it works as intended. One question I have is do I also need to make a change to this line (450):

`outdata[dcol] = dMx["SCND_ANY"]`

since the BACKUP ToO is technically not a secondary? This is more of a bookkeeping issue so please let me know if I should change it to `dMx['BGS_ANY']`.

geordie666:
Oh, yes, excellent point. That should indeed be changed to

`outdata[dcol] = dMx["BGS_ANY"]`

We might want to check with @dylanagreen (if you didn't already) as to whether `fiberassign` will still find ToOs with the `BGS_ANY` bit set instead of the `SCND_ANY` bit. But, yes, formally you should make that update to `BGS_ANY`.

geordie666:
On further review, I don't think this approach will work for `fiberassign` because we currently don't write out `BGS_TARGET` to the `ToO-fiber.ecsv` and `ToO.ecsv` files. 

I think perhaps the cleanest solution, then, is to instead put these new bits directly in the `desi_mask` and associated `DESI_TARGET` column instead of using `BGS_TARGET`. Perhaps as bits `41` and `42`?

jbanovetz:
Thanks for looking at it again! I will give that a try locally and report back!

jbanovetz:
I made the changes and it looks like it worked! I reran `select_too` and my `fba_launch` command. Is there any other test we need to do?

py/desitarget/ToO.py
```diff
@@ -457,8 +457,8 @@ def finalize_too(inledger, survey="main"):
             # JB: Due to lack of available bits in the Secondary Program,
             # I added the BACKUP ToO bits to the BGS survey target mask
```
geordie666: This comment is now out of date. Either remove it or modify it.

py/desitarget/data/targetmask.yaml
```diff
@@ -40,6 +40,10 @@ desi_mask:
     # - [STD_FAINT_BEST,    38, "High quality faint standard stars",                 {obsconditions: DARK}]
     # - [STD_BRIGHT_BEST,   39, "High quality bright standard stars",                {obsconditions: BRIGHT}]
 
+    # JB: BACKUP ToO Targets
+    - [BACKUP_TOO_HIP,    41, "Targets of Opportunity from rolling ledger",   {obsconditions: BACKUP,      flavor: 'TOO', updatemws: False}]
```
geordie666: `flavor` and `updatemws` don't need to be included, here. Those are only relevant for secondary targets.

py/desitarget/data/targetmask.yaml
```diff
@@ -40,6 +40,10 @@ desi_mask:
...
+    - [BACKUP_TOO_LOP,    42, "Targets of Opportunity from rolling ledger",   {obsconditions: BACKUP,      flavor: 'TOO', updatemws: False}]
```
geordie666: `flavor` and `updatemws` don't need to be included, here. Those are only relevant for secondary targets.

geordie666 (commented):
A few more comments to address before we can merge this.

In addition, please update the `doc/changes.rst` file to describe this PR.

geordie666:
@jbanovetz: I think this is ready to merge from the `desitarget` side once you update the changes docs to document this PR.

One question, though, will complementary changes to `fba_launch` be required? If so we should make sure those are also made before/when we merge this PR.

jbanovetz:
I added the PR to `changes.rst` and committed it.

I don't think this requires any changes to `fba_launch`. The only change I made was changing some hard coded paths to some personal repos.

jbanovetz:
Actually there might be some changes needed. I ran my `fba_launch` command from above and it is now finding no ToOs in my test case (should be 500 BACKUP ones just as a test). I will do some debugging and try and find a solution

jbanovetz:
Sorry, for the delay, I did so debugging and I couldn't find any because I forgot to change the `TIMESTAMP` column for my injected sources. Upon rerun, I was able to get all 500 ToOs!

For more description on my test (in case we want to look this up later). I took a BACKUP tile (`46163`) that I got from Dylan and took a random sampling of 500 of the targets observed with this tile. I then made 500 ToOs with a `PROGRAM=BACKUP` from those targets and added them to my local `ToO-input.ecsv` file. I then ran `select_too` pointing to the directory above in the first description and produced the `ToO-fiber.ecsv` file and confirmed that it ingested the `BACKUP` ToOs. I then had to change the `TIMESTAMP` to something before `2026-06-25T20:04:30+00:00` and then ran `fba_launch`. I didn't actually have to change anything in `fba_launch` for this to work. The generated `too.fits` file has all 500 ToOs.

I am not sure if there is an extra step I need to do to make sure that they are valid targets or if being a potential assignment is enough?

geordie666:
@dylanagreen should assess your "potential assignments being enough" from the `fiberassign` side question.

One question from me: Is this compatible with your recently merged branch in PR #874?  So, will the "all" files also be automatically written for `BACKUP` targets? It might be worth merging `main` into your branch and checking that still all naturally flows.

jbanovetz:
I just did the check and it successfully ingests them with `select_too`! Here is the directory (I wanted to preserve the original one): `/global/cfs/cdirs/desi/users/banovetz/ToO_BACKUP_test_2`

dylanagreen:
@jbanovetz, have you looked at the output `fiberassign-*` file for the test tile you ran? The ToOs being in the input -too file means that at the very least `fba_launch` recognized them as targets that could be assigned on this tile, but if none of them get through to the `fiberassign-*` file then somewhere in the `fba_run` step they might be getting lost (`fba_run` does do some cuts on bitmasks when loading just as an extra sanity check, so we want to make sure that they actually propagated all the way through to the assignment step itself).

jbanovetz:
Well after some digging around and then giving up and asking Claude, it looks like it is Claude-1 me-0 (at least for now). Claude found an error in my branch (I forgot an `[ii]` when setting the bit mask). 

It also correctly identified that we did need to make a change to `fiberassign/targets.py` file. I needed to add my two new masks to the list for `default_main_sciencemask()`. Running `fba_launch` now, I get from 500 targets, 430 are potential assignments, and 7 get assigned. I made a PR for that change in `fiberassign` here: https://github.com/desihub/fiberassign/pull/520

I am a little surprised it is only 7 as the `BACKUP_HIP` mask has the highest priority. I don't know if there is a way to check if it is a priority issue easily or not?

araichoor:
> I am a little surprised it is only 7 as the BACKUP_HIP mask has the highest priority. I don't know if there is a way to check if it is a priority issue easily or not?

@jbanovetz : what I would first do is to check is to look at the potential assignments for the 500 backup too's.
I could play around a bit to explore more, but  your scratch folder `/pscratch/sd/b/banovetz/BACKUP_ToO_testing/046/` is not readable by others...

araichoor:
> I am a little surprised it is only 7 as the BACKUP_HIP mask has the highest priority. I don't know if there is a way to check if it is a priority issue easily or not?

thanks for posting the `fba_launch` command :); I ve run that on my side.
I see that 430/500 of the backup too's are reachable by a fiber.
the issue is that your backup too's are `BACKUP_TOO_LOP`, not `BACKUP_TOO_HIP`, hence they ve a very low priority of 10.

araichoor:
hacking your `Too-fiber.ecsv` file to change the backup too's to `BACKUP_TOO_HIP` and rerunning `fba_launch` provides 403 assigned backup too's.
here: `/global/cfs/cdirs/desi/users/raichoor/tmpdir/ToO_BACKUP_test/`.

jbanovetz:
Thank you for that check @araichoor! I got the same number of potential assignments and didn't realize I left the priority as `LO`. That would explain the 7 then!

geordie666:
I'm closing this (without merging) now that it's been superseded by PR #900.

Issue #895, opened 2026-07-27
github tests require data.desi.lbl.gov to run any tests


sbailey:
Github tests currently require data.desi.lbl.gov to download a snapshot of surveyops.  During outages, this means that none of the tests run because of this crash:
```
Run wget -nv [https://data.desi.lbl.gov/public/epo/example_files/surveyops_${SURVEYOPS_VERSION}_ops.tar.gz](https://data.desi.lbl.gov/public/epo/example_files/surveyops_$%7BSURVEYOPS_VERSION%7D_ops.tar.gz)
ERROR: no certificate subject alternative name matches
	requested host name ‘data.desi.lbl.gov’.
To connect to data.desi.lbl.gov insecurely, use `--no-check-certificate'.
Error: Process completed with exit code 5.
```

Consider updating the test harness to make this a non-fatal download and make the tests robust to not having these data, so that at least it can run the other tests.

Potential downside: if this download started failing for other reasons, the github tests would "pass" and we probably wouldn't notice that something was being unexpectedly skipped.

sbailey:
The situation has gotten worse: github tests run on azure cloud nodes, which are also used by hackers and are starting to get blocked by NERSC cyber security.  So even when data.desi.lbl.gov is available, there is a random chance that the test runner is on a blocked node and can't download the file anyway.

I'm working with NERSC security about options, but I suspect that answer won't be for them to allow-list all azure servers.  It's probably more viable for us to make a mini version of surveyops_2.0_ops.tar.gz for testing and bundle it with the repo, or otherwise make download failures exit early and skip those tests.

Related: #896 is updating test runner infrastructure related to healpy+numpy; I'll try to get a workaround in there

sbailey:
#896 gets the github tests working again in the sense that it will fail/skip more gracefully if the surveyops download fails.

It appears that the test just need $DESI_SURVEYOPS/ops/tiles-main.ecsv .  That file is a bit big (3.7M), but we could bundle a subset of it with desitarget itself for the purposes of tests.  I suggest:
  * If $DESI_SURVEYOPS/ops/tiles-main.ecsv exists, use it (allows tests at NERSC to use the real file)
  * Otherwise use a mini-copy bundled with desitarget itself (allows github tests to run without having to download from data.desi.lbl.gov).

The work is to identify what subset of tiles-main.ecsv to use and setup infrastructure for test_qa.py to fall back if needed.

geordie666:
Addressed in #897.

PR #896, opened 2026-08-10
pip install healpy and pinned numpy together

sbailey:
This PR mimics desihub/fiberassign#515 to pip install healpy in the same line as pinned numpy. Otherwise testing with numpy<2 results in an incompatible healpy version.  Provided that this works, we can then cherry-pick commit 2c4c9c5f81e82a0397b7da7fa5aafdd61df42f2c into the other open PR branches that are having test failures.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -79,7 +79,13 @@ jobs:
               env:
                 SURVEYOPS_VERSION: '2.0'
               run: |
-                wget -nv https://data.desi.lbl.gov/public/epo/example_files/surveyops_${SURVEYOPS_VERSION}_ops.tar.gz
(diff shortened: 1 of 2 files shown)


sbailey:
## Summarizing changes ##

Original issue this PR was intended to fix:

  * Installed healpy and numpy in same pip command to ensure compatible versions

Additional test updates due to NERSC starting to block azure nodes where tests are running (see #895):

  * Download stops faster if data.desi.lbl.gov is unavailable
  * Tests skipped if $DESI_SURVEYOPS/ops/tiles-main.ecsv is unavailable (e.g. failed download; previously

Tests pass again, so I'm going to self-merge this and merge changes into other open PR branches to unblock their tests.  I'll leave #895 open because we might want a cleaner fix.

PR #897, opened 2026-08-13
Add a minimal tiles-main.ecsv file so the DESI_SURVEYOPS snapshot isn't needed

geordie666:
This PR adds a stripped-down version of the `tiles-main.ecsv` file that is sufficient for the `test_qa.py` unit test. This file will be used in the event that the `$DESI_SURVEYOPS` environment variable doesn't exist, hopefully ensuring that:

1. The `DESI_SURVEYOPS` snapshot doesn't need to be downloaded and installed.
2. But the `DESI_SURVEYOPS` directory will be used when available, i.e. if tests are run at NERSC.
3. Full coverage is maintained.

This should address #895.


geordie666:
@sbailey: This update obviates the need to download the `DESI_SURVEYOPS` snapshot by creating a small fit-to-purpose `tiles-main.ecsv` file, which is probably the easiest way to address #895. But, please tell me if you actually intended us to keep trying to download `DESI_SURVEYOPS` and only have a backup option in the event of repeated failure.

Note that coverage is failing for reasons that were fixed in #874.

.github/workflows/python-package.yml
```diff
@@ -75,17 +75,8 @@ jobs:
...
-                SURVEYOPS_VERSION: '2.0'
-              run: |
-                echo "=== Determine Egress IP ==="
-                curl -s https://ifconfig.me || curl -s https://ipinfo.io/ip || curl -s https://api.ipify.org
-                echo
-                echo "=== Download attempt (verbose) ==="
-                (wget -v --timeout=10 --tries=1 https://data.desi.lbl.gov/public/epo/example_files/surveyops_${SURVEYOPS_VERSION}_ops.tar.gz && tar xzf surveyops_${SURVEYOPS_VERSION}_ops.tar.gz) || echo "Download of surveyops failed, moving on"
             - name: Run the test
-              run: DESI_SURVEYOPS=$(pwd) pytest
+              run: pytest
```
sbailey, line 79: Please also make the same changes to the coverage section below.

py/desitarget/test/test_qa.py
```diff
@@ -19,10 +19,13 @@
...
-if ('DESI_SURVEYOPS' in os.environ) and os.path.exists(os.path.expandvars('$DESI_SURVEYOPS/ops/tiles-main.ecsv')):
-    surveyops_ok = True
-else:
-    surveyops_ok = False
+surveyops_ok = ('DESI_SURVEYOPS' in os.environ) and os.path.exists(
+    os.path.expandvars('$DESI_SURVEYOPS/ops/tiles-main.ecsv'))
+if not surveyops_ok:
+    os.environ["DESI_SURVEYOPS"] = str(
+        resources.files('desitarget').joinpath('test/t'))
+log.info(f'DESI_SURVEYOPS is set to {os.environ["DESI_SURVEYOPS"]}')
```
sbailey: Let's move this logic into the TestQA `setUpClass` so that `tearDownClass` can reset/unset `DESI_SURVEYOPS` if needed.  As-is, the act of importing this test can have the side effect of changing $DESI_SURVEYOPS which could potentially impact other tests depending upon the order in which they are run.

py/desitarget/test/t/surveyops/ops/tiles-main.ecsv
```diff
@@ -0,0 +1,25 @@
+# %ECSV 1.0
```
sbailey, line 1: Minor: Consider putting this file under test/t/surveyops/ops which provides slightly more conceptual isolation that this directory is mimicking the surveyops svn and might be cleaner in the future if we further expand it with mini-test files for other surveyops tests.

sbailey (changes requested):
Overall approach is good: use mini test file on github without having to download anything, but use the real surveyops directory if running on a system where that exists (e.g. NERSC).

A few updates requested inline.

.github/workflows/python-package.yml
geordie666: Done in [0ad81a4](https://github.com/desihub/desitarget/pull/897/commits/0ad81a4d3b87950f09347cf63c5e519baad0d535).

py/desitarget/test/test_qa.py
geordie666: Done in [1b92b3c](https://github.com/desihub/desitarget/pull/897/commits/1b92b3cb0091d98b8268afbbcac8fed65aae1c03).

py/desitarget/test/t/surveyops/ops/tiles-main.ecsv
geordie666: Done in [64edbd2](https://github.com/desihub/desitarget/pull/897/commits/64edbd2b9fa52c2e50d1f9861b864567700544c9).

sbailey (approved):
Looks good. Thanks for this update which will simplify our testing workflow. Merge when ready.

PR #899, opened 2026-09-04
Run geomask tests from a temp directory to avoid leftover files

sbailey:
bundle_bricks() writes its driver/gnu-parallel/tasks scripts to the current working directory, so test_bundle_bricks was leaving targets-driver.sh, targets-gnu-parallel.sh, and targets.tasks behind in whatever directory pytest was run from.  This PR updates the tests to use a temporary directory and cleanup after itself so that it doesn't leave files behind in the code directory when running tests.


sbailey:
Superseded by #901 which fixes the issue in this PR plus additional similar cases.  Closing without merging.

PR #901, opened 2026-09-09
Use tempdirs in tests instead of writing to the current directory

sbailey:
test_geomask, test_io, test_mtl, and test_veto all wrote output files/directories directly to the current working directory (or a relative path), which fails on a read-only filesystem, e.g. /global/common/software when running from a NERSC batch node. Switch them to tempfile.mkdtemp()/TemporaryDirectory, restoring cwd afterward where the code under test (bundle_bricks) writes to cwd itself.

Supersedes #899, fixing that problem plus more.  Co-authored with Claude.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -6,11 +6,12 @@ desitarget Change Log
 ------------------
 
 * Fix tests that wrote to the current directory instead of a tempdir,
-  so that tests pass on a read-only filesystem.
+  so that tests pass on a read-only filesystem [`PR #901`_].
 * Python 3.14 support: Update multiprocessing to use fork instead of
   forkserver [`PR #898`_].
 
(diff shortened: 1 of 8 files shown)


sbailey:
Implementation note: where possible, we used the pattern
```python
origdir = os.getcwd()
with tempfile.TemporaryDirectory() as testdir:
    try:
        os.chdir(testdir)
        # do something
    finally:
        os.chdir(origdir)
```
or
```python
self._testdir_obj = tempfile.TemporaryDirectory()
self.testdir = self._testdir_obj.name
...
self._testdir_obj.cleanup()
```
instead of caching `tempdir` and later calling `shutil.rmtree(tempdir)`, which could be dangerous if `tempdir` was accidentally updated to a different value without realizing the implications of auto-removing it later.  The pattern used here is most robust to guarantee that only the temporary directory itself is the thing that could be cleaned up.

PR #903, opened 2026-09-18
Fix uninitialized-memory bug in make_mtl() for zcat-unmatched targets

sbailey:
@geordie666 upon further investigation, it appears that the intermittent failures of test_mtl were due to a real bug with uninitialized memory in make_mtl.  As explained by Claude and lightly edited by me:

> make_mtl() allocated Z, NUMOBS, ZWARN, ZTILEID (and the main-survey Z_QN/IS_QSO_QN/DELTACHI2 columns) with np.empty() but only overwrote them for targets matched to the zcat, leaving raw uninitialized memory for any target with no zcat entry (the normal case for trim=False). This usually looked like 0.0, but occasionally leaked a stale, reused heap value, causing intermittent, heap-state-dependent failures such as TestMTL.test_midz_no_lock_in spuriously matching two rows on Z == 0.5 (seen in occasional CI as "cannot assign 2 input values to the 1 output values").

The fix here is to ensure that every column gets a default, albeit still using the original loop over columns and `np.empty` to ensure consistent order and dtype.  A new test is added that consistently fails without the patch in mtl.py.

@geordie666 this PR ensures reproducible defaults for every column for non-matched targets, but do you agree with these defaults or want to switch them to something else?  We used -1 for most defaults, though the previous code was usually getting away with the default for `np.empty` often being 0.  What should unmatched targets have?

@geordie666 if you agree that this uninitialized memory actually is a bug, do you know what the consequences are for past production mtl operations?  Or is it just a test artifact that any rows were unmatched?

Related: PR #902 was the incorrect workaround but I think this PR fixes the actual underlying issue. Co-authored with Claude.


py/desitarget/test/test_mtl.py
```diff
@@ -325,6 +325,33 @@ def test_midz_no_lock_in(self):
...
+        mtl = make_mtl(t, "DARK", zcat=zcat, trim=False)
+
+        # exactly one target should never be matched to the zcat.
+        unmatched = ~np.isin(mtl["TARGETID"], zcat["TARGETID"])
+        self.assertEqual(np.sum(unmatched), 1)
+
+        # that target was never observed, so should carry well
+        # defined defaults instead of uninitialized memory (which is
+        # usually, but not reliably, zero).
+        self.assertTrue(np.all(mtl["Z"][unmatched] == -1))
```
geordie666, line 346: Consider changing these comparisons to reflect the explicit type of columns that are floats (`Z`, `Z_QN`, `DELTACHI2`); informative, although not strictly necessary:

```
self.assertTrue(np.all(mtl["Z"][unmatched] == -1.))
```

and the same for:

```
Z_QN -1.
DELTACHI2 -1.
```

py/desitarget/mtl.py
```diff
@@ -843,8 +843,32 @@ def make_mtl(targets, obscon, zcat=None, scnd=None, trim=False,
...
+    # columns that are guaranteed to be overwritten in full, for every
+    # row, immediately below -- so np.empty() is safe for them, even
+    # for targets that have no match in the zcat.
+    fullyoverwritten = {"NUMOBS_MORE", "TARGET_STATE", "TIMESTAMP", "VERSION"}
+
+    # every other column is only conditionally overwritten (via
+    # zmatcher), below, so needs a real default here. Otherwise a
+    # target with no match in the zcat (i.e. that was never observed)
+    # would be left with uninitialized memory in that column.
+    zcoldefaults = {"NUMOBS": 0, "Z": -1, "ZWARN": -1, "ZTILEID": -1,
```
geordie666, line 857: Consider changing the "defaults" dictionary to explicitly reflect the type of the columns that are floats (`Z`, `Z_QN`, `DELTACHI2`); informative, although not strictly necessary, i.e.:

```
{"NUMOBS": 0, "Z": -1., "ZWARN": -1, "ZTILEID": -1, 
"Z_QN": -1., "IS_QSO_QN": -1, "DELTACHI2": -1.}
```

geordie666 (approved):
Thanks @sbailey (and Claude). This is an esoteric, really interesting (and non-malicious) bug, which is why, I think, we've never quite tracked down the issue. In particular, the unit test that intermittently fails improves coverage but, I think, never tests a case we actually use in practice.

Carefully reviewing the MTL code, we, in reality, call the `make_mtl()` function in two distinct ways:

1. When we have a `zcat` and we're dynamically updating the MTLs, we always call `make_mtl()` with `trimtozcat=True`. In this case, the `np.empty()` commands identified by Claude produce entries that are always subsequently discarded.

2. When we do not have a `zcat` and we're making _initial_ MTL entries, we call `make_mtl()` with `trimtozcat=False`. In this situation, the `np.empty()` commands will be propagated to the outputs. But, that propagation doesn't really matter as it produces the `UNOBS` entries for which the redshift-related columns are meaningless (and which are subsequently superseded with a new ledger entry once we have redshifts/observations). I could spool through the `UNOBS` entries in every ledger to try to find one that doesn't conform to the expected defaults, but any such cases wouldn't actually be problematic, or change the context of this PR.

The intermittently failing unit test covers a case where we pass a `zcat` to `make_mtl()` but do _not_ pass `trimtozcat=True`. It's possible, in fact, that passing `trimtozcat=True` in that unit test would have circumvented the problem!

Nevertheless, the chain-of-events identified by Claude is at least _possible_ even if it isn't currently really utilized. So, I approve of the updates.

I checked Claude's chosen defaults by picking an `UNOBS` entry in a ledger and examining the relevant values:

```
from desitarget.io import read_mtl_ledger
a = read_mtl_ledger("mtl-dark1b-hp-4693.ecsv", unique=False)

zcoldefaults = {"NUMOBS": 0, "Z": -1, "ZWARN": -1, "ZTILEID": -1,
    "Z_QN": -1, "IS_QSO_QN": -1, "DELTACHI2": -1}

for k in zcoldefaults:
    print(k, a[0][k], zcoldefaults[k])
    ...: 
NUMOBS 0 0
Z -1.0 -1
ZWARN -1 -1
ZTILEID -1 -1
Z_QN -1.0 -1
IS_QSO_QN -1 -1
DELTACHI2 -1.0 -1
```

So, Claude's proposed defaults look sensible, although we might want to use the actual types for clarity (see my review comments).

This tl;dr review in brief:

- I think Claude's proposal is reasonable and is one way to fix the intermittently failing unit test.
- I think the fix will be benign as, to the best of my knowledge, we never actually use the unit-tested case in daily operations.
- Formally, I think it's sensible to make the proposed update because the code _could_ be used in the unit-tested manner, even though we _haven't_.

sbailey:
I added the more explicit types for the constants (-1 vs. 1.0), and also expanded the pre-existing `test_mtl` function to confirm expected dtype for every column and not just names.

PR #904, opened 2026-09-21
Update Decisions for LyA QSOs to handle 1a tiles in the 1b era

geordie666:
This PR updates the MTL code so that the LyA decisions requested for the 1b phase are also applied on 1a tiles. 

Such updates are necessary because the LyA decisions were _previously_ only applied on `DARK1B` tiles in the original main survey area, but now there are new southerly `DARK` (_not_ `DARK1B`) tiles that are part of the 1b observations.

The PR includes:

- Modifying the `update_lya_1b()` function that changes the total number of observations for QSO targets from 4 to 6 so that it only updates QSO targets that haven't _yet_ had their total number of observations changed from 4 to 6.
- Changing the MTL logic so that ELG targets on 1a tiles that turn out to meet the LyA quasar criteria after a spectrum is obtained are also updated to receive 6 total observations.
- Decrementing QSO targets that turn out to be at low redshift to only have 2 total observations (i.e. 6 -> 2 instead of just 4 -> 2).

As part of this PR, the `io.write_mtl_ledger()` function was also updated to no longer check whether "mixed" Data Releases were being written to a single ledger. This sort of write is now the expected behavior in the 1b/DR9+DR11 era.


geordie666:
One additional update related to ELGs-that-turn-out-to-be-LyA-QSOs needing 6 observations:

- Ensure these ELGs actually achieve the `DONE` priority (`2`) rather than having the higher LyA QSO priority forever.

Such ELGs could never have received 6 observations on `DARK1B` tiles, but could be visited 6 times on `DARK` tiles in the 1b era.

h/t: That this was possibly an issue was uncovered by @segasai's audit of `desitarget` using Opus/Fable.

